PR #99, opened 2026-01-21
Make ScanBundle streaming over gRPC

### LINKED ISSUE #63: ScanBundle: files should be streamed/paginated in the response
To avoid some initial latency, and issues with really large amount of files, we should try to stream the data and delete files. It is unclear to me right now what's the best way to do this (do we want a stream interface in floecat?).
Also, since we have data and delete files, we may need two streams (potentially something that can be optimized in some cases, but right now, without any additional insight we need all delete files before handling any data file).

mcusack-floe:
Addresses issue #63.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,10 +108,7 @@ REACTOR_SERVICE := -pl service -am
 REACTOR_REST := -pl protocol-gateway/iceberg-rest -am
 
 # ---------- Version / Artifacts ----------
-VERSION := $(shell sed -n 's:.*<revision>\(.*\)</revision>.*:\1:p' pom.xml | head -n1)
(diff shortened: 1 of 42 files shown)


mrouvroy-floe (commented):
Some tests we could add :
- pruning behavior that skips a file (to ensure stream doesn’t fail / doesn’t emit null)
- delete/data classification consistency (including unknown/unset file_content)
- cancellation (client stops reading → query state updates)

u3064628111-afk (approved):

mrouvroy-floe:
Overriden by https://github.com/eng-floe/floecat/pull/136

PR #105, opened 2026-01-27
Allow for create account to take optional resource id

kwesterfeld-floe:
- When synchronizing with an external account model, it must be possible to create a commmon account id for identity propagation to floecat.


kwesterfeld-floe:
Added missing translation keys and tests

mrouvroy-floe (approved):

PR #112, opened 2026-01-29
Session token validation (external IDP/JWT token integration)

kwesterfeld-floe:
* Session token validation

  - Add quarkus-oidc extension
  - Add token validation if header is set
  - Add config flag to disable dev-context fallback when
    principal/query headers are missing
  - Adjust default test configurations to disable loading of quarkus
    development oidc/keycloak service (which would require docker)
  - Add integration tests for
      - passing token
      - passing token with invalid aud claim
      - passing token with missing aud claim
      - passing token with invalid session header name
      - passing token with missing session header
      - passing token with missing principal header
      - passing malformed JWT
      - passing expired token
      - passing token with exp is before iat
      - passing token not yet valid (nbf)
      - passing token with invalid JWT private key signer
  - Update README.md to include basic guidance on setup of external
    OIDC token integration and pointer to external docs for configuration.
  - Update vs.code test run configuration to allow wider
    use for debugging all unit tests (not just KV-related ones)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -16,6 +16,17 @@
       "vmArgs": [
         "-Xmx1g",
 
+        "-Djava.util.logging.manager=org.jboss.logmanager.LogManager",
+        "--add-reads=org.apache.arrow.flight.core=ALL-UNNAMED",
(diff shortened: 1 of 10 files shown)


kwesterfeld-floe:
Maxence has a usecase to warm the planner cache that would need to omit the query id header.  More to think about.

PR #121, opened 2026-02-03
Credential modeling and resolver pluggable

kwesterfeld-floe:
- Create AuthCredentials for use in AuthConfig
  - Model each of the paths required for various authentication schemes stored in AuthCredentials:
      - bearer token
      - oauth client id/client secret - aws keys, assume role, assume role with web identity - token exchange: rfc 8693, azure on-behalf-of, gcp workload federation
  - Make the auth credentials secret mapping in grpc optional; allow inline
  - Add resolver pattern to allow extensions to resolve the secret ref
  - Add some tests

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/context/impl/InboundContextInterceptor.java
+++ b/service/src/main/java/ai/floedb/floecat/service/context/impl/InboundContextInterceptor.java
@@ -126,7 +126,10 @@ public class InboundContextInterceptor {
     String engineVersion = readHeader(headers, ENGINE_VERSION_HEADER);
     String engineKind = readHeader(headers, ENGINE_KIND_HEADER);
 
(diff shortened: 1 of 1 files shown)


mrouvroy-floe:
#### Auth resolver seems to be a no-op ? 
Both ConnectorsImpl.resolveCredentials and ReconcilerService.resolveCredentials only apply resolved creds when the optional is present, and the only resolver bound in production (DefaultCredentialResolver) always returns empty, so auth.secret_ref never produces any credentials/headers/options. 
Are we planning to implement a new resolver (or make the default throw/log) so that unresolved keys fail fast ?

#### Doc is missing 
I think we need to doc the new fields (especially auth.credentials, token exchange inputs, expected headers/properties) and explain the resolver contract (how secret_ref is resolved, how to implement/plug in a resolver) so new user can use it easily

PR #123, opened 2026-02-03
Add token-based auth to CLI; harden principal headers and dev auth defaults

mrouvroy-floe:
# Add token-based auth to CLI; harden principal headers and dev auth defaults

This change introduces token-based authentication to the Floecat CLI, while hardening server-side auth defaults to ensure dev-only mechanisms cannot be used in production.

The CLI now cleanly supports both token auth and principal-based dev auth, with explicit behavior and clearer error messaging. On the server side, permissive auth paths are disabled by default and only enabled in dev/test profiles.

###  Token-based authentication for the CLI (recommended path)
- Added --token flag and FLOECAT_TOKEN environment variable.
- Tokens are sent via a configurable session header (default: Authorization).
- When token auth is active:
        - Server identity and account are derived from token claims.
        - The CLI account command becomes optional and informational only.
        - Clear user-facing hints are printed for UNAUTHENTICATED and PERMISSION_DENIED errors.

### Explicit dev-mode principal auth (fallback only)
- When no token is provided, the CLI can send a x-principal-bin header derived from:
        - current account
        - minimal read/write permissions for exploration
        - This mode is intended only for dev/test environments.

### Server-side auth hardening (secure by default)
- In non-dev profiles:
        - x-principal-bin is rejected
        - implicit dev context is disabled
        - Dev/test profiles explicitly re-enable these features.
        - Configuration makes it clear that client flags are not a security boundary.

### Clearer client-side validation and errors
- CLI now:
        - fails fast on non-UUID account IDs in principal auth mode
        - explains when account has no effect under token auth
        - Improved error messages and hints for common auth misconfigurations.

### DIFF
--- a/client-cli/src/main/java/ai/floedb/floecat/client/cli/CliAuthInterceptor.java
+++ b/client-cli/src/main/java/ai/floedb/floecat/client/cli/CliAuthInterceptor.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2026 Yellowbrick Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


mcusack-floe:
It seems we are both implementing the same thing. I've been adding OIDC stand-alone authentication, and part of that involved updating the CLI to accept an access token.

PR #136, opened 2026-02-09
Replace FetchScanBundle with fully streaming scan API 

### LINKED ISSUE #63: ScanBundle: files should be streamed/paginated in the response

mrouvroy-floe:
# Replace FetchScanBundle with streaming scan API 

This PR replaces the legacy FetchScanBundle RPC with a streaming, handle-based scan API:
- InitScan
- StreamDeleteFiles
- StreamDataFiles
- CloseScan

The new API separates scan initialization, delete enumeration, and data file enumeration, while preserving correctness and supporting large tables without materializing full scan bundles in memory.

This change is fully wired end-to-end:
- gRPC API and protos
- service implementation and query lifecycle
- scan session management
- CLI, Trino connector, and Iceberg REST gateway
- unit + integration tests

---

## Motivation

FetchScanBundle forced the server to:
- compute all data + delete files eagerly
- return them in a single response
- couple data files to deletes via positional indices

This created scaling, memory, and extensibility limits, especially for:
- large tables
- many delete files
- future per-file delete applicability

The new API removes those constraints while keeping the initial implementation simple and correct.

---

### 1.	InitScan
- Validates query + pinned snapshot
- Applies pruning hints (required columns, predicates)
- Returns:
         - ScanHandle (opaque)
         - TableInfo (schema, partition spec, metadata location)
        
### 2.	StreamDeleteFiles
- Streams delete files in batches
- Assigns stable delete_ids
- Must be fully consumed before data streaming
- Server enforces ordering (FAILED_PRECONDITION otherwise)

### 3.	StreamDataFiles
- Streams data files in batches
- Each DataFile includes a DeleteRef
- Current implementation emits all_deletes=true

### 4.	CloseScan
- Best-effort cleanup
- Idempotent
- Sessions are also auto-released on stream completion or cancellation

---

## What this enables next (future work)

This API is designed to unlock incremental future improvements without breaking changes:
- Per-file delete applicability
        - Replace all_deletes=true with selective delete_ids
        - Use partition / equality / sequence metadata to compute applicability
- Delete index shrinking
        - Avoid propagating irrelevant deletes to unrelated data files
        - Reduce memory and planning overhead for large delete sets
- More efficient EE scanning
        - Early exit / partial planning for large tables
        - Adaptive batching based on downstream consumers
- Additional file metadata
        - Column stats, bloom filters, sketches
        - Optional, opt-in via InitScanRequest
- Alternative execution models
         - Parallel consumers of the same scan handle
         - Engine-specific optimizations 

Fixes : https://github.com/eng-floe/floecat/issues/63

### DIFF
--- a/client-cli/src/main/java/ai/floedb/floecat/client/cli/QueryCliSupport.java
+++ b/client-cli/src/main/java/ai/floedb/floecat/client/cli/QueryCliSupport.java
@@ -27,23 +27,29 @@ import ai.floedb.floecat.common.rpc.SnapshotRef;
 import ai.floedb.floecat.execution.rpc.ScanFile;
 import ai.floedb.floecat.query.rpc.BeginQueryRequest;
 import ai.floedb.floecat.query.rpc.BeginQueryResponse;
+import ai.floedb.floecat.query.rpc.DataFile;
+import ai.floedb.floecat.query.rpc.DataFileBatch;
+import ai.floedb.floecat.query.rpc.DeleteFileBatch;
 import ai.floedb.floecat.query.rpc.DescribeInputsRequest;
 import ai.floedb.floecat.query.rpc.DescribeInputsResponse;
 import ai.floedb.floecat.query.rpc.EndQueryRequest;
 import ai.floedb.floecat.query.rpc.EndQueryResponse;
 import ai.floedb.floecat.query.rpc.ExpansionMap;
-import ai.floedb.floecat.query.rpc.FetchScanBundleRequest;
 import ai.floedb.floecat.query.rpc.GetQueryRequest;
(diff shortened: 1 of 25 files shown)


mrouvroy-floe:
@jchauzi-floe Let me know if that matches what you wanted for future EE/SE improvements
Note: This fully replace the old RPC so would liekly break some current integration tests in core ?

jchauzi-floe (approved):

PR #138, opened 2026-02-09
Credential modeling and  secret storage

kwesterfeld-floe:
* Credential modeling and resolver pluggable
  
    - Add a first-class AuthCredential model that covers bearer tokens, OAuth
      client secrets, AWS keys/assume-role, and token exchange variants.
    - Remove the unused `secret_ref` from AuthConfig; this is now AuthCredential
      responsibility.
    - Make secret resolution optional at the API boundary and support inline
      credentials when appropriate.
    - Introduce a secrets manager abstraction with a non-prod implementation and an
      AWS Secrets Manager-backed implementation for binary payloads.
    - Wire credential resolution through the service/reconciler path and
      update catalog connector handling to consume the resolved model.

* Secrets manager in aws
  
   - Implement the aws version of secrets manager in a way that allows for ABAC; optional assume
     role with tags for account-scoped policy.
   - Add doc on how this is done
  
* Normalize CLI credentials and move provider-specific mapping into connectors
  
    - Simplify CLI credential schema to only carry provider; push provider-specific keys through properties
    - Teach Delta’s auth factory to interpret CLI properties directly for Databricks cache usage
    - Add shared AWS profile helper for SigV4 + Glue and wire profile-based creds into Iceberg
    - Refresh CLI docs and tests to reflect the new provider-centric CLI behavior

* Secure auth storage, masking, and validation tests

  - Enforce that connector auth credentials are never persisted and always redacted in API responses
  - Add masking/redaction coverage for token exchange and AWS profile handling
  - Add Databricks auth factory tests and secrets manager tag assertions

* Connector credentials masking

 - Ensure that any AuthCredential is always diverted to secret store
 - Return masked content only for AuthCredential, which is only resolved "in-VM" for connectors
   and reconciliation

### DIFF
--- a/connectors/catalogs/delta/src/main/java/ai/floedb/floecat/connector/delta/uc/impl/DatabricksCliTokenProvider.java
+++ b/connectors/catalogs/delta/src/main/java/ai/floedb/floecat/connector/delta/uc/impl/DatabricksCliTokenProvider.java
@@ -178,6 +178,9 @@ final class DatabricksCliTokenProvider implements AccessTokenProvider {
   }
 
   static long minExpirySeconds(long expiresInSeconds) {
-    return Math.max(TimeUnit.MINUTES.toSeconds(5), expiresInSeconds);
+    if (expiresInSeconds <= 0) {
+      return TimeUnit.MINUTES.toSeconds(5);
+    }
+    return expiresInSeconds;
   }
 }


mcusack-floe (approved):

connectors/catalogs/delta/src/main/java/ai/floedb/floecat/connector/delta/uc/impl/DatabricksCliTokenProvider.java
```diff
@@ -201,7 +177,7 @@ private static Instant parseExpiry(JsonNode tokNode) {
     return sec > 0 ? Instant.now().plusSeconds(sec) : null;
   }
 
-  private static String enc(String s) {
-    return URLEncoder.encode(s, StandardCharsets.UTF_8);
+  static long minExpirySeconds(long expiresInSeconds) {
+    return Math.max(TimeUnit.MINUTES.toSeconds(5), expiresInSeconds);
```
mrouvroy-floe: This seems wrong: it extends expiry to at least 5 minutes even if the server says expires_in=30. 
That makes the cache believe tokens are valid longer than reality so we will send expired access tokens.

kwesterfeld-floe: You're right. The floor should only apply when expires_in is missing or non‑positive. If the server supplies expires_in, we must respect it. I'll change it to use expires_in when >0 and fallback to 5 minutes only when it's absent or <= 0.

Ironically I just changed this from something the AI wrote that set the expiration wrong (it set it to 1 second).

PR #141, opened 2026-02-10
QueryService: support query replay via client query_id and upfront input pinning

mrouvroy-floe:
## QueryService: support query replay via client query_id and upfront input pinning

This PR extends `QueryService.BeginQuery` to support **deterministic query replay** by allowing clients to:
- Provide an optional **client-supplied `query_id`** (duplicates are rejected)
- Optionally supply **`common.QueryInput` records** so lifecycle metadata can be pinned at creation time

When inputs are provided, the lifecycle service now resolves tables/views, pins snapshots, computes view expansions, and resolves governance obligations up front.

### API changes
- `BeginQueryRequest`
  - `optional string query_id`
  - `repeated common.QueryInput inputs`

Both fields are optional; existing clients are unaffected.

### Behavior
- Client-supplied `query_id` must be unique (`ALREADY_EXISTS` on duplicates)
- Server-generated IDs behave as before
- Pinned snapshots, expansions, and obligations are stored on the query context and returned in `QueryDescriptor`

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/query/impl/QueryContextStoreImpl.java
+++ b/service/src/main/java/ai/floedb/floecat/service/query/impl/QueryContextStoreImpl.java
@@ -85,8 +85,15 @@ public class QueryContextStoreImpl implements QueryContextStore {
 
   @Override
   public void put(QueryContext ctx) {
-    // Insert only if absent
-    cache.asMap().compute(ctx.getQueryId(), (k, existing) -> existing != null ? existing : ctx);
+    putIfAbsent(ctx);
+  }
+
+  @Override
+  public boolean putIfAbsent(QueryContext ctx) {
+    return cache
+            .asMap()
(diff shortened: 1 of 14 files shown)


service/src/main/java/ai/floedb/floecat/service/query/impl/QueryContextStoreImpl.java
```diff
@@ -85,8 +85,15 @@ public Optional<QueryContext> get(String queryId) {
 
   @Override
   public void put(QueryContext ctx) {
-    // Insert only if absent
-    cache.asMap().compute(ctx.getQueryId(), (k, existing) -> existing != null ? existing : ctx);
+    putIfAbsent(ctx);
+  }
+
+  @Override
+  public boolean putIfAbsent(QueryContext ctx) {
```
mmenager-floe: Why do you need a separate function? The difference between `put` and `putIfAbsent` isn't supper clear to me.

PR #143, opened 2026-02-11
Centralize dependency/plugin versions in floecat root pom

kwesterfeld-floe:
- Move module-local dependency/plugin versions into floecat/pom.xml
  - Standardize naming to version.* and version.plugin.*
  - Update child POMs to consume root-managed version properties


mrouvroy-floe (approved):
Nice ! Thanks for doing this !

PR #162, opened 2026-02-19
Make telemetry docgen skippable and stabilize doc output resolution

mrouvroy-floe:
## Make telemetry docgen skippable and stabilize doc output resolution  

This makes telemetry doc generation configurable and improves test isolation.

###  **Add Maven skip flag**  
  Introduces `telemetry.docgen.skip` (default: `false`) and wires it to the `exec-maven-plugin`, allowing docgen to be skipped with: ``` mvn -Dtelemetry.docgen.skip=true ```

### **Simplify doc directory resolution**  
`MetricCatalogDocgen` now resolves the output directory in this order:
1. `telemetry.doc.dir` system property (if set)
2. `${user.dir}/docs/telemetry`
3. `docs/telemetry` (fallback)

Fixes: https://github.com/eng-floe/core/issues/556

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,6 +27,7 @@
   <description>Floecat, Root</description>
 
   <properties>
+    <telemetry.docgen.skip>false</telemetry.docgen.skip>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <version.quarkus>3.29.4</version.quarkus>
     <maven.compiler.source>25</maven.compiler.source>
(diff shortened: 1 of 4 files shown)


mrouvroy-floe:
@jchauzi-floe If we merge this, I think we can just set 
`<telemetry.docgen.skip>true</telemetry.docgen.skip>` 
in the root pom of core and that will prevent doc from being built ?

jchauzi-floe:
> @jchauzi-floe If we merge this, I think we can just set `<telemetry.docgen.skip>true</telemetry.docgen.skip>` in the root pom of core and that will prevent doc from being built ?

It's working indeed

PR #166, opened 2026-02-20
Introduce an Arrow Flight transport for system table scans alongside the existing gRPC API.

mrouvroy-floe:
# Introduce an Arrow Flight transport for system table scans alongside the existing gRPC API.

## Summary
This PR adds Arrow Flight transport for system table scans and folds in the follow-up refactors needed to run cleanly with FloeDB producers.

### Transport + registry changes
- Adds a Flight server to the service for system table scanning.
- Extends the system objects registry model with `FlightEndpointRef`.
- Evolves `StorageTableDetails` to:
  - `oneof { path | flight_endpoint }`
- Extends `SystemTableDef` and `SystemTableNode` to support Flight-backed `STORAGE` tables.
- Enforces exactly one of `storagePath` or `flightEndpoint` for storage-backed system tables.
- Emits backend endpoint metadata in `UserObjectsBundle` so workers can uniformly discover access strategy.

### Protocol and base refactor (merged into same PR)
- Unifies protocol on `SystemTableFlightCommand` with:
  - `SystemTableTarget { name | id }`
  - `query_id` support in command/header consistency checks
- Removes static overlay helpers from `SystemTableCommands` (no global registration path).
- Refactors `SystemTableFlightProducerBase` to:
  - use canonical table names as primary routing key,
  - resolve name->id (and optional id->name) via runtime hooks/injected resolver,
  - remove pseudo-resource-id behavior,
  - support name-only producers (no overlay requirement for external/core modules).
- Makes routing checks deterministic/cheap (`supportsDescriptor` / `supportsTicket`), pushing deep validation to execution path.
- Keeps streaming lifecycle robust (worker-thread execution, guarded cleanup, cancellation-safe behavior).

### Shared execution/streaming
- Shares inbound auth, engine context, correlation ID, and query ID handling between gRPC and Flight.
- Unifies Arrow batch streaming via shared `ArrowBatchSerializer` used by both transports.
- Adds/updates scan-plan utilities to support reusable record-to-arrow paths and paged streaming scenarios.

### Outcome
Workers can use a single Arrow Flight-based algorithm for system table scans when a `flight_endpoint` is present, while non-Floecat/external producers can still operate name-first without internal id coupling.

## Type of Change

- [ ] fix (bug fix)

## Testing

Validated via:

- New integration test `SystemTableFlightIT` exercising:
  - `getFlightInfo`
  - `getStream`
  - projection and schema consistency
- New unit tests for `ArrowBatchSerializer`
- Updated existing catalog/registry tests to cover new validation rules
- Full test suite (`make verify`) including service-spawned ITs

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

### Config Changes

New configuration properties:

- `floecat.flight.port`
- `floecat.flight.host`
- `floecat.flight.tls`
- `floecat.flight.memory.max-bytes`

For integration tests that spawn the service JVM, `--add-opens=java.base/java.nio=ALL-UNNAMED` is required due to Apache Arrow direct memory access on newer JDKs.

## Checklist

## Additional Notes

- Flight and gRPC now share a single inbound context extraction path to avoid drift in auth/query/correlation handling.
- STORAGE system tables must now define exactly one of `storagePath` or `flightEndpoint`; validation enforces this.
- Follow-up work may include:
       - TLS support for Flight and 
       - worker-side full Flight scan-node adoption.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -23,6 +23,7 @@ services:
     environment:
       FLOECAT_HTTP_HOST: "0.0.0.0"
       FLOECAT_HTTP_PORT: "9100"
+      JAVA_TOOL_OPTIONS: "--add-opens=java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED"
     networks:
       - floecat
     volumes:


core/proto/src/main/proto/floecat/query/system_objects_registry.proto
```diff
@@ -108,16 +108,48 @@ message SystemTable {
...
+ *
+ * Workers MUST NOT use the registry ticket field; tickets for actual scans
+ * come from getFlightInfo. The ticket field is reserved for future use
+ * (e.g., optional routing hint from STORAGE services).
+ */
+message FlightEndpointRef {
+  /** Host reachable from workers. Must be externally routable (never localhost in prod). */
+  string host = 1;
+  /** Port of the Arrow Flight server. */
+  int32 port = 2;
```
kwesterfeld-floe: I do think you did the right thing by using a binary-encoded command, which allows for projection and filter/pushdown.  And here you sort of forbid the encoding of "ticket", which would align with the simpler approach that I went with a that does use a fixed ticket model, like `sys.session`, where the intent is to never need to do anything other than retrieve the entire flight.   In this case, the caller and service can know a priori what the ticket agreement is.  

Perhaps this is naive, and the adoption of the command binary message pattern should be generalized and employable as library code that floe core can reuse.  There are currently 6 flights that are possible in floe core, so thinking about that sys.session and sys.query come to mind as needing projection and pushdown potentially.  It would be great to quickly bring those online.

PR #170, opened 2026-02-21
Delta-iceberg emulation

mcusack-floe:
## Summary

Adds an emulation layer to allow delta tables to be queried via the Iceberg REST catalog api. When DuckDB or some other client tries querying a delta table registered with Floecat, the Iceberg service converts the delta metadata and deletion vectors on the fly into the equivalent Iceberg metadata, manifest and manifest lists and generates position delete files.

This also involves a significant refactor of the iceberg-rest interface to remove brittle hand-crafting of iceberg metadata and rely on the Apache Iceberg library to do the metadata artifact creation.

The feature is backed by end-to-end smoke tests that start up Floecat, load real iceberg and delta tables into localstack, and then use DuckDB to query these tables via the Iceberg REST API.

## Type of Change

- [ ] fix (bug fix)
- [X] refactor (no behavior change)
- [X] test (adds/updates tests)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,7 +142,7 @@ jobs:
           make test-localstack 2>&1 | tee "$MVN_LOG"
 
       - name: Upload Test Artifacts
-        if: always()
+        if: ${{ failure() || cancelled() }}
         uses: actions/upload-artifact@v4
         with:
           name: ci-localstack-test-artifacts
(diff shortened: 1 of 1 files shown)


kwesterfeld-floe (approved):
This looks very thorough.   Mostly the bulk is tests, so great job getting that coverage!

In general the synchronous model here for a gateway is likely fine for performance, as everything is going to be blocked on remote I/O.  But also for that very reason, if we grow the capabilities to handle large numbers of connected clients, we will likely want to extensively convert to async/mutiny.  It will be painful at that time to do so, or at least time consuming.  Otoh, maybe the agents will eat that task right up and it won't be such a big deal.

PR #248, opened 2026-03-31
Stats system: target-native architecture refactor and pluggable engine framework 

mrouvroy-floe:
## Summary

This PR delivers the target-native storage and planner resolution and the pluggable engine framework 

The idea: instead of storing stats in flat column-keyed rows, every stats record is a TargetStatsRecord, a proto-encoded entity keyed by (table_id, snapshot_id, target) where target is one of TABLE, COLUMN, EXPRESSION, or FILE. 
This makes the storage model, the planner API, the reconciler write path, and the future engine dispatch layer all speak the same language.

### Proto schema (stats.proto, planner_stats_bundle.proto)
- New TargetStatsRecord envelope: 
```proto
oneof value { TableValueStats, ScalarStats, FileTargetStats }
```
Expression targets store stats as ScalarStats: identity is carried by the target, not the value type.

- StatsMetadata carries: 
      - StatsProducer (SOURCE_NATIVE, UPSTREAM_METADATA_DERIVED, ENGINE_COMPUTED), 
      - StatsCompleteness (COMPLETE, PARTIAL, UNAVAILABLE), 
      - StatsCaptureMode (SYNC, ASYNC), 
      - confidence_level,
      - StatsCoverage (rows/files/row-groups/bytes scanned), 
      - captured_at, 
      - refreshed_at.

EngineExpressionStatsTarget: identity is engine_kind + opaque engine_expression_key bytes. engine_version is deliberately excluded from identity so stats survive engine upgrades.

### Core SPI module (core/stats)
- New Maven artifact floecat-core-stats; OSS engine authors take this as their only dependency.
- StatsCaptureEngine : pluggable SPI. Implement + @ApplicationScoped = auto-discovered by the registry. 
Five methods: id(), priority() (lower wins; default 100), capabilities(), supports(), capture().
- StatsStore:  pluggable SPI for persistence 

### Target-native internal refactor
All internal layers upadated to use the new target-native stats type : connectors, reconciler, planner, etc...

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [ ] No breaking changes


service/src/main/java/ai/floedb/floecat/service/statistics/engine/impl/PersistedStatsCaptureEngine.java
```diff
@@ -16,61 +16,64 @@
...
-      StatsEngineCapabilities.builder()
+          StatsKind.ROW_COUNT,
+          StatsKind.NULL_COUNT,
+          StatsKind.NDV,
+          StatsKind.MIN_MAX,
+          StatsKind.HISTOGRAM);
+  private static final StatsCapabilities CAPABILITIES =
+      StatsCapabilities.builder()
           .connectors(Set.of()) // empty = all connectors
           .targetTypes(Set.of(StatsTargetType.TABLE, StatsTargetType.COLUMN, StatsTargetType.FILE))
```
mcusack-floe, line 61: We don't advertise that we can serve EXPRESSION, yet we can capture it?

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/DirectReconcilerBackend.java
```diff
@@ -349,50 +347,21 @@ public void ingestSnapshot(ReconcileContext ctx, ResourceId tableId, Snapshot sn
...
-    for (ColumnStats columnStats : stats) {
-      StatsGroupKey key = new StatsGroupKey(columnStats.getTableId(), columnStats.getSnapshotId());
-      grouped.computeIfAbsent(key, k -> new ArrayList<>()).add(columnStats);
-    }
-    for (var entry : grouped.entrySet()) {
-      statsRepository.putColumnStats(
-          entry.getKey().tableId, entry.getKey().snapshotId, entry.getValue());
-    }
+    return statsStore
+        .getTargetStats(tableId, snapshotId, StatsTargetIdentity.tableTarget())
```
mcusack-floe: This seems to only detect that the table is present. It's not testing whether their are any table, column, file or expression stats actually there, which means backfilling stats will never work.

service/src/main/java/ai/floedb/floecat/service/query/catalog/StatsProviderFactory.java
```diff
@@ -104,9 +137,29 @@ private Optional<StatsProvider.TableStatsView> safeTableStats(
...
+        var request =
+            new StatsCaptureRequest(
+                tableId,
+                snapshotId,
+                StatsTargetIdentity.columnTarget(columnId),
+                // Empty requested kinds means "accept any available stat family".
+                Set.of(),
+                Set.of(),
+                StatsExecutionMode.SYNC,
+                // Empty connector type means no connector-based engine filtering.
```
mcusack-floe: Is not setting a connector type dangerous? We could end up selecting engines for delta and iceberg

service/src/main/java/ai/floedb/floecat/service/statistics/impl/TableStatisticsServiceImpl.java
```diff
@@ -508,94 +263,43 @@ private Boolean processColumnStats(
...
-      var columnStats =
-          raw.toBuilder().setTableId(next.tableId()).setSnapshotId(next.snapshotId()).build();
-      validateStatsMetadata(columnStats.hasMetadata() ? columnStats.getMetadata() : null);
-      var fingerprint = canonicalFingerprint(columnStats);
-
-      if (next.idempotencyKey() == null) {
-        stats.putColumnStats(next.tableId(), next.snapshotId(), columnStats);
-        upserted.incrementAndGet();
-        continue;
+    for (var raw : req.getRecordsList()) {
```
mcusack-floe, line 266: We should validate full key/payload identity consistency before persisting this record.

core/connectors/common/src/main/java/ai/floedb/floecat/connector/common/resolver/StatsProtoEmitter.java
```diff
@@ -129,91 +165,150 @@ private static long resolveId(
...
+  }
 
-    var upstream =
-        LogicalTypeProtoAdapter.upstreamStamp(
-            toTableFormat(format),
-            "",
-            Long.toString(snapshotId),
-            Timestamps.fromMillis(Math.max(0L, upstreamCreatedAtMs)),
-            Map.of());
+  public static List<TargetStatsRecord> toTargetColumnStats(
```
mcusack-floe, line 180: The connector conversion helper dropped upstream provenance on column targets. toTargetColumnStats() no longer sets ScalarStats.upstream, while file-column scalars still do. That is a regression from the old emitter behavior and makes column targets less auditable than file targets.

service/src/main/java/ai/floedb/floecat/service/statistics/engine/impl/PersistedStatsCaptureEngine.java
mrouvroy-floe: Expression is intentionally unsupported in PersistedStatsCaptureEngine for now, as we can't compute stats for those yet

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/DirectReconcilerBackend.java
mrouvroy-floe: Fixed: 
- statsAlreadyCapturedForTargetKind(...) maps StatsTargetKind to StatsTargetType
- then calls statsStore.countTargetStats(tableId, snapshotId, targetType) > 0
So backfill detection is target-aware now.

mcusack-floe (approved):
General comment on this and the other two PRs is to exercise caution when merging with main, particularly in ReconcilerService and associated classes. Also, we need to accurately determine what stats to backfill and how we determine that stats for a snapshot are missing/incomplete.

PR #250, opened 2026-04-01
Source-native OSS stats capture through the orchestrator + engine registry path

mrouvroy-floe:
## Summary

This delivers introduces Iceberg/Delta native capture engines, wires query-time and reconciler `STATS_ONLY` capture through the shared control plane, and keeps stats persistence centralized via `StatsStore`.

### Core flow
- Added orchestrator-first resolution path:
  - store-first lookup
  - SYNC inline capture attempt
  - ASYNC enqueue fallback when still missing and capturable
- Routed explicit capture through `StatsCaptureControlPlane` with `StatsOrchestrator` implementation.

### Native engines
- Added native OSS engines:
  - `IcebergNativeStatsCaptureEngine`
  - `DeltaNativeStatsCaptureEngine`
- Engines capture source-native stats and persist through the pluggable `StatsStore`.
- Added metadata normalization defaults for native captured records.

### Reconciler integration
- `STATS_ONLY` path now routes through control-plane / engine registry.
- Selector scope is propagated through capture requests.
- Reconciler cleanup:
  - clarified `statsProcessed` semantics
  - removed dead code paths

### API/safety improvements
- Added safer `StatsCaptureRequest.builder(...)` to avoid fragile positional constructor usage.
- Migrated production call sites and tests to builder-based construction.
- Removed redundant sync-miss double registry lookup in `StatsOrchestrator.resolve()`.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist

### Follow-ups: 
  - bounded sync wait/re-read
  - target-granular async follow-up
  - metadata+stats decoupling policy change (metadata now + queued stats only)


reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReconcilerService.java
```diff
@@ -347,6 +354,38 @@ public Result reconcile(
                   snapshotId ->
                       isStatsCaptureCompleteForScope(
```
mcusack-floe: There is an issue in isStatsCaptureCompleteForScope, which called statsAlreadyCapturedForTargetKind(..., STK_COLUMN). This latter function only answers “does any column stat exist for this snapshot?”, it does not answer “do stats exist for all requested selectors in includeSelectors?” This can cause us to skip backfilling missing stats.

Also, this function is now called multiple times per snapshot, instead of only once as it was before which is inefficient.

mcusack-floe (approved):
General comment on this and the other two PRs is to exercise caution when merging with main, particularly in ReconcilerService and associated classes. Also, we need to accurately determine what stats to backfill and how we determine that stats for a snapshot are missing/incomplete.

PR #256, opened 2026-04-15
Batch-First Stats Capture End-to-End (Orchestrator, Reconciler, Engines, SPI)

mrouvroy-floe:
## Summary 

This PR moves Floecat stats capture to a batch-first model end-to-end and removes legacy single-item routing in production paths.

## Why
- Reduce repeated connector/file opens for multi-target requests.
- Make stats outcomes explicit per requested target.
- Align sync/async behavior across planner resolution, orchestrator routing, and reconciler follow-up.
- Improve OSS maintainability by simplifying API surface and tightening contracts.

### Core stats SPI
- Added/standardized batch-native types:
  - `StatsCaptureBatchRequest`
  - `StatsCaptureBatchResult`
  - `StatsCaptureBatchItemResult`
  - explicit `StatsTriggerOutcome` usage per item
- Tightened validation/normalization in request/result models.
- Removed dead single-item trigger surfaces where no longer used in runtime flow.
- Clarified capability semantics and connector normalization behavior.

### Engine registry and routing
- `StatsEngineRegistry` now routes capture through batch logic as the canonical path.
- Preserves per-item fallback behavior on `UNCAPTURABLE` across candidate engines.
- Enforces batch invariants (size/order) and per-item outcome mapping.

### Orchestrator
- Batch-first capture/resolve flows are now primary.
- Single-item calls are wrappers (or removed where obsolete) around batch behavior.
- Added grouped async enqueue behavior for unresolved/degraded items.
- Added structured outcome logging and batch-aware metrics.

### Reconciler
- STATS_ONLY flow submits true batch capture requests per grouped snapshot scope.
- Follow-up jobs carry unresolved target scope (not table-wide-only intent).
- Improved partial/degraded semantics and warning surfaces.
- Cleaned dead paths and aligned behavior between fast-path and connector-assisted paths.

### Native engines
- Iceberg and Delta native engines implement true batch capture overrides.
- One connector open / grouped processing per `(table,snapshot)` where possible.
- Maintains snapshot `0` support semantics for Delta.
- Connector-backed capture hints aligned with target-kind batching.

### Docs and contracts
- Updated SPI and architecture docs for batch-first routing.
- Documented targeted async follow-up contract.
- Clarified engine implementor expectations for batch behavior and outcomes.

## Behavior changes
- Batch capture is now the production default path.
- Explicit per-item outcomes are first-class (`CAPTURED`, `QUEUED`, `UNCAPTURABLE`, `DEGRADED`).
- Async follow-up can be targeted to unresolved items, reducing redundant recapture work.
- Legacy selector-heavy internal paths were removed/refactored in favor of explicit targets.

## Observability
- Added/updated batch-oriented metrics and structured logs:
  - item and group counts
  - engine batch call counts
  - outcome distribution
  - store-hit ratio tracking
- Unified outcome logging fields for easier ops/debugging.

## Tests
- Expanded unit/integration coverage across:
  - batch model validation
  - registry fallback/routing behavior
  - orchestrator mixed-outcome batches
  - reconciler targeted follow-up behavior
  - native engine batch correctness (including Delta snapshot `0`)
- Added contract-style tests for non-captured outcomes and degraded follow-up semantics.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`


reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReconcilerService.java
```diff
@@ -1151,9 +1196,16 @@ private Optional<String> enqueueStatsOnlyCapture(
...
-    ReconcileScope scope = ReconcileScope.of(namespacePaths, tableDisplayName, columns);
+    List<Long> sortedSnapshotIds =
+        snapshotIds == null ? List.of() : snapshotIds.stream().sorted().toList();
+    List<String> sortedTargets =
+        targetSpecs == null ? List.of() : targetSpecs.stream().sorted().toList();
+    ReconcileScope scope = ReconcileScope.of(namespacePaths, tableDisplayName, List.of());
+    if (!sortedSnapshotIds.isEmpty() || !sortedTargets.isEmpty()) {
+      scope =
+          ReconcileScope.of(
+              namespacePaths, tableDisplayName, List.of(), sortedSnapshotIds, sortedTargets);
```
mcusack-floe: List.of() means we're losing a scoped set of columns to fetch?

service/src/main/java/ai/floedb/floecat/service/statistics/StatsOrchestrator.java
```diff
@@ -135,22 +255,64 @@ private void tryEnqueueAsyncCapture(StatsCaptureRequest request) {
...
+      LinkedHashSet<String> statsTargets = new LinkedHashSet<>();
+      for (StatsCaptureRequest request : groupedRequests) {
+        snapshotIds.add(request.snapshotId());
+        statsTargets.add(StatsTargetScopeCodec.encode(request.target()));
+      }
+      ReconcileScope scope =
+          ReconcileScope.of(
+              namespaceScope,
+              tableDisplay,
+              List.of(),
```
mcusack-floe: List.of() means we're losing a scoped set of columns to fetch?

service/src/main/java/ai/floedb/floecat/service/statistics/StatsOrchestrator.java
```diff
@@ -79,52 +105,146 @@ public StatsOrchestrator(
...
+    for (StatsCaptureRequest request : requests) {
+      if (request.snapshotId() < 0L) {
+        recordAsyncSkip(
+            request,
+            "invalid_snapshot_id",
+            "Skipping async enqueue because snapshot id is invalid");
+        continue;
+      }
+      StatsCaptureRequest asyncRequest = toAsyncRequest(request);
+      AsyncCandidateKey candidateKey = AsyncCandidateKey.from(asyncRequest);
```
mcusack-floe: AsyncCandidate key only includes connectorType, targetCase, executionMode. But capability matching also depends on requestedKinds() and samplingRequested().

mcusack-floe (changes requested):

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReconcilerService.java
mrouvroy-floe: fixed

service/src/main/java/ai/floedb/floecat/service/statistics/StatsOrchestrator.java
mrouvroy-floe: fixed

service/src/main/java/ai/floedb/floecat/service/statistics/StatsOrchestrator.java
mrouvroy-floe: done

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReconcilerService.java
mcusack-floe, line 2062: STATS_ONLY completeness is checked against existing selector-based coverage, not the explicit requested stats targets. That means we can skip capture when unrelated column stats already exist, even though the requested target is still missing.

The completion check here should use the explicit scoped stats targets when destinationStatsTargets is present, rather than relying only on selector-based coverage. In practice, that means STATS_ONLY should only skip a snapshot if all requested targets for that snapshot are already captured; otherwise it should continue building capture requests for the missing targets.

service/src/main/java/ai/floedb/floecat/service/statistics/StatsOrchestrator.java
```diff
@@ -79,52 +106,146 @@ public StatsOrchestrator(
...
 
-  private void maybeEnqueue(StatsCaptureRequest request, boolean enqueueEligible) {
-    if (!enqueueEligible) {
+  private void tryEnqueueAsyncCaptureBatch(List<StatsCaptureRequest> requests) {
+    if (requests == null || requests.isEmpty()) {
       return;
     }
-    tryEnqueueAsyncCapture(request);
+    Map<TableKey, List<StatsCaptureRequest>> groupedByTable = new LinkedHashMap<>();
+    Map<AsyncCandidateKey, Boolean> asyncCandidateCache = new LinkedHashMap<>();
```
mcusack-floe: The async enqueue candidate cache is keyed too broadly here. StatsCaptureEngine.supports() is allowed to make dynamic, request-specific decisions, but the cache only keys on connector type, target case, mode, requested kinds, and sampling. That means a support result from one request can be reused for a different table or target where the answer should be different, causing us to either skip enqueue for work that is actually supported or enqueue work that no engine can handle.

The fix is to stop reusing candidate decisions across requests unless the cache key includes the full set of fields that can affect supports(). The safest change is to evaluate async candidates per request with no shared cache; if caching is still wanted, the key needs to include enough request identity to preserve correctness, such as table-specific context and any other fields that engines can inspect dynamically.

mcusack-floe (changes requested):

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReconcilerService.java
mrouvroy-floe: Fixed.

STATS_ONLY completeness is now target-aware when `destinationStatsTargets` is present:
`isStatsCaptureCompleteForScope(...)` first checks `backend.statsCapturedForTargets(...)` and only skips when all requested targets for that snapshot are already captured.
If a TABLE target is requested with scoped selectors, we additionally require selector coverage.

service/src/main/java/ai/floedb/floecat/service/statistics/StatsOrchestrator.java
mrouvroy-floe: Fixed.

Removed cross-request async candidate memoization. Candidate support is now evaluated per request (`statsEngineRegistry.candidates(asyncRequest)`) with no shared cache key reuse.

So support decisions are no longer reused across different tables/targets/requests.

service/src/main/java/ai/floedb/floecat/service/statistics/StatsOrchestrator.java
```diff
@@ -157,23 +281,66 @@ private void tryEnqueueAsyncCapture(StatsCaptureRequest request) {
           table.get().getUpstream().getTableDisplayName().isBlank()
               ? table.get().getDisplayName()
               : table.get().getUpstream().getTableDisplayName();
-      ReconcileScope scope = ReconcileScope.of(namespaceScope, tableDisplay, List.of());
+      LinkedHashSet<Long> snapshotIds = new LinkedHashSet<>();
```
mcusack-floe: This loses the original snapshot-to-target pairing. If groupedRequests contains (s42,tA) and (s43,tB), the follow-up scope stores snapshotIds=[42,43] and statsTargets=[tA,tB], and the reconciler later rebuilds (42,tA) (42,tB) (43,tA) (43,tB). With schema evolution, that can enqueue invalid target/snapshot pairs and fail jobs. We need to preserve per-request identity in scope instead of storing snapshots and targets as separate lists.

mcusack-floe (approved):
I'm approving, but note the comment about unioning stats targets, which could lead to invalid capture jobs, particularly over schema changes.

PR #257, opened 2026-04-15
Treat DynamoDB TransactionConflict as retryable in txnWriteCas

malmeida-floe:
DynamoDB returns TransactionConflict when concurrent transactions touch the same item (e.g. RESET ALL racing with updateLastActivity). Previously only ConditionalCheckFailed was treated as a CAS miss (returning false); TransactionConflict was rethrown as an unrecoverable XX000/UNKNOWN error.

Now both reasons return false from txnWriteCas, which lets SessionCache.updateWithRetry handle the conflict: it refreshes the session from DynamoDB and retries the mutation (up to 3 attempts by default, configured via app.session.update-retries).

Fixes: eng-floe/core#1060

## Testing

- Tests executed against staging environment are now passing.

- [X] No breaking changes

### DIFF
--- a/storage/aws/src/main/java/ai/floedb/floecat/storage/kv/dynamodb/DynamoDbKvStore.java
+++ b/storage/aws/src/main/java/ai/floedb/floecat/storage/kv/dynamodb/DynamoDbKvStore.java
@@ -398,14 +398,25 @@ public final class DynamoDbKvStore implements KvStore, KvAttributes {
         .onFailure(TransactionCanceledException.class)
         .recoverWithItem(
             t -> {
-              // Return false only for conditional check failure; otherwise rethrow.
+              // Return false for retryable conflicts; throw on unexpected reasons.
               TransactionCanceledException ex = (TransactionCanceledException) t;
+              boolean sawRetryable = false;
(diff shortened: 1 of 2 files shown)


mrouvroy-floe (approved):

mcusack-floe (approved):
I think this is ok, but we might be swallowing none retryable bugs or other issues outside of CAS updates: https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/dynamodbv2/model/TransactionCanceledException.html

We should probably loop over all the reasons, check that the reason is not None, and also check whether the reason is TransactionConflict or Conditional check failed and set sawRetryable = true and continue if so. Then at the end of the loop return false if sawRetryable otherwise throw the exception. Inside the loop, we throw an exception if we  didn't continue.

malmeida-floe:
@mcusack-floe Good point. I changed the PR so that we throw if we don't consider the error retryable.

mcusack-floe (approved):

mcusack-floe:
> @mcusack-floe Good point. I changed the PR so that we throw if we don't consider the error retryable.

Looks good.

PR #264, opened 2026-04-17
(fix): ensure account_id is propagated.

asaoutkin-floe:
## Summary
The old code used deferred(() -> grpcCtx.call(...)) which attached the gRPC context only during the supplier execution. The lazy Multi returned by bundles.stream() was then consumed after the context detached, so principal.get() on the worker thread returned PrincipalContext.getDefaultInstance() with empty accountId. The fix uses an emitter that keeps the gRPC context attached via grpcCtx.run() for the entire subscription lifecycle — the bundles.stream() items are emitted inside the run block, so principal.get() always finds the correct context.

## Type of Change

- [x] fix (bug fix)
## Testing

Unit test added that demonstrates the issue. `make test-e2e` in private repo was failing previously before this fix. Now passes with the fix.

- [x] `make fmt`
- [ ] `make verify`

- [x] No breaking changes

## Checklist

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/query/impl/UserObjectsServiceImpl.java
+++ b/service/src/main/java/ai/floedb/floecat/service/query/impl/UserObjectsServiceImpl.java
@@ -87,25 +87,29 @@ public class UserObjectsServiceImpl extends BaseServiceImpl implements UserObjec
                     String queryId = mustNonEmpty(request.getQueryId(), "query_id", correlationId);
                     var ctxOpt = queryStore.get(queryId);
                     if (ctxOpt.isEmpty()) {
-                      throw GrpcErrors.notFound(
-                          correlationId, QUERY_NOT_FOUND, Map.of("query_id", queryId));
+                      emitter.fail(
+                          GrpcErrors.notFound(
+                              correlationId, QUERY_NOT_FOUND, Map.of("query_id", queryId)));
+                      return;
                     }
 
                     QueryContext ctx = ctxOpt.get();
                     if (!ctx.isActive()) {
-                      throw GrpcErrors.preconditionFailed(
-                          correlationId,
-                          QUERY_NOT_ACTIVE,
-                          Map.of("query_id", queryId, "state", ctx.getState().name()));
+                      emitter.fail(
+                          GrpcErrors.preconditionFailed(
+                              correlationId,
+                              QUERY_NOT_ACTIVE,
+                              Map.of("query_id", queryId, "state", ctx.getState().name())));
+                      return;
                     }
 
-                    try {
-                      bundles.stream(correlationId, ctx, request.getTablesList())
-                          .subscribe()
-                          .with(emitter::emit, emitter::fail, emitter::complete);
-                    } catch (Throwable t) {
-                      emitter.fail(t);
-                    }
+                    var subscription =
+                        bundles.stream(correlationId, ctx, request.getTablesList())
+                            .subscribe()
+                            .with(emitter::emit, emitter::fail, emitter::complete);
+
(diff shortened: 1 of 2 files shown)


mcusack-floe (changes requested):
Disregard

service/src/main/java/ai/floedb/floecat/service/query/impl/UserObjectsServiceImpl.java
```diff
@@ -64,43 +64,50 @@ public Multi<UserObjectsBundleChunk> getUserObjects(GetUserObjectsRequest reques
...
+                    if (!ctx.isActive()) {
+                      throw GrpcErrors.preconditionFailed(
+                          correlationId,
+                          QUERY_NOT_ACTIVE,
+                          Map.of("query_id", queryId, "state", ctx.getState().name()));
+                    }
 
-                      return bundles.stream(correlationId, ctx, request.getTablesList());
-                    }))
+                    try {
```
mcusack-floe: We currently subscribe to bundles.stream(...) and discard the returned Cancellable, so the inner stream is not tied to the outer gRPC stream lifecycle. If the client cancels, the inner bundle stream can continue running. Capture the subscription and cancel it from the emitter termination path. Also handle request-validation failures with emitter.fail(...) inside grpcCtx.run(...) rather than throwing and relying on outer exception propagation. Please add a service-level cancellation test for UserObjectsServiceImpl so this path is covered end-to-end.

Fix is something like:

```
return Multi.createFrom()
    .<UserObjectsBundleChunk>emitter(
        emitter -> {
          grpcCtx.run(
              () -> {
                workStartNs.compareAndSet(0L, System.nanoTime());
                var principalContext = principal.get();
                var contextCorrelationId = InboundContextInterceptor.CORR_KEY.get();
                var principalCorrelationId = principalContext.getCorrelationId();
                var correlationId =
                    contextCorrelationId != null && !contextCorrelationId.isBlank()
                        ? contextCorrelationId
                        : principalCorrelationId;
                correlationRef.set(correlationId == null ? "" : correlationId);
                authz.require(principalContext, "catalog.read");

                String queryId = mustNonEmpty(request.getQueryId(), "query_id", correlationId);
                var ctxOpt = queryStore.get(queryId);
                if (ctxOpt.isEmpty()) {
                  emitter.fail(
                      GrpcErrors.notFound(
                          correlationId, QUERY_NOT_FOUND, Map.of("query_id", queryId)));
                  return;
                }

                QueryContext ctx = ctxOpt.get();
                if (!ctx.isActive()) {
                  emitter.fail(
                      GrpcErrors.preconditionFailed(
                          correlationId,
                          QUERY_NOT_ACTIVE,
                          Map.of("query_id", queryId, "state", ctx.getState().name())));
                  return;
                }

                var subscription =
                    bundles.stream(correlationId, ctx, request.getTablesList())
                        .subscribe()
                        .with(emitter::emit, emitter::fail, emitter::complete);

                emitter.onTermination(subscription::cancel);
                emitter.onCancellation(subscription::cancel);
              });
        })
    .runSubscriptionOn(Infrastructure.getDefaultExecutor());
```

mcusack-floe (changes requested):

mrouvroy-floe:
LGTM. But do we need a similar fix in PlannerStatsServiceImpl RPC maybe ?

mrouvroy-floe (approved):

mcusack-floe (approved):

asaoutkin-floe:
> LGTM. But do we need a similar fix in PlannerStatsServiceImpl RPC maybe ?

Haven't checked, but I'll do so out of courtesy this week and will get back to this thread and/or open a PR as applicable.

PR #267, opened 2026-04-17
[PERF] Improve GetUserObjects RPC latency and debugability - part 2

mrouvroy-floe:
## Summary

This PR improves GetUserObjects latency and debugability by:
- making slow-path timing visible in logs (not only traces),
- reducing repeated resolver/catalog work,
- removing synchronous advisory writes from the hot path,
- tightening Iceberg stats/caching behavior.

It also includes small correctness/cleanup adjustments discovered during perf iteration.

## Refactor RPC/query timing observability with phase spans
Refactors timing into explicit phase spans/events and adds slow-call log breakdown so phase timings remain available even without a trace backend.

## Lazily resolve default catalog and skip not-found query fetch
Avoids unnecessary default-catalog resolution and short-circuits not-found query fetch paths to reduce avoidable overhead.

## Type lookup: split layered resolution and direct system lookup
Separates system/user type lookup paths to reduce resolver initialization cost and make lookup behavior clearer.

## TypeResolver: enforce context-memoized construction
Forces resolver creation through context memoization to prevent repeated expensive resolver setup within the same request context.

## Iceberg stats: reuse first scan for delete-file stats
Reuses scan results so delete-file stats are not collected via a second pass.

## Iceberg connector: pool REST catalogs across captures
Adds pooled REST catalog reuse (plus cleanup like dead guard removal in rest props) to reduce repeated auth/init overhead.

## Iceberg connector: cache table loads and remove global catalog lock
Caches table loads and removes global lock contention in catalog acquisition to improve concurrency and cold-path behavior.

## Engine hints: invalidate graph cache after persistence
Invalidates relevant graph cache entries when hints are persisted to avoid stale reads right after hint writes.

## Stats capture: async bounded persistence with in-flight dedupe
Moves stats persistence off the request hot path with bounded async queue and in-flight dedupe; adds better async persistence observability.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist


service/src/main/java/ai/floedb/floecat/service/statistics/engine/impl/AbstractNativeStatsCaptureEngine.java
```diff
@@ -221,16 +241,143 @@ record ->
...
-  private void persistRecord(TargetStatsRecord record) {
+  private PersistSchedule enqueueStatsPersistence(List<TargetStatsRecord> records) {
+    if (records == null || records.isEmpty()) {
+      return PersistSchedule.EMPTY;
+    }
+    long startNs = System.nanoTime();
+    int dedupedCount = 0;
+    List<PersistEntry> accepted = new ArrayList<>();
+    for (TargetStatsRecord record : records) {
+      String storageId = StatsTargetIdentity.storageId(record.getTarget());
```
mcusack-floe: This needs to dedupe jobs based on target tableId and SnapshotId too.

connectors/catalogs/iceberg/src/main/java/ai/floedb/floecat/connector/iceberg/impl/IcebergConnectorFactory.java
```diff
@@ -95,25 +101,84 @@ yield new IcebergFilesystemConnector(
...
+            tableCatalog,
+            ndvEnabled,
+            ndvSampleFraction,
+            ndvMaxFiles,
+            false);
       }
     };
   }
 
+  private static CatalogCacheEntry acquireRestCatalog(Map<String, String> props) {
```
mcusack-floe: In IcebergConnectorFactory.java acquireRestCatalog() returns a shared cached RESTCatalog instance. Then connectors are constructed with closeCatalogOnClose=false, so they do not own that catalog lifetime anymore. But IcebergConnectorFactory.java and IcebergConnectorFactory.java can still close the shared catalog during refresh/eviction based only on TTL.

Problem is specifically that acquireRestCatalog() hands out a shared live object without any usage tracking, while eviction can close that same object concurrently, which makes the lifetime management unsafe.

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -1354,48 +1540,226 @@ private UserObjectsBundleChunk flushResolutionChunk() {
       return resolutionsChunk(ctx.getQueryId(), seq++, resolutions);
     }
 
-    private void logStreamTiming() {
+    private void publishStreamTelemetry(String outcome) {
```
mcusack-floe, line 1553: publishStreamTelemetry() calls emitPhaseSpans(totalNanos, schedulingNanos, outcome). Inside this flow, nested spans like decoration_relation and decoration_columns use offsets such as decorationCursorNs, which are relative to the start of the decoration phase, not relative to the whole request.

But UserObjectBundleService.java startPhaseSpan() always does:

startTimestampNs = streamStartEpochNs + safeStartOffset
So every offset is interpreted as “from stream start”, even for nested spans. That makes nested spans start too early unless their parent already began at offset 0.

mcusack-floe (changes requested):

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
mrouvroy-floe: Fixed

connectors/catalogs/iceberg/src/main/java/ai/floedb/floecat/connector/iceberg/impl/IcebergConnectorFactory.java
mrouvroy-floe: Added refcount so we know about lifecycle

service/src/main/java/ai/floedb/floecat/service/statistics/engine/impl/AbstractNativeStatsCaptureEngine.java
mrouvroy-floe: done

service/src/main/java/ai/floedb/floecat/service/statistics/engine/impl/AbstractNativeStatsCaptureEngine.java
```diff
@@ -142,18 +157,24 @@ public Optional<StatsCaptureResult> capture(StatsCaptureRequest request) {
...
                     : withCanonicalMetadata(bundleRecord, request);
-            String storageId = StatsTargetIdentity.storageId(toPersist.getTarget());
-            if (seenTargets.add(storageId)) {
-              persistRecord(toPersist);
+            String persistKey = persistInFlightKey(toPersist);
+            if (seenTargets.add(persistKey)) {
+              persistCandidates.add(toPersist);
             }
           }
+          PersistSchedule schedule = enqueueStatsPersistence(persistCandidates);
```
mcusack-floe, line 171: This changes SYNC capture from inline persistence to executor-backed persistence. On main, successful sync capture called persistRecord(...) before returning; now we enqueue and return immediately, so StatsStore may still be empty after a successful sync capture. That seems to break the read-after-write behavior StatsOrchestrator.resolve() relies on for sync mode.

service/src/main/java/ai/floedb/floecat/service/statistics/engine/impl/AbstractNativeStatsCaptureEngine.java
```diff
@@ -221,18 +242,160 @@ record ->
...
+      if (!STATS_PERSIST_IN_FLIGHT.add(persistKey)) {
+        dedupedCount++;
+        continue;
+      }
+      accepted.add(new PersistEntry(persistKey, record));
+    }
+    if (accepted.isEmpty()) {
+      return new PersistSchedule(System.nanoTime() - startNs, 0, dedupedCount);
+    }
+    try {
```
mcusack-floe, line 269: This can silently drop captured stats under load. If the single-threaded executor’s queue is full, execute(...) throws, we log "Dropping async stats persistence" below, and still return a successful capture result even though nothing was written to StatsStore. That turns transient backpressure into permanent stats loss. Can we fall back to inline persistence here, or otherwise surface failure so the caller can retry/requeue instead of discarding the capture?

mcusack-floe (changes requested):

service/src/main/java/ai/floedb/floecat/service/statistics/engine/impl/AbstractNativeStatsCaptureEngine.java
mrouvroy-floe: Fixed.
Rejection no longer drops records. On async queue rejection we now fallback to caller-thread persistence (fallback=caller_thread), so there is no data-loss path.
Test: fallsBackToCallerThreadPersistenceWhenQueueRejects in IcebergNativeStatsCaptureEngineTest.java

service/src/main/java/ai/floedb/floecat/service/statistics/engine/impl/AbstractNativeStatsCaptureEngine.java
mrouvroy-floe: I intentionally did not block SYNC on persistence completion.

SYNC’s goal is low-latency in-query stats delivery; waiting for store writes adds avoidable tail latency.
The captured record already makes the current response correct; persistence is for reuse by future requests.
Blocking on writes would couple query latency to storage write health/backpressure.

I just added some improvements though :
- Adding request-key singleflight for SYNC capture (tableId + snapshotId + target + connectorType), so concurrent misses share one capture attempt.
- Adding a short post-capture hold window on that same singleflight key, so immediate follow-up SYNC calls reuse the captured result instead of recapturing while persistence is still in flight.
- On wait timeout/failure, we now return miss (and async fallback path) instead of immediate recapture, to avoid capture storms.
- Adding timed eviction for completed singleflight entries.
- Tests added for concurrent, sequential, and timeout behaviors.

mcusack-floe: Got it. Makes sense.

mcusack-floe (approved):

PR #270, opened 2026-04-20
New blog on NDV and added author block

mcusack-floe:
## Summary

Adds author block to each blog and a new blog on NDV calculation.

## Type of Change

- [ ] fix (bug fix)

- [X] No breaking changes


site-src/_includes/page__meta.html
mrouvroy-floe, line 1: This file is used in more than one place (single post pages and post lists).
Right now it always reads data from `page`, which can cause the wrong author/date to appear in post lists.

Can we make it use:
`document = post | default: page`
and then read author/date from `document`?

This makes the template pick the “current post being shown” when in a list, and the “current page” otherwise.

site-src/_posts/2026-04-20-estimating-ndv.md
```diff
@@ -0,0 +1,173 @@
...
+title: "Sampling-Based NDV Estimation in Iceberg Tables"
+date: 2026-04-20
+comments: true
+author: mark_cusack
+toc: true
+toc_sticky: true
+classes: wide
+
+header:
+  image: /images/2026-04-20-graph-theme.png
```
mrouvroy-floe: This image file is very large (~1.7MB), which can make the page load slower for readers.
Can we replace it with a compressed version (for example WebP or optimized JPG) while keeping the same look/size?
We don't want to risk perf because of an image

site-src/_includes/page__meta.html
mcusack-floe: Fixed

site-src/_posts/2026-04-20-estimating-ndv.md
mcusack-floe: Changed to a smaller jpg.

site-src/assets/css/main.scss
```diff
@@ -29,32 +37,120 @@ body {
   margin: 0 auto;
   padding-left: 1rem;
   padding-right: 1rem;
+  width: 100%;
+  overflow-x: hidden;
 }
 
-.layout--home .page,
+#main,
```
mrouvroy-floe: overflow-x: hidden is applied to .layout--single .page__content and multiple ancestors.
toc_sticky: true relies on position: sticky, and sticky behavior commonly stops working when ancestors have non-visible overflow.
This risks the TOC no longer sticking while scrolling on post pages.

site-src/_posts/2026-04-20-estimating-ndv.md
mrouvroy-floe: The post now references the smaller JPG, but both 2026-04-20-graph-theme.png and 2026-04-20-graph-theme copy.png are still in git at ~1.7MB each and are not referenced anywhere.
We should probably remove them

mcusack-floe: Removed these images

mrouvroy-floe (approved):

PR #271, opened 2026-04-20
Handle capture-now child cancellation and avoid terminalizing interrupted reconcile jobs

mcusack-floe:
## Summary

This fixes cancellation behavior across captureNow, the reconcile job stores, and the scheduler.

captureNow now falls back to cancelling active child EXEC jobs when the parent PLAN job has already succeeded but aggregated work is still in progress. Both job-store implementations also make repeated cancel(...) calls idempotent for JS_CANCELLING, so a second cancel no longer forces a premature JS_CANCELLED.

On the scheduler side, missing executors are now recorded as job failures instead of escaping, and plain worker-thread interrupts no longer mark jobs cancelled. Only explicit store-backed cancellation requests terminalize jobs as JS_CANCELLED; interrupt-driven worker shutdown now leaves recovery to lease expiry/reclaim.

## Type of Change

- [X] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist

### DIFF
--- a/reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReconcilerScheduler.java
+++ b/reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReconcilerScheduler.java
@@ -195,25 +195,6 @@ public class ReconcilerScheduler {
   }
 
   void runLease(ReconcileJobStore.LeasedJob lease) {
-    ReconcileExecutor executor =
-        executorRegistry
-            .executorFor(lease)
-            .or(
-                () ->
-                    executorRegistry.orderedExecutors().stream()
-                        .filter(candidate -> candidate.supportsJobKind(lease.jobKind))
-                        .filter(
-                            candidate ->
-                                candidate.supportsExecutionClass(
-                                    lease.executionPolicy.executionClass()))
-                        .filter(candidate -> candidate.supportsLane(lease.executionPolicy.lane()))
-                        .filter(candidate -> candidate.supports(lease))
-                        .findFirst())
-            .orElseThrow(
-                () ->
-                    new IllegalStateException(
-                        "No reconcile executor available for lease " + lease.jobId));
-    cancellations.register(lease.jobId, Thread.currentThread());
     long started = System.currentTimeMillis();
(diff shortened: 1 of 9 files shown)


reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReconcilerScheduler.java
```diff
@@ -502,11 +524,11 @@ void runLease(ReconcileJobStore.LeasedJob lease) {
               progress.snapshotsProcessed,
               progress.statsProcessed,
               null);
-        } else if (Thread.currentThread().isInterrupted() || interrupted.get()) {
+        } else if (interruptedNow) {
```
mrouvroy-floe, line 527: RunLease now maps non-store thread interrupts to lease_lost without writing a terminal state. 
That only works if lease expiry/reclaim will requeue the job, but the in-memory store has no lease-expiry reclaim and only leases JS_QUEUED jobs. In memory mode, an interrupted worker can leave jobs stuck in JS_RUNNING/JS_CANCELLING indefinitely.

mcusack-floe: Brought in-memory queue into line with the durable queue, so this should be addressed. Test added.

mrouvroy-floe: Original issues looks good.
But in-memory leaseNext now leases JS_CANCELLING jobs, but markRunning flips them to JS_RUNNING before ReconcilerScheduler does its cancellationRequestedInStore check. Since isCancellationRequested is state-based, the cancel signal can be lost and a cancelled job can execute.

mcusack-floe: Addressed on the latest check in. in-mem and durable state transition models are now the same.

mrouvroy-floe (approved):

PR #287, opened 2026-05-05
Feat: Implement secure credential vending in Iceberg REST

mcusack-floe:
## Summary

This PR adds storage-authority-based credential vending end to end, backed by the secrets manager, and threads it through the Iceberg REST gateway, transaction planning, connectors, CLI, and docs.

The main product change is a new storage authority model and API. The service now supports creating, updating, deleting, listing, and resolving storage authorities, with credentials stored in secrets manager rather than embedded in table or connector state. The gateway uses that authority layer to resolve storage credentials for table operations, metadata IO, connector provisioning, staged commits, and table/view loads. As part of that cleanup, the old magic storage-authority-lookup table-id path was removed and replaced with a first-class location-scoped resolve RPC.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist


service/src/main/resources/application.properties
```diff
@@ -39,8 +39,10 @@ quarkus.grpc.clients.floecat.plain-text=true
 ## Floecat repository seeding
 floecat.seed.enabled=${FLOECAT_SEED_ENABLED:true}
 floecat.seed.mode=${FLOECAT_SEED_MODE:fake}
+floecat.seed.sync.enabled=${FLOECAT_SEED_SYNC_ENABLED:true}
 ## Pointer store configuration
 floecat.kv=${FLOECAT_KV:memory}
```
kwesterfeld-floe, line 44: Does this mean that we have to flip this to something non-KV in floecat deployment?

service/src/main/java/ai/floedb/floecat/service/security/RolePermissions.java
```diff
@@ -46,7 +48,8 @@ public final class RolePermissions {
           "account.delete");
   private static final List<String> PLATFORM_PERMS =
       List.of("account.read", "account.write", "account.delete");
-  private static final List<String> SYSTEM_OBJECTS_PERMS = List.of("system-objects.read");
+  private static final List<String> SYSTEM_OBJECTS_PERMS =
+      List.of("system-objects.read", STORAGE_AUTHORITY_RESOLVE_INTERNAL);
```
kwesterfeld-floe: This change looks like a security issue.  The planner uses the system objects "assumeRole" on the core side, which creates a scoped JWT that contains this set of limited permissions.  I'm not 100% sure, but to me it looks like what this means is the ability to call assumeRole with this role would give the resulting JWT the ability to call floecat APIs that contain STORAGE_AUTHORITY_RESOLVE_INTERNAL and this would be an unauthenticated path.

kwesterfeld-floe (commented):
I think the default dev/test for floecat.secrets should be "kv" and default should be "aws", rather than the other way around.  Otherwise, we will have to be 100% sure we never can change the prod value or omit it on the deployment in the application-base deployment for floecat.

service/src/main/java/ai/floedb/floecat/service/security/RolePermissions.java
mcusack-floe: I've removed STORAGE_AUTHORITY_RESOLVE_INTERNAL from the system-objects role. There is a narrower use of it but only for the reconciler-worker role, which is a dedicated machine principal.

service/src/main/resources/application.properties
mcusack-floe: The current implementation has now removed that manual floecat.secrets flip. In production, the prod build profile selects AWS secrets automatically.

PR #297, opened 2026-05-14
Refactor Durable Reconciler Queue to Native Domain-Based Storage

mcusack-floe:
## Summary

This PR overhauls the durable reconciler queue from a repair-prone pointer-driven model into an explicit domain-based architecture with clear ownership boundaries for canonical job indexes, ready-queue state, lease coordination, and projection/payload-reference state. It removes opportunistic self-healing from reads, lease scans, and maintenance, makes canonical job-state transitions the sole authority for job-index mutation, and moves the hot durable paths onto native Dynamo-oriented storage patterns for job lookup/listing, ready leasing, and lease expiry handling while keeping memory mode behavior aligned. Along the way it cleans up legacy pointer-era abstractions, replaces ambiguous adapter layers with native store contracts and write batches, tightens tests around current invariants instead of repair behavior, fixes several live regressions uncovered during large capture runs, and updates the documentation to describe the new reconciler and queue architecture accurately.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [ ] No breaking changes

## Checklist

## Additional Notes

Changes auto sync behaviour from backfilling every snapshot to just the current snapshot. Old behaviour doesn't work well with streaming Delta tables that can have thousands of snapshots.


reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/GrpcRemoteReconcileExecutorClient.java
mrouvroy-floe, line 832: `CaptureScope.snapshot_selection` is not mapped here. `fromProtoScope(...)` rebuilds `ReconcileScope` without passing snapshot selection, so remote workers always fall back to UNSPECIFIED/ALL behavior.

Service-side lease serialization already includes this field (`ReconcileExecutorControlImpl.toProtoScope`), so this currently regresses correctness only on remote execution.

Can we include 
```
scope.hasSnapshotSelection() ? fromProtoSnapshotSelection(...) : ReconcileSnapshotSelection.unspecified()
```
 in this mapping ?

And also add a unit test for `fromProtoScope(...)` that asserts `snapshot_selection` is preserved for CURRENT/LATEST_N/EXPLICIT/ALL?

client-cli/src/main/java/ai/floedb/floecat/client/cli/ConnectorCliSupport.java
mrouvroy-floe, line 587: `--dest-view` currently rejects `--snapshot/--current/--columns`, but still allows `--latest-n`/`--all`. Later we only set `snapshot_selection` when `destView` is blank, so these flags are silently ignored.

Can we reject all snapshot-scope flags with `--dest-view` (or explicitly support them) to avoid no-op CLI combinations?

client-cli/src/test/java/ai/floedb/floecat/client/cli/ConnectorCliSupportTest.java
mrouvroy-floe, line 1: Let's add CLI tests for `--dest-view --latest-n` and `--dest-view --all` (currently accepted but snapshot scope is not applied for views). Either reject these combinations or make behavior explicit.

mrouvroy-floe (approved):

PR #298, opened 2026-05-14
service/context: preserve request context across Vert.x gRPC handoff

kwesterfeld-floe:
## Summary

There is an intermittent problem with request-context propagation through the
gRPC server call path:

```
vertx -> interceptor -> blocking interceptor wrapper -> service
```

In some requests, inbound call context is resolved correctly in the interceptor,
but by the time the request reaches the service, the gRPC Context no longer
contains that state. This was first observed with principal propagation, where
calls could see an empty/default principal even though the inbound JWT or
session header was valid, but the same gap can also affect other request-scoped
context such as engine information.

This change makes request-context lookup resilient to that propagation gap by
introducing ContextStore, which mirrors request-scoped values into both:

- the gRPC Context
- the current Vert.x local context

Context consumers continue to read the same request context keys, but now fall
back to the Vert.x-local copy when the gRPC Context is missing or empty. The
interceptor also clears the Vert.x-local request state at call close so the
fallback remains request-scoped.

## Why

The failure is not in JWT parsing. The issue is that the gRPC context can be lost across the blocking interceptor handoff, while the Vert.x request context remains available.

## What

```
    - Preserve request-scoped principal, engine, query, correlation, and auth
      values when gRPC context is lost after the blocking interceptor handoff.
    - Route request context reads through a shared context store that falls back
      from gRPC context to Vert.x local request state.
    - Keep the interceptor wiring close to the original gRPC context flow while
      extending the fallback to other request-scoped context consumers.
    - Add regression coverage for the context loss path and the account create
      failure it can trigger.
```

## Tests

- Added ContextStoreTest.
- Added AccountCreateContextPropagationIT, which repeatedly calls CreateAccount with a valid session JWT and fails only on the known propagation signatures.

## Type of Change

- [x] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist

## Additional Notes

### DIFF
--- a/core/scanner/src/main/java/ai/floedb/floecat/scanner/utils/EngineContext.java
+++ b/core/scanner/src/main/java/ai/floedb/floecat/scanner/utils/EngineContext.java
@@ -138,4 +138,8 @@ public final class EngineContext {
   public boolean hasEngineHeaders() {
     return hasEngineKind;
   }
+
+  public boolean isEmpty() {
+    return !hasEngineKind && engineVersion.isBlank();
+  }
 }
--- a/service/src/main/java/ai/floedb/floecat/service/account/impl/AccountServiceImpl.java
+++ b/service/src/main/java/ai/floedb/floecat/service/account/impl/AccountServiceImpl.java
@@ -33,6 +33,7 @@ import ai.floedb.floecat.account.rpc.UpdateAccountRequest;
 import ai.floedb.floecat.account.rpc.UpdateAccountResponse;
 import ai.floedb.floecat.catalog.rpc.Catalog;
 import ai.floedb.floecat.common.rpc.MutationMeta;
+import ai.floedb.floecat.common.rpc.PrincipalContext;
 import ai.floedb.floecat.common.rpc.ResourceId;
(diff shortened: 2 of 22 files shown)


kwesterfeld-floe:
Closing in favour of https://github.com/eng-floe/floecat/pull/299.

That PR addresses the same intermittent gRPC Context drop, but at the root cause: the deprecated io.vertx.grpc.BlockingServerInterceptor.wrap that the chain currently uses does not carry io.grpc.Context across its
buffered-event replay. Combined with Quarkus's GrpcDuplicatedContextGrpcInterceptor (which hops via runOnContext on Vert.x-context mismatch), that's what leaves the service-method body with an empty PrincipalContext.

This PR's ContextStore + Vert.x-locals mirror is functionally close to what vertx-grpc-context-storage would provide, and we confirmed neither survives the duplicated-context hop in isolation. https://github.com/eng-floe/floecat/pull/299 replaces the wrap
with one that captures the gRPC Context at event arrival and re-applies it on the worker around every listener callback, so existing principal.get() / Context.Key.get() call sites work unchanged — no per-consumer
migration needed.

PR #299, opened 2026-05-16
service/context: preserve grpc Context across blocking worker hop

kwesterfeld-floe:
## Summary

The deprecated io.vertx.grpc.BlockingServerInterceptor.wrap that backed our BlockingInboundContextInterceptor does not carry the io.grpc.Context across its buffered-event replay. Combined with Quarkus's GrpcDuplicatedContextGrpcInterceptor — which hops via duplicated.runOnContext when an inbound event arrives on a Vert.x context that doesn't match the chain's duplicated one — the gRPC Context ThreadLocal is gone by the time the service-method body reads it. That is the production symptom: an empty PrincipalContext at createAccount (and elsewhere) on a subset of inbound calls even though the interceptor resolved a valid principal.

This PR replaces the deprecated wrap with a custom worker-thread wrap modelled on Quarkus's internal BlockingServerInterceptor + BlockingExecutionHandler pattern, tightened in one way: the gRPC Context is captured at
event arrival (inside scheduleOrEnqueue), not at execute time, so it survives the buffer/replay even when no inner ContextualizedServerCallListener is in the chain to re-attach it. The active CDI request-context state is captured on the caller and re-activated on the worker. Per-event dispatch is chained via executeBlocking onComplete so listener callbacks stay ordered.

This supersedes #298, which addressed the same symptom by mirroring each request-scoped value into Vert.x duplicated-context locals via a ContextStore facade. That approach is conceptually similar to vertx-grpc-context-storage (which was verified does not survive the runOnContext hop because duplicated contexts have per-duplicate locals), and it required updating every consumer call site. The fix here addresses the loss at its source — the buffered-event replay running on a thread without the original gRPC Context attached — so existing KEY.get() and principal.get() call sites work unchanged.

## Commit Summary

- Replace deprecated Vert.x BlockingServerInterceptor.wrap, which does not carry the gRPC Context across its buffered-event replay; under the Quarkus duplicated-context interceptor's runOnContext hop this dropped the request principal and correlation context on a subset of inbound calls.
- Introduce a worker-thread wrap that captures the gRPC Context at event arrival, captures the active CDI request context state on the caller, and re-applies both on the worker around each listener callback. Per-event dispatch is chained via executeBlocking onComplete so events stay in order.
- Pass the inner helper to the wrap via a method reference, removing the runtime JDK proxy that previously bridged its non-ServerInterceptor shape.
- Add a floecat.interceptor.blocking.legacy-wrap config flag (env FLOECAT_INTERCEPTOR_BLOCKING_LEGACY_WRAP) that reverts to the deprecated wrap for before/after validation, with a startup warning.
- Cover the new wrap with focused tests: single-call propagation, buffered-event replay, concurrent calls without cross-call bleed, all five listener callbacks forwarded in order through the buffer, error propagation closing the call with the right status, and a reproducer that reads Context.Key inside the handler and fails with the deprecated wrap selected.

## Type of Change

- [x] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist

## Additional Notes

- Validation toggle. floecat.interceptor.blocking.legacy-wrap (env: FLOECAT_INTERCEPTOR_BLOCKING_LEGACY_WRAP=true) reverts to the deprecated wrap. Off by default; logs a startup warning when on. The same toggle flips the test suite, so mvn test -Dtest=CtxPropagatingBlockingWrapTest is a self-contained before/after harness.
- JDK proxy removed. ServerInterceptor is a functional interface; the inner helper is now passed via inbound::interceptCall rather than a runtime Proxy.newProxyInstance bridge. No behaviour change.
- #298 disposition. If #298 has not merged, close it in favour of this PR. If it has, the follow-up cleanup removes ContextStore, PrincipalContexts, EngineContext.isEmpty(), the ContextStore.clear() call in the interceptor close handler, and every ContextStore.get(KEY) consumer migration.
- Why not the gRPC ContextStorage override path. io.vertx:vertx-grpc-context-storage rebinds gRPC Context storage to Vertx.currentContext().localContextData(). Duplicated Vert.x contexts have per-duplicate locals, so storage written on context A is not visible on context A-dup after runOnContext. Source inspection confirms it; reproducer test (now removed; was retained briefly in the branch history) demonstrated it.
- References. Vert.x deprecated BlockingServerInterceptor (no context propagation); Quarkus internal BlockingServerInterceptor + BlockingExecutionHandler (the model); Quarkus #14665 / PR #14697; Quarkus #26830.

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/context/impl/CtxPropagatingBlockingWrap.java
+++ b/service/src/main/java/ai/floedb/floecat/service/context/impl/CtxPropagatingBlockingWrap.java
@@ -1,0 +1,259 @@
+/*
+ * Copyright 2026 Yellowbrick Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ai.floedb.floecat.service.context.impl;
+
+import io.grpc.Context;
+import io.grpc.Metadata;
+import io.grpc.ServerCall;
+import io.grpc.ServerCallHandler;
+import io.grpc.ServerInterceptor;
+import io.grpc.Status;
+import io.quarkus.arc.Arc;
+import io.quarkus.arc.InjectableContext;
+import io.quarkus.arc.ManagedContext;
+import io.vertx.core.Vertx;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.function.Consumer;
+import org.jboss.logging.Logger;
+
+/**
+ * Worker-thread {@link ServerInterceptor} wrap that preserves both the gRPC {@link Context} and the
+ * CDI request context across the worker hop.
+ *
+ * <p>The pattern mirrors {@code
+ * io.quarkus.grpc.runtime.supports.blocking.BlockingServerInterceptor} + {@code
+ * BlockingExecutionHandler}: the chain build is dispatched via {@code executeBlocking}, and each
+ * subsequent listener event is re-dispatched via {@code executeBlocking} with the gRPC {@link
+ * Context} captured at scheduling time and re-attached on the worker. Unlike Quarkus's class (which
+ * only activates for {@code @Blocking}-annotated methods), this wrap applies to every method.
+ *
+ * <p>Replaces the deprecated {@code io.vertx.grpc.BlockingServerInterceptor.wrap}, which does
+ * <i>not</i> propagate {@code io.grpc.Context} across its buffered-event replay. Combined with
+ * Quarkus's {@code GrpcDuplicatedContextGrpcInterceptor} hopping via {@code runOnContext} on
+ * Vert.x-context mismatch, the deprecated wrap drops the principal/correlation context on a subset
+ * of inbound calls.
+ *
+ * <p>Package-private so the test suite can instantiate it directly without going through the full
+ * Quarkus interceptor chain.
+ */
+final class CtxPropagatingBlockingWrap implements ServerInterceptor {
+  private static final Logger LOG = Logger.getLogger(CtxPropagatingBlockingWrap.class);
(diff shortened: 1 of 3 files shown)


service/src/main/java/ai/floedb/floecat/service/context/impl/CtxPropagatingBlockingWrap.java
```diff
@@ -0,0 +1,259 @@
...
+                    } catch (RuntimeException ignored) {
+                      // ignore
+                    }
+                  }
+                }
+                return null;
+              },
+              false)
+          .onComplete(
+              p -> {
```
mrouvroy-floe, line 224: Queue-drain race can leave inbound events permanently stuck
`onComplete` polls one event and then flips `isConsumingFromIncomingEvents` to `false`.
If a callback arrives between the `poll()` and that flag write, `scheduleOrEnqueue()` observes `isConsumingFromIncomingEvents=true` and enqueues the event, then no further drain is triggered. 

That can drop terminal callbacks (`onHalfClose`/`onComplete`) for a request.

service/src/test/java/ai/floedb/floecat/service/context/impl/CtxPropagatingBlockingWrapTest.java
```diff
@@ -0,0 +1,546 @@
...
+class CtxPropagatingBlockingWrapTest {
+
+  private static final Context.Key<String> KEY = Context.key("test-principal");
+
+  /**
+   * Selects which wrap the tests run against. Honours {@code
+   * floecat.interceptor.blocking.legacy-wrap} (env: {@code FLOECAT_INTERCEPTOR_BLOCKING_LEGACY_WRAP})
+   * so the same env var that flips the production wiring also flips these tests — with the legacy
+   * wrap selected, at least one test below is expected to fail (see {@link
+   * #eventHandlerRunsOffEventLoop}).
```
mrouvroy-floe: The Javadoc references `eventHandlerRunsOffEventLoop`, but that method does not exist in this class ?

service/src/main/java/ai/floedb/floecat/service/context/impl/CtxPropagatingBlockingWrap.java
kwesterfeld-floe: Fixed

service/src/test/java/ai/floedb/floecat/service/context/impl/CtxPropagatingBlockingWrapTest.java
kwesterfeld-floe: Fixed

mrouvroy-floe (approved):

PR #312, opened 2026-05-28
Require row_count in scalar stats and remove ambiguous value_count

mcusack-floe:
## Summary

This changes the stats contract so scalar stats use required row_count plus optional null_count, and removes the old value_count field that had inconsistent semantics across connectors.

Key changes:

- Replace ScalarStats.value_count with required ScalarStats.row_count.
- Normalize connector emission so row_count always means total rows/value slots.
- Keep null_count optional and separate from row_count.
- Update Delta, Iceberg, common connector emitters, reconciler rollup, service/query consumers, CLI output, and Iceberg REST compatibility code to the new contract.
- Remove legacy fallback/read-path handling for value_count.

## Type of Change

- [X] fix (bug fix)

## Testing

-  [X] `make fmt`
- [X] `make verify`

- [ ] No breaking changes

## Checklist

- [X ] PR is scoped and focused


quantumish:
What's the purpose of keeping `row_count` inside the data structure for column stats when the information almost always present in an encapsulating file-level stats object? I didn't run into issues when I just naively got rid of `value_count` everywhere.

mcusack-floe:
> What's the purpose of keeping `row_count` inside the data structure for column stats when the information almost always present in an encapsulating file-level stats object? I didn't run into issues when I just naively got rid of `value_count` everywhere.

We need row_count here because it is possible to select a column-only capture for which no file stats are collected.

PR #318, opened 2026-06-02
Make sure storage create() is atomic

jchauzi-floe:
Otherwise we could have half-created resources and not always clean-up.

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java
+++ b/service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java
@@ -24,6 +24,7 @@ import ai.floedb.floecat.storage.spi.BlobStore;
 import ai.floedb.floecat.storage.spi.PointerStore;
 import com.google.protobuf.Timestamp;
 import com.google.protobuf.util.Timestamps;
+import io.micrometer.core.instrument.Metrics;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
@@ -33,8 +34,11 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
+import org.jboss.logging.Logger;
 
 public class GenericResourceRepository<T, K extends ResourceKey> extends BaseResourceRepository<T> {
+
+  private static final Logger log = Logger.getLogger(GenericResourceRepository.class);
 
   private final ResourceSchema<T, K> schema;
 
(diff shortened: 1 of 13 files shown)


mrouvroy-floe:
How do we handle already-partial pointer state ?

The new create() path won’t create partial state anymore, because the batch either writes all pointers or none. 
But partial state can still exist : 
- from before this PR (do we care ?) , 
- from other non-atomic paths like createIfAbsent(), which still creates the canonical pointer first and then the secondary pointers.

For instance:
- canonical/by-id pointer exists
- secondary/by-name pointer is missing
- the canonical pointer points to the same blob that this create() call would write

Before this change, calling create() again would repair that by creating the missing secondary pointer. 
Now, it throws AbortRetryableException. But retry will never make progress: every retry sees the same canonical-present / secondary-missing state and throws the same exception again.

jchauzi-floe:
> How do we handle already-partial pointer state ?
> 
> The new create() path won’t create partial state anymore, because the batch either writes all pointers or none. But partial state can still exist :
> 
>     * from before this PR (do we care ?) ,

I would say we do not care at this moment?

> 
>     * from other non-atomic paths like createIfAbsent(), which still creates the canonical pointer first and then the secondary pointers.
> 
> 
> For instance:
> 
>     * canonical/by-id pointer exists
> 
>     * secondary/by-name pointer is missing
> 
>     * the canonical pointer points to the same blob that this create() call would write
> 
> 
> Before this change, calling create() again would repair that by creating the missing secondary pointer. Now, it throws AbortRetryableException. But retry will never make progress: every retry sees the same canonical-present / secondary-missing state and throws the same exception again.

So, it is not clear to me that `create` should be able to repair the state (I would assume only `update` or something can modify an existing entry). It would mean we need to make sure other paths (like `createIfAbsent`) cannot leave partial states.

service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java
```diff
@@ -95,41 +177,91 @@ public boolean createIfAbsent(T value) {
...
+  /**
+   * A stable partial-pointer state that an atomic create/createIfAbsent can never itself produce (a
+   * legacy orphan, or a non-atomic writer that died mid-flight). Under the strict no-repair
+   * contract we surface it as a non-retryable {@link CorruptionException}. Log it and bump a
+   * counter so the (rare) anomaly is visible and can be reconciled out of band rather than
+   * vanishing into a generic 500.
+   */
+  private CorruptionException partialStateAnomaly(String operation, String message) {
+    log.errorf("partial pointer state in %s.%s: %s", schema.resourceName, operation, message);
+    Metrics.counter(
```
mrouvroy-floe: Should this metric go through the service’s injected MeterRegistry / telemetry path instead of Micrometer’s static global registry? That way it gets automatically documented and checked by floecat like the rest of its metrics

mrouvroy-floe (approved):

PR #331, opened 2026-06-13
stats: exclude operational timestamps from target-stats content hash

jchauzi-floe:
## Problem

`SubmitLeasedFileGroupExecutionResult` resubmits fail `ABORTED MC_CONFLICT` →
`NameConflictException: pointer bound to different blob: .../target-generations/<gen>/targets/file-<sha>`,
so jobs burn attempts to permanent `FAILED`.

`StatsRepository.blobUri` content-addresses each target blob by `sha256(record.toByteArray())`,
but records carry wall-clock timestamps (`StatsMetadata.captured_at`/`refreshed_at`, every
`UpstreamStamp.fetched_at`). An otherwise-identical re-capture therefore hashes to a different
blob, and rebinding the stable target pointer to it throws. The idempotency fingerprint already
excludes operational timestamps by contract — the blob hash just never used it.

## Fix

- **`TargetStatsRecords.contentHashImage()`** — returns a copy with operational timestamps cleared
  (`captured_at`/`refreshed_at` + every `fetched_at`, including per-column scalars in file records).
  `blobUri` now hashes that image instead of raw bytes, so timestamp-only resubmits re-bind the
  stable pointer to the same blob (last-writer-wins on content) instead of conflicting. **Stored
  records keep their real timestamps** — this is hash-only, provenance/CLI unaffected.
- **`StatsCanonicalizer.canonicalUpstream`** — stop folding `fetched_at` into the idempotency
  fingerprint, honoring its own "operational timestamps intentionally excluded" docstring, so the
  fingerprint and blob identity now agree.
- Side effect: a crash between pointer bind and idempotency-record write is no longer fatal — a
  lost idempotency record causes a harmless re-bind, not a conflict.

## Tests

- New regression tests (table + file/per-column upstream) in `StatsRepositoryTargetStorageTest`.
- `StatsRepositoryTargetStorageTest`, `TableStatisticsServiceImplFingerprintTest`, `StatsIT` (9/9)
  all green; `fmt:check` clean. Existing `conflictingPutTargetStatsForSameTargetIsRejected` still
  passes — real content differences still conflict, only operational timestamps are excluded.

## Deploy note

Governs **new** writes only. Targets already bound under the old full-bytes hash will conflict
once on the first post-deploy resubmit into the existing active generation (new stripped-hash blob
≠ stale pointer's old blob). Clear `targets-active`/generation pointers (or delete/recreate) for
affected snapshots after rollout.

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/repo/impl/StatsRepository.java
+++ b/service/src/main/java/ai/floedb/floecat/service/repo/impl/StatsRepository.java
@@ -427,6 +427,13 @@ public class StatsRepository implements StatsStore {
     }
 
     private boolean createIfAbsent(String pointerKey, String blobUri, TargetStatsRecord value) {
+      // The target pointer is bound, so ifAbsent must leave the existing record untouched. Now
+      // that content-hash images can map distinct records to one blobUri (timestamp-only resubmits
+      // share a blob), writing the blob before this check would overwrite the live record's bytes
+      // and still return false on the CAS miss. Check the pointer first and write nothing.
+      if (pointerStore.get(pointerKey).isPresent()) {
+        return false;
+      }
       boolean blobExistedBefore = blobStore.head(blobUri).isPresent();
       putBlob(blobUri, value);
       Pointer reserve = PointerReferences.blobPointer(pointerKey, blobUri, 1L);
--- a/service/src/main/java/ai/floedb/floecat/service/repo/impl/StatsRepository.java
+++ b/service/src/main/java/ai/floedb/floecat/service/repo/impl/StatsRepository.java
@@ -358,7 +358,7 @@ public class StatsRepository implements StatsStore {
         record.getSnapshotId(),
         generationId,
         StatsTargetIdentity.storageId(record.getTarget()),
-        Hashing.sha256Hex(record.toByteArray()));
+        Hashing.sha256Hex(TargetStatsRecords.contentHashImage(record).toByteArray()));
   }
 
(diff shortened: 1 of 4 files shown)


service/src/main/java/ai/floedb/floecat/service/repo/impl/StatsRepository.java
```diff
@@ -358,7 +358,7 @@ private String blobUri(TargetStatsRecord record, String generationId) {
         record.getSnapshotId(),
         generationId,
         StatsTargetIdentity.storageId(record.getTarget()),
-        Hashing.sha256Hex(record.toByteArray()));
+        Hashing.sha256Hex(TargetStatsRecords.contentHashImage(record).toByteArray()));
```
mcusack-floe, line 361: With this change putTargetStatsIfAbsent no longer preserves existing content once multiple records can share a blob URI. With this PR, records that differ only in excluded timestamps hash to the same blobUri, but this path still calls putBlob(blobUri, value) before checking whether the pointer already exists. If the pointer CAS then fails, the method returns false even though the existing target’s blob may already have been overwritten with the new record bytes. That breaks the expected ifAbsent behavior. I think this needs to check the pointer first and avoid writing anything when the target already exists.

jchauzi-floe: Added a pointer check before anything in `createIfAbsent`

mcusack-floe (approved):

PR #332, opened 2026-06-14
Truncate oversized downstream-exception detail in gRPC status mapping

jchauzi-floe:
## Problem

A downstream `DynamoDbException` (e.g. the `TransactWriteItems` 100-item breach) carries the **entire offending request** — tens of KB — in its message. `GrpcErrors` mapped it to `INTERNAL` and copied that text verbatim into the `Error` detail's `root_message` param. The `MC_INTERNAL` catalog template (`errors_en.properties`) contains `{root_message}`, so `LocalizeErrorsInterceptor` **re-rendered** the full dump into both the `grpc-message` description and the re-packed `grpc-status-details-bin` trailer.

That trailer then exceeded the HTTP/2 `max-header-list-size`, so Netty / the Envoy waypoint reset the stream with `RST_STREAM PROTOCOL_ERROR`. The reconcile worker reported a transport error instead of the real cause — the actual DynamoDB reason was only recoverable from floecat's own server-side `RpcLoggingInterceptor` log. This masks *any* large downstream error, not just the DynamoDB one.

Note `GrpcErrors.build`'s message-hiding for `INTERNAL` was **not** sufficient: the interceptor's locale re-render re-expands `{root_message}`, so the fix has to bound the text at its source param.

## Fix

A `clampDetail(String)` helper in `GrpcErrors` that:
- collapses whitespace (multi-line SDK dumps → header-safe single line), and
- when over the cap, **keeps head + tail and elides the middle** — AWS places the actionable constraint *after* the embedded value dump, so a head-only truncation would discard it.

Cap is `floecat.errors.max-detail-chars` (default **512**), read the same way as the existing `floecat.errors.debug-details` flag. Applied to the `root_message` param, the resolved status message, and the `DebugInfo` detail — bounding every copy that reaches the status/trailer.

## Tests

`GrpcErrorsTest` (+3 new): head/tail-elide preservation, short/whitespace passthrough, and an end-to-end check that a 40 KB cause yields a header-safe rendered `MC_INTERNAL` message and bounded serialized status. Full targeted run green: `GrpcErrorsTest` 7/7, `GrpcErrorsContractArchTest` 5/5, `MessageCatalogTest` 2/2, `BaseServiceImplTest` 1/1. `mvn fmt:check` clean.

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/error/impl/GrpcErrors.java
+++ b/service/src/main/java/ai/floedb/floecat/service/error/impl/GrpcErrors.java
@@ -353,6 +353,8 @@ public final class GrpcErrors {
 
   private static final int DEFAULT_MAX_DETAIL_CHARS = 512;
   private static final Pattern WHITESPACE = Pattern.compile("\\s+");
+  private static final String ELISION_PREFIX = " ...[";
+  private static final String ELISION_SUFFIX = " chars elided]... ";
 
   private static boolean fetchDebugDetailsFlag() {
     try {
@@ -392,7 +394,8 @@ public final class GrpcErrors {
    *
    * <p>Whitespace is collapsed and, when the text exceeds the cap, the middle is elided while head
    * and tail are preserved — AWS messages place the actionable constraint <em>after</em> the
-   * embedded dump, so a head-only truncation would discard it.
+   * embedded dump, so a head-only truncation would discard it. The elision marker is counted
+   * against the cap, so the returned string never exceeds {@code max}.
    */
   static String clampDetail(String raw) {
     if (raw == null) {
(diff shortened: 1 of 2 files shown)


kwesterfeld-floe (approved):

service/src/main/java/ai/floedb/floecat/service/error/impl/GrpcErrors.java
```diff
@@ -364,6 +368,48 @@ private static boolean debugDetailsEnabled() {
...
+    }
+    String collapsed = WHITESPACE.matcher(raw).replaceAll(" ").trim();
+    int max = maxDetailChars();
+    if (collapsed.length() <= max) {
+      return collapsed;
+    }
+    int dropped = collapsed.length() - max;
+    int head = (max + 1) / 2;
+    int tail = max - head;
+    String elision = " ...[" + dropped + " chars elided]... ";
```
mrouvroy-floe: Should that marker be integrated into the max cap ? When we cap, we are emitting something with length 538, not 512

jchauzi-floe: Changed

mrouvroy-floe (approved):

PR #334, opened 2026-06-15
Change to chunked inline submission of file group results

mcusack-floe:
## Summary

This refactors floecat remote reconcile result handling from blob-manifest based submission to chunked inline gRPC submission for both file-group execution and snapshot finalization. On the service side, file-group chunks now persist file stats artifacts incrementally, carry partial aggregate records forward, and snapshot finalization merges those partials instead of reloading per-group stats blobs.
It also hardens remote execution by adding worker-control deadlines/keepalives, better heartbeat/completion coordination, stricter child-state validation during snapshot finalization, and destination namespace auto-creation during planning. Tests were updated across the reconciler and service layers to cover the new chunked result flow.

## Type of Change

- [X] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist


service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
```diff
@@ -129,70 +163,84 @@ public boolean persistFailure(
...
+                        .toList());
+        if (!mergedAggregates.isEmpty()) {
+          persistence.persistStats(mergedAggregates);
+        }
+      }
+      case DIRECT_STATS -> {
+        if (!requestsStatsOutputs) {
+          return;
+        }
+        requirePlannerDirectStatsRecordCount(
```
mrouvroy-floe, line 247: In the direct-stats success path we validate directStatsRecordCount with persistence.listFileStats(tableId, snapshotId).size(), which only counts StatsTargetType.FILE records.

But the planner records directStatsRecordCount as the total number of records returned by direct capture, and direct capture is allowed to return table and column stats as well as file stats. For example, Delta direct capture passes requestedKinds into buildTargetStats(...), so a table-stats-only request can legitimately produce a single table record and zero file records.

That means remote chunked finalization will persist the table/column records successfully in persistStatsChunk, then reject the success marker because actual=0 from listFileStats(...) does not match the planner’s total count. Local finalization still loads the original direct-stats blob and checks records.size(), so this creates a remote-only regression for table-only, column-only, or mixed direct-stats captures.

mcusack-floe: Remote chunked finalization no longer validates directStatsRecordCount against listFileStats().size(), and instead uses the persisted per-chunk direct-stats count carried on ReconcileSnapshotTask, so table-only, column-only, and mixed direct-stats captures no longer fail in the remote path just because there are zero file stats.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
```diff
@@ -477,108 +524,177 @@ private static PutIndexArtifactItem toPutIndexArtifactItem(
...
-                      ? ""
-                      : preUploadedArtifact.contentType())
-              .setUploadedArtifactUri(
-                  preUploadedArtifact.uploadedArtifactUri() == null
-                      ? ""
-                      : preUploadedArtifact.uploadedArtifactUri())
-              .build());
+    List<TargetStatsRecord> fileStats =
+        chunkStats.stream().filter(TargetStatsRecord::hasFile).toList();
+    if (!fileStats.isEmpty()) {
```
mrouvroy-floe, line 591: `chunkFileGroupResult()` chunks `result.statsRecords()` as-is, and the Java capture path returns a mixed list: file records plus table/column aggregate records from `FileGroupTargetStatsRollup.complete(...)`. On the service side, `chunkPartialAggregates()` derives partial aggregates from file records when a chunk has any file stats, and only consumes the non-file aggregate records when the chunk has no file stats.

That makes the final aggregate result depend on arbitrary byte chunk boundaries. If the appended table/column records land in the same chunk as file records, they are ignored. If they land in an aggregate-only chunk, they are merged in addition to the file-derived partials from earlier chunks, so finalization can double-count table/column stats. It also means connector-provided aggregate records in mixed chunks are silently replaced by recomputed values.

Can we make file-group chunking/persistence use one source of truth for aggregate partials?

mcusack-floe: Ive addressed this. The java file group executor and finalizer had drifted away from what upstream Rust executors and finalizers do. They are now aligned.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/FileGroupExecutionSupport.java
mrouvroy-floe, line 52: This patch makes file-group result chunks file-stats-only, and the service now derives table/column partial aggregates from those file records. However, `requestedStatsTargetKinds()` still maps a table-only policy to only `StatsTargetKind.TABLE` and a column-only policy to only `StatsTargetKind.COLUMN`. That set is passed into `capturePlannedFileGroup()`, and the Java adapter now filters the result down to `record.hasFile()`.

So if direct stats is unavailable and we fall back to file groups, table-only or column-only capture can request no file records from the connector, discard the aggregate records it did request, and finalize with no aggregate stats to persist. `StatsOrchestrator` can create exactly those table-only/column-only policies.

Can we make the file-group capture path request `StatsTargetKind.FILE` whenever table or column stats are going to be derived from file-group chunks, even if `FILE_STATS` was not requested as an output?

mcusack-floe: Fair, I've fixed the file-group side so that file stats are always generated if either / or just table or col stats are requested.

mrouvroy-floe (approved):

PR #339, opened 2026-06-18
feat(stats): add planner stats contract, sketches, and serving QoS

mrouvroy-floe:
## Summary

Adds the Floecat-side planner statistics path.

This PR defines the planner stats serving contract, extends column stats with typed sketch payloads, computes/rolls up planner-relevant stats, and adds bounded serving behavior for planner requests.

Main changes:
- Add planner stats proto contract for target stats requests, payload levels, serving options, result statuses, and degradation metadata.
- Add typed sketch payload support for NDV, quantiles, MCV/frequencies, and tuple-style stats.
- Extend stats computation and rollup so file-level stats can produce table/column-level planner stats.
- Add NDV sketch merging, avg_width derivation, and sketch metadata propagation.
- Add planner stats serving QoS: target caps, response byte accounting, partial/downgraded responses, stale fallback, and cache-backed batch lookup.
- Add tests for stats contract behavior, rollup, repository storage, planner stats serving, cache/orchestrator behavior, and serving performance path.

This is the Floecat half of the planner stats prototype. Core/planner consumption lands separately.

Part 1 of stats end-to-end contract serie: 
- Floecat : https://github.com/eng-floe/floecat/pull/339
- Floescan : https://github.com/eng-floe/core/pull/1515
- Planner : https://github.com/eng-floe/core/pull/1516
- Explain : https://github.com/eng-floe/core/pull/1517

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [ ] No breaking changes

Breaking change note:

This changes/extends the stats proto contract used for planner stats prototyping. The branch is intended to pair with the matching Core planner branch.

## Checklist


core/proto/src/main/proto/floecat/catalog/stats.proto
```diff
@@ -39,7 +39,8 @@ message Ndv {
     NdvApprox approx = 2;
   }
 
-  repeated NdvSketch sketches = 20;
+  // Raw sketch artifacts for replay/diagnostics. Role is always SKETCH_ROLE_NDV here.
+  repeated SketchPayload sketches = 20;
```
jchauzi-floe, line 43: **B1** — Backward-compat: field 20 keeps its number but swaps message type `NdvSketch` -> `SketchPayload`. Field 1 changes wire type (string `type` -> enum `role`), so already-persisted TargetStatsRecord blobs that carry NDV sketches won't parse. Reserve 20 and assign a new field number, or backfill old blobs.

core/proto/src/main/proto/floecat/catalog/stats.proto
```diff
@@ -204,8 +240,16 @@ message ScalarStats {
   Ndv ndv = 7;
   optional string min = 8;
   optional string max = 9;
-  bytes histogram = 10;
-  bytes tdigest = 11;
+  // Average uncompressed width in bytes per non-null value.
+  // In the Parquet-footer path: ceiling(getTotalUncompressedSize() / row_count),
+  // minimum 1.  Sub-byte-per-row encodings (bit-packed integers, dictionary strings)
+  // return 1 rather than 0.  Exact decoded width requires a full value-decoding capture.
+  optional int64 avg_width_bytes = 10;
```
jchauzi-floe, line 248: **C4** — Backward-compat: fields 10/11 reuse the old `histogram`/`tdigest` numbers with new types and no `reserved`. Safe today only because those fields were never populated. Add `reserved 10, 11;` (+ the names) and renumber the new fields.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
```diff
@@ -543,75 +546,33 @@ private static List<TargetStatsRecord> nonNullStatsRecords(List<TargetStatsRecor
...
+    List<TargetStatsRecord> nonNullStats = nonNullStatsRecords(statsRecords);
+    return nonNullStats.stream().filter(record -> !record.hasFile()).toList();
+  }
+
+  private static List<TargetStatsRecord> mergedPartialAggregates(
+      ReconcileFileGroupTask plannedTask, List<TargetStatsRecord> chunkPartials) {
+    List<TargetStatsRecord> nonNullChunkPartials = nonNullStatsRecords(chunkPartials);
+    if (nonNullChunkPartials.isEmpty()) {
       return plannedTask.partialAggregateRecords();
     }
```
jchauzi-floe, line 639: **B2** — `partial_aggregate_records` are appended per chunk without merging, so the EXEC_FILE_GROUP job item grows O(chunks x columns) with inline sketch bytes. Wide tables with many chunks will hit the DynamoDB 400KB item limit (cf. tpcds_1 chunked-submit). Merge per-column at append time, or stage them in a blob.

service/src/main/java/ai/floedb/floecat/service/query/catalog/PlannerStatsBundleService.java
```diff
@@ -200,55 +236,71 @@ public Multi<TargetStatsBundleChunk> streamTargets(
...
-                          .connectorType(connectorType)
-                          .build())
-              .toList();
-      List<StatsResolutionResult> resolved =
-          statsOrchestrator.resolveBatch(StatsCaptureBatchRequest.of(requests));
-      Map<String, Optional<TargetStatsRecord>> byTarget = new LinkedHashMap<>();
-      for (int i = 0; i < requests.size(); i++) {
-        StatsResolutionResult r = i < resolved.size() ? resolved.get(i) : null;
-        if (r != null && !r.hasStats() && r.outcome().name().startsWith("TIMEOUT")) {
+      long deadlineNanos = System.nanoTime() + policy.latencyBudget().toNanos();
```
jchauzi-floe: **C1** — latency_budget_ms is recomputed per table (providerLookup) and tables run sequentially, so worst-case block time is ~N_tables x budget (up to ~1s at max-tables=50), not the 20ms the proto promises. Is this expected? If not, compute one deadline in streamTargets and thread it through per-table lookups.

service/src/main/java/ai/floedb/floecat/service/statistics/StatsOrchestrator.java
```diff
@@ -164,6 +246,149 @@ public StatsResolutionResult resolve(StatsCaptureRequest request) {
...
+        new java.util.LinkedHashMap<>(requests.size());
+    java.util.List<StatsCaptureRequest> cacheMisses = new java.util.ArrayList<>();
+    for (StatsCaptureRequest req : requests) {
+      String key = ai.floedb.floecat.stats.identity.StatsTargetIdentity.storageId(req.target());
+      StatsCacheKey cacheKey =
+          new StatsCacheKey(
+              req.tableId().getAccountId(), req.tableId().getId(), req.snapshotId(), key);
+      TargetStatsRecord cached = statsCache.getIfPresent(cacheKey);
+      if (cached != null) {
+        observeSyncOutcome(StatsSyncOutcome.HIT, 0);
```
jchauzi-floe: **C2** — observeSyncOutcome(HIT, 0) with startNanos=0 records System.nanoTime() (JVM uptime) as the latency on every cache/batch hit, poisoning the SYNC_LATENCY p50/p95/p99. Same at L325. Add an outcome-only path that skips the timer on hits.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/FileGroupTargetStatsRollup.java
```diff
@@ -355,22 +368,27 @@ ScalarStats toScalar() {
...
+      }
       return builder.build();
     }
 
     private void mergeNdv(ScalarStats scalar) {
       if (!scalar.hasNdv()) {
         return;
       }
-      if (scalarContributors == 1) {
+      if (singleSourceNdv == null) {
```
jchauzi-floe: **C3** — mergeNdv only unions theta sketches; for exact/non-theta NDV it keeps singleSourceNdv and nulls it once a 2nd source contributes, so aggregateNdv returns null. Applied across file groups, any table with non-theta NDV that spans >1 group loses NDV entirely. Fall back to summing/maxing NdvApprox.estimate.

service/src/main/java/ai/floedb/floecat/service/query/catalog/PlannerStatsBundleService.java
```diff
@@ -592,16 +881,70 @@ private OptionalLong resolveSnapshot(TableWork work) {
...
           .setTarget(stats.getTarget())
-          .setStatus(BundleResultStatus.BUNDLE_RESULT_STATUS_FOUND)
+          .setStatus(status)
           .setStats(stats)
           .build();
     }
 
+    private static StatsResultStatus resultStatusForHit(
+        TargetNeed need, TargetStatsLookupResult lookupResult) {
+      if (lookupResult.stale()) {
```
jchauzi-floe: **G1** — Status is derived only from stale()/downgraded(); it never checks whether the record actually has the requested sketch. A within-cap SKETCHES request served from a scalar-only record returns HIT_COMPLETE with no sketches. Downgrade to SCALAR_DOWNGRADE when level>=SKETCHES and no sketch is present.

service/src/main/java/ai/floedb/floecat/service/query/catalog/PlannerStatsBundleService.java
```diff
@@ -507,9 +677,26 @@ protected TargetStatsBundleChunk batch() {
...
+         * Worst-case total response ≈ max_response_bytes (target stats) + constraint_bytes.
+         * Constraint size is schema-scoped (O(tables × columns)), not query-scoped, and is
+         * always small relative to a well-designed schema.
+         * Re-evaluate with an explicit constraint_max_bytes option if very wide tables +
+         * bulk constraint requests become a production concern. */
         if (includeConstraints && !work.constraintsEmitted()) {
-          constraintsOut.add(processConstraints(work));
+          var constraintResult = processConstraints(work);
+          constraintsOut.add(constraintResult);
+          responseBytes += constraintResult.getSerializedSize();
```
jchauzi-floe: **G2** — Constraint bytes are added to responseBytes, which tryCharge() reads, even though the contract says constraints don't count against max_response_bytes. With constraints in the request, targets can be wrongly OMITTED_BY_BUDGET. Track constraint bytes in a separate counter.

service/src/main/java/ai/floedb/floecat/service/query/catalog/PlannerStatsBundleService.java
```diff
@@ -592,16 +881,70 @@ private OptionalLong resolveSnapshot(TableWork work) {
...
           .setStats(stats)
           .build();
     }
 
+    private static StatsResultStatus resultStatusForHit(
+        TargetNeed need, TargetStatsLookupResult lookupResult) {
+      if (lookupResult.stale()) {
+        return StatsResultStatus.STATS_RESULT_HIT_STALE;
+      }
+      if (need.downgraded()) {
```
jchauzi-floe: **G3** — STATS_RESULT_HIT_PARTIAL is never emitted - resultStatusForHit only returns STALE/SCALAR_DOWNGRADE/HIT_COMPLETE. Either emit it on partial materialization or drop it from the contract for now.

service/src/main/java/ai/floedb/floecat/service/query/catalog/PlannerStatsBundleService.java
```diff
@@ -931,24 +1440,48 @@ private void markConstraintsEmitted() {
       this.constraintsEmitted = true;
     }
 
-    private void ensureTargetBatchLoaded(TargetStatsLookup lookup, long snapshotId) {
+    private void ensureTargetBatchLoaded(
+        TargetStatsLookup lookup, long snapshotId, ServingPolicy servingPolicy) {
       if (loadedSnapshot.isPresent() && loadedSnapshot.getAsLong() == snapshotId) {
         return;
```
jchauzi-floe, line 1336: **G4** — On a batch-read failure (StatsRepository.getTargetStatsBatch -> allOf().join() rethrows), loadedSnapshot is left unset, so each remaining target re-runs the full parallel batch read: N retries and N ERROR results for one failure. Cache a failed/empty marker for the snapshot and short-circuit.

core/connectors/common/src/main/java/ai/floedb/floecat/connector/common/GenericStatsEngine.java
```diff
@@ -112,15 +166,81 @@ public Result<K> compute() {
...
+       * Per-file NDV: fresh ColumnNdv accumulators per file; merge serialized theta
+       * bytes into the cumulative union; pass finalized per-file ColumnNdv to
+       * buildFileColumnAggs() so file records carry a mergeable theta sketch.
+       * Falls back to bootstrapPerFileNdv when ndvProvider is absent (Puffin path).
+       */
+      Map<K, ColumnNdv> perFileNdvByKey = null;
       if (ndvProvider != null && !ndvByName.isEmpty()) {
         try {
-          ndvProvider.contributeNdv(file.path(), ndvByName);
+          Map<String, ColumnNdv> perFileByName = new LinkedHashMap<>();
```
jchauzi-floe, line 212: **G5** — Table-level NDV loses rows_seen: the cumulative ColumnNdv only gets mergeTheta(bytes)+finalizeTheta(), and ColumnNdv.finalizeTheta never sets rowsSeen/rowsTotal. Per-file records keep it, table-level doesn't, breaking rows_seen-based extrapolation. Fold perFile.approx.rowsSeen into the cumulative.

core/connectors/common/src/main/java/ai/floedb/floecat/connector/common/ndv/ParquetAvgWidthProvider.java
```diff
@@ -0,0 +1,139 @@
...
+    if (sinks == null || sinks.isEmpty()) {
+      return;
+    }
+
+    try (ParquetFileReader reader = ParquetFileReader.open(parquetLookup.apply(filePath))) {
+      for (BlockMetaData block : reader.getFooter().getBlocks()) {
+        long blockRowCount = block.getRowCount();
+        for (ColumnChunkMetaData col : block.getColumns()) {
+          // Parquet column paths are dot-separated; use the first component as the top-level name.
+          // Use indexOf instead of split() to avoid regex allocation per column chunk.
```
jchauzi-floe, line 131: **G6** — Nested/repeated columns have multiple leaf chunks per block sharing one top-level name, and blockRowCount is added once per leaf -> denominator inflated, avg_width floored to 1. Add blockRowCount once per top-level column per block.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
```diff
@@ -185,7 +184,8 @@ public boolean persistChunk(
             .build();
```
jchauzi-floe, line 249: **G7** — plannedTask (and its baseline partials) is resolved before the idempotency lambda; mergedPartialAggregates appends only the current chunk. Two overlapping chunk submissions both read the same baseline and last-writer-wins drops partials. Re-resolve inside the creator (like persistSuccess does).

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/FileGroupTargetStatsRollup.java
```diff
@@ -320,17 +327,23 @@ void add(TargetStatsRecord source, ScalarStats scalar) {
...
-        nanCount = (nanCount == null ? 0L : nanCount) + Math.max(0L, scalar.getNanCount());
-      }
-      min = lowerEncoded(logicalType, min, scalar.hasMin() ? scalar.getMin() : null);
-      max = higherEncoded(logicalType, max, scalar.hasMax() ? scalar.getMax() : null);
+      min = lowerEncoded(decodedLogicalType, min, scalar.hasMin() ? scalar.getMin() : null);
+      max = higherEncoded(decodedLogicalType, max, scalar.hasMax() ? scalar.getMax() : null);
       mergeNdv(scalar);
+      /* Guard > 0: files with avg_width_bytes=0 (e.g. all-null columns or zero-row files)
+       * are excluded from the weighted average to prevent them from pulling the result toward 0. */
+      if (scalar.hasAvgWidthBytes() && scalar.getAvgWidthBytes() > 0 && scalar.getRowCount() > 0) {
```
jchauzi-floe, line 340: **G8** — avg_width re-aggregation weights group averages by the full group rowCount, but the level-1 average excluded zero-width/zero-row files from its divisor. Groups with many null/zero-width rows get over-weighted. Carry the actual weight (totalRowsForWidth) in the partial.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
```diff
@@ -223,6 +234,9 @@ void finalizeChunkedSuccess(
         if (!requestsStatsOutputs) {
           return;
         }
+        if (snapshotTask.directStatsPersistedRecordCount() > 0) {
+          return;
+        }
         SnapshotFinalizeChildStateService.ChildState childState =
             requireReadyChildState(lease, coverage);
         Set<FloecatConnector.StatsTargetKind> aggregateKinds = requestedAggregateKinds(lease);
```
jchauzi-floe, line 242: **G9** — finalizeChunkedSuccess flat-maps each group's accumulated partials, which already hold one record per chunk (append-without-merge). row_count/null_count get summed across chunk copies, so a resend double-counts. Dedup by (groupId, chunkIndex) before merge.

core/stats/src/main/java/ai/floedb/floecat/stats/identity/StatsTargetScopeCodec.java
```diff
@@ -49,6 +50,13 @@ public static String encode(StatsTarget target) {
             + SEP
             + b64(normalized.getEngineVersion());
       }
+      case COMPOSITE ->
+          COMPOSITE_PREFIX
+              + target.getComposite().getColumnIdsList().stream()
```
jchauzi-floe: **G10** — Codec is asymmetric: encode() handles composite, decode() has no composite branch and falls through to Optional.empty(). Composite targets round-trip to nothing on the reconcile path. Add a COMPOSITE_PREFIX branch to decode().

core/stats/src/main/java/ai/floedb/floecat/stats/identity/StatsTargetIdentity.java
```diff
@@ -205,10 +209,32 @@ public static String storageId(StatsTarget target) {
...
+   * request order. Duplicates are de-duplicated.
+   */
+  public static String compositeIdentityHashHex(CompositeColumnStatsTarget target) {
+    Objects.requireNonNull(target, "target");
+    if (target.getColumnIdsCount() < 2) {
+      throw new IllegalArgumentException("composite target requires at least 2 column_ids");
+    }
+    var sorted = new ArrayList<>(target.getColumnIdsList());
+    sorted.sort(Long::compare);
+    var sb = new StringBuilder("X").append(SEP);
```
jchauzi-floe: **G11** — Two issues: (1) `sb.append(SEP)` calls append(int) -> writes "31", not the 0x1f separator byte, so column-id boundaries are ambiguous/collidable; (2) no `id>0` check and no dedup, so [5,5] passes the <2 guard - contradicts the proto (positive, ascending) and the javadoc (dedup). Build a byte[] writing SEP as a byte; reject id<=0 and distinct() first.

core/proto/src/main/proto/floecat/catalog/stats.proto
```diff
@@ -147,12 +172,23 @@ message FileStatsTarget {
...
+}
+
 message StatsTarget {
   oneof target {
     TableStatsTarget table = 1;
     ColumnStatsTarget column = 2;
     EngineExpressionStatsTarget expression = 3;
     FileStatsTarget file = 4;
+    // Multi-column target: composite NDV, functional dependencies, join-overlap sketches.
+    CompositeColumnStatsTarget composite = 5;
```
jchauzi-floe, line 192: **M1** — Adding `composite` here isn't handled by the existing target switches in TargetStatsRecordValidator.validate (~L64-148) and Schemas.validateTargetValueCompatibility (~L226-247). They're arrow-form switch *statements*, so javac won't flag the missing case and a composite TargetStatsRecord bypasses validation via PutTargetStats. Add an explicit case COMPOSITE (validate or reject) before this is producible.

core/connectors/common/src/main/java/ai/floedb/floecat/connector/common/ConnectorStatsViewBuilder.java
```diff
@@ -127,13 +129,13 @@ public static <K> List<FloecatConnector.ColumnStatsView> toColumnStatsView(
...
-            var sb = ai.floedb.floecat.catalog.rpc.NdvSketch.newBuilder();
-            if (s.type != null) sb.setType(s.type);
+            var sb = SketchPayload.newBuilder().setRole(SketchRole.SKETCH_ROLE_NDV);
+            if (s.type != null) sb.setSketchType(s.type);
             if (s.data != null) sb.setData(ByteString.copyFrom(s.data));
-            if (s.encoding != null) sb.setEncoding(s.encoding);
-            if (s.compression != null) sb.setCompression(s.compression);
-            if (s.version != null) sb.setVersion(s.version);
-            if (s.params != null && !s.params.isEmpty()) sb.putAllProperties(s.params);
+            if (s.encoding != null) sb.putParams("encoding", s.encoding);
```
jchauzi-floe, line 141: **M2** — SketchPayload.completeness and captured_at_ms are left default when building payloads. Set them, or note they're intentionally unused for now.

core/connectors/common/src/main/java/ai/floedb/floecat/connector/common/StatsEngine.java
```diff
@@ -62,6 +62,14 @@ interface ColumnAgg {
     Object min();
 
     Object max();
+
+    // Average uncompressed width in bytes per non-null value, from Parquet footer metadata.
+    // Uses getTotalUncompressedSize() per ColumnChunk; ceiling-divided by row count so
```
jchauzi-floe, line 67: **M4** — Doc says avg_width is 'per non-null value' but the divisor is all rows (and the proto formula agrees with the code). A mostly-null column reports far too small. Align the doc and the code on one denominator.

core/connectors/common/src/main/java/ai/floedb/floecat/connector/common/ndv/ParquetAvgWidthProvider.java
```diff
@@ -0,0 +1,139 @@
...
+ * provides a reasonable proxy until a full value-decoding capture supplies exact decoded widths.
+ */
+public final class ParquetAvgWidthProvider {
+
+  /** Mutable accumulator passed to {@link #contributeAvgWidth}. */
+  public static final class AvgWidthAcc {
+    private long totalCompressedBytes;
+    private long totalRowCount;
+
+    public void add(long compressedBytes, long rowCount) {
```
jchauzi-floe: **M5** — Field/param/javadoc say 'compressed' but the value fed in is getTotalUncompressedSize() (correct, and the proto says uncompressed). Rename to uncompressed so a future 'fix' doesn't break it.

connectors/catalogs/iceberg/src/main/java/ai/floedb/floecat/connector/iceberg/impl/IcebergConnectorFactory.java
```diff
@@ -72,6 +72,15 @@ static FloecatConnector create(
       }
     } catch (NumberFormatException ignore) {
     }
+    // stats.ndv.theta_k: passed as a connector property and also embedded in the
+    // CapturePolicy.properties map so Floescan (remote executor) can read it via
```
jchauzi-floe: **M6** — Comment says theta_k is forwarded via CapturePolicy.properties, but nothing writes that key - the remote executor path uses default k. Either populate the property or drop the comment.

service/src/main/java/ai/floedb/floecat/service/query/catalog/PlannerStatsBundleService.java
```diff
@@ -546,44 +736,143 @@ protected TargetStatsBundleChunk end() {
...
+        }
+        case NONE -> {}
+      }
+    }
+
+    private boolean tryCharge(TargetStatsResult result) {
+      if (servingPolicy.maxResponseBytes() <= 0) {
+        return true;
+      }
+      int candidateBytes = result.getSerializedSize();
```
jchauzi-floe: **M7** — tryCharge charges getSerializedSize() of the bare result, ignoring repeated-field framing + the chunk envelope, so the real response can exceed max_response_bytes. Soft cap, but worth a fixed framing allowance.

service/src/main/java/ai/floedb/floecat/service/query/catalog/PlannerStatsBundleService.java
```diff
@@ -200,55 +236,71 @@ public Multi<TargetStatsBundleChunk> streamTargets(
...
-            StatsTargetIdentity.storageId(requests.get(i).target()),
-            r != null ? r.stats() : Optional.empty());
+        byTarget.put(target.storageId(), TargetStatsLookupResult.fromResolution(r));
       }
       return Map.copyOf(byTarget);
     };
   }
 
+  private static Optional<Duration> remainingLatencyBudget(long deadlineNanos) {
+    long remainingNanos = deadlineNanos - System.nanoTime();
```
jchauzi-floe: **M8** — remainingLatencyBudget(long) is unused - remove it. Separately, the sync-capture path (StatsOrchestrator.resolvePlannerBatch) re-reads the store per target for targets the batch already proved missing; skip the redundant read.

core/proto/src/main/proto/floecat/catalog/stats.proto
mrouvroy-floe: Will leave as-is for now, assuming no backward compatibility yet

core/proto/src/main/proto/floecat/catalog/stats.proto
mrouvroy-floe: Will leave as-is for now, assuming no backward compatibility yet

mcusack-floe:
Duplicate targets can incorrectly consume the sketch/full payload budgets before deduplication, so later unique targets get downgraded even though the duplicates are dropped. 

In PlannerStatsBundleService tupleTargets/sketchTargets are incremented while iterating the raw TargetStatsNeed list, and only afterwards does the code sort and dedupe by storageId. A request like [col1 FULL, col1 FULL, col2 FULL] with maxTupleTargets=2 will downgrade col2 to sketches/scalar even though there are only two unique targets. This is a behavioral regression in request normalization, and I don’t see a test covering duplicate requests under payload budgets.

mcusack-floe:
In PlannerStatsBundleService max_response_bytes can silently suppress constraints for later tables despite the new contract comment saying constraints bypass the byte cap. 

TargetStatsIterator.hasMoreWork() stops the whole iterator once byteBudgetExhausted is set and omitAllRemainingByBudget() forces nextTableIndex = tableWorks.size(). That means after the first oversized target, no later tables are visited, so their constraints are never emitted even when include_constraints=true. The existing coverage only exercises the byte-cap path for a single table, so this multi-table regression is currently untested.

jchauzi-floe (approved):

PR #345, opened 2026-06-23
arrow: add @ArrowDecimal support for BigDecimal record components

jchauzi-floe:
## What

Adds Arrow `DECIMAL` support to the reflection-backed `ArrowRecordWriters.fromRecordClass`, via a new `@ArrowDecimal(precision, scale)` record-component annotation.

Until now the writer's type switch covered `String`/`byte[]`/`boolean`/`int`/`long`/`double`/`Instant`/`LocalDate`/`UUID` and threw `"Unsupported component type"` for anything else — including `BigDecimal`. There was no way to emit an Arrow decimal column from a record class, even though the rest of the arrow module (`ArrowDecimalTypes` / `ArrowSchemaUtil`, the scanner path) already knows how to build one. `BigDecimal` carries no precision/scale, so the annotation supplies that metadata; everything else routes through the existing decimal machinery.

## Changes

- **`ArrowDecimal`** (new annotation) — `@ArrowDecimal(precision, scale)`, `@Target(RECORD_COMPONENT)`, mirroring the existing `@ArrowFieldName`.
- **`ArrowRecordWriters`** — a `BigDecimal` branch that:
  - validates precision/scale through floecat's canonical `LogicalType.decimal(precision, scale)` (rejects `precision < 1`, `scale < 0`, `scale > precision`);
  - builds the field type via the shared `ArrowDecimalTypes.decimalType(precision, scale)` helper (the same one `ArrowSchemaUtil` uses);
  - writes through `DecimalVector` (128-bit) **or** `Decimal256Vector` (precision > 38), matching what `ArrowDecimalTypes` produces and how the scanner's `ArrowConversion` handles decimals;
  - coerces an input whose scale exceeds the column via the annotation's `rounding()` mode, which **defaults to `RoundingMode.UNNECESSARY`** (throws rather than silently dropping precision) — lossy rounding is an explicit caller opt-in (e.g. `rounding = HALF_UP`); padding to a larger scale is always lossless;
  - rejects an **unannotated** `BigDecimal` with a specific `"must be annotated with @ArrowDecimal"` error (Arrow decimals need explicit precision/scale, so reject-unless-told is preserved).
- **`docs/types.md`** — new `Record reflection writer` subsection documenting the `@ArrowDecimal` precision/scale mapping (Decimal128 ≤ 38, Decimal256 39..76) and the rounding contract.
- **`ArrowRecordWritersTest`** — annotated-decimal round-trip (schema precision/scale, value, `HALF_UP` rounding, null), a 256-bit / high-precision round-trip exercising the `Decimal256Vector` path, unannotated-rejection, and invalid-precision/scale rejection; the unsupported-type test now uses `Object` since `BigDecimal` is conditionally supported.

Backward-compatible: behavior is identical for every non-`BigDecimal` record, and `fromRecordClass` has no other callers in floecat.

## Why

`sys.query` exposes duration columns as `DECIMAL(18,3)`. The consumer-side wire schema is built from a record (`QueryRow`) through this writer, so it needs a decimal branch. core PR eng-floe/core#1541 currently works around this with a forked copy of the writer in core; this lets that PR annotate `QueryRow` and delete the fork.

## Verification

`mvn -pl core/arrow test` — all `core/arrow` tests pass (`ArrowRecordWritersTest` = 10, covering the 128-bit and 256-bit paths, HALF_UP rounding, default-UNNECESSARY padding/excess-scale rejection, nulls, and unannotated/invalid rejection). `fmt:check` clean.

### DIFF
--- a/core/arrow/src/main/java/ai/floedb/floecat/arrow/ArrowRecordWriters.java
+++ b/core/arrow/src/main/java/ai/floedb/floecat/arrow/ArrowRecordWriters.java
@@ -198,6 +198,7 @@ public final class ArrowRecordWriters {
         }
         int precision = decimal.precision();
         int scale = decimal.scale();
+        RoundingMode rounding = decimal.rounding();
         // Validate precision/scale through floecat's canonical decimal rules before building
         // the Arrow type (ArrowType.Decimal itself does not validate these).
         LogicalType.decimal(precision, scale);
@@ -205,7 +206,10 @@ public final class ArrowRecordWriters {
         writer =
             (root, i, row) -> {
               BigDecimal val = get(getter, row, BigDecimal.class);
-              BigDecimal scaled = (val == null) ? null : val.setScale(scale, RoundingMode.HALF_UP);
+              // Padding to a larger scale is lossless; excess scale uses the annotation's rounding
+              // mode (UNNECESSARY by default, which throws rather than silently dropping
+              // precision).
+              BigDecimal scaled = (val == null) ? null : val.setScale(scale, rounding);
               // ArrowDecimalTypes uses a 256-bit vector for precision > 38, a 128-bit one
               // otherwise.
               FieldVector v = root.getVector(name);
--- a/core/arrow/src/main/java/ai/floedb/floecat/arrow/ArrowRecordWriters.java
+++ b/core/arrow/src/main/java/ai/floedb/floecat/arrow/ArrowRecordWriters.java
@@ -16,6 +16,7 @@
 
 package ai.floedb.floecat.arrow;
 
+import ai.floedb.floecat.types.LogicalType;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodType;
@@ -33,6 +34,7 @@ import java.util.UUID;
 import org.apache.arrow.vector.BigIntVector;
 import org.apache.arrow.vector.BitVector;
 import org.apache.arrow.vector.DateDayVector;
+import org.apache.arrow.vector.Decimal256Vector;
 import org.apache.arrow.vector.DecimalVector;
 import org.apache.arrow.vector.FieldVector;
 import org.apache.arrow.vector.FixedSizeBinaryVector;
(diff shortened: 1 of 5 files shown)


core/arrow/src/main/java/ai/floedb/floecat/arrow/ArrowRecordWriters.java
```diff
@@ -185,6 +188,24 @@ private static <T extends Record> SchemaAndWriters<T> buildSchemaAndWriters(Clas
...
+        ArrowDecimal decimal = comp.getAnnotation(ArrowDecimal.class);
+        if (decimal == null) {
+          throw new IllegalArgumentException(
+              "BigDecimal component " + name + " must be annotated with @ArrowDecimal");
+        }
+        int scale = decimal.scale();
+        field = field(name, ArrowDecimalTypes.decimalType(decimal.precision(), scale));
+        writer =
+            (root, i, row) -> {
+              DecimalVector v = (DecimalVector) root.getVector(name);
```
mrouvroy-floe: This path can produce a `Decimal256Vector`: `ArrowDecimalTypes.decimalType()` returns bit width 256 for precision 39..76, but the writer always casts the materialized vector to `DecimalVector`. An annotated field like `@ArrowDecimal(precision = 50, scale = 2)` will build the schema successfully and then fail at write time with `ClassCastException`.

Either reject precision > 38 here with a clear `IllegalArgumentException`, or handle `Decimal256Vector` alongside `DecimalVector`. Please add a regression test for the chosen behavior.

core/arrow/src/main/java/ai/floedb/floecat/arrow/ArrowRecordWriters.java
```diff
@@ -185,6 +188,24 @@ private static <T extends Record> SchemaAndWriters<T> buildSchemaAndWriters(Clas
...
+              BigDecimal val = get(getter, row, BigDecimal.class);
+              if (val == null) {
+                v.setNull(i);
+              } else {
+                v.setSafe(i, val.setScale(scale, RoundingMode.HALF_UP));
```
mrouvroy-floe: Is silent HALF_UP rounding intended for this adapter? Arrow’s decimal writer normally rejects scale mismatches, so this changes input values before writing. For catalog/stat data movement, I’d expect either exact-scale input or an explicit caller-controlled rounding policy. If rounding is intended, can we document that contract on `@ArrowDecimal`; otherwise this should probably reject excess scale.

mrouvroy-floe:
We might need to update floecat/docs/types.md too wrt to new precision & rounding behavior

core/arrow/src/main/java/ai/floedb/floecat/arrow/ArrowRecordWriters.java
jchauzi-floe: Fixed in `6e062dc2` — the writer now handles `Decimal256Vector` (precision > 38) alongside `DecimalVector`, matching what `ArrowDecimalTypes.decimalType` produces and how the scanner's `ArrowConversion` branches. Added a regression test `fromRecordClass_writesDecimal256ForHighPrecision` (`@ArrowDecimal(precision = 50, scale = 2)` with a 40-significant-digit value that round-trips through the 256-bit path).

core/arrow/src/main/java/ai/floedb/floecat/arrow/ArrowRecordWriters.java
jchauzi-floe: Good catch — silent rounding was the wrong default for a shared writer. As of `12ed2423`, `@ArrowDecimal` takes a `rounding()` that defaults to `RoundingMode.UNNECESSARY`: an input whose scale exceeds the column now throws (`ArithmeticException`) rather than silently dropping precision, while padding to a larger scale stays lossless. Lossy rounding is an explicit caller opt-in (e.g. `rounding = HALF_UP`).

`sys.query` genuinely needs it — `cache_efficiency` is `new BigDecimal(double)` (~16 fractional digits) and the `getOtherTimeMS` timings divide ns by 1e6 with no scale cap — so those columns set `rounding = HALF_UP` in eng-floe/core#1541. Tests cover both the default padding-OK and excess-scale-throws behavior.

jchauzi-floe:
Done — added a `Record reflection writer` subsection to `docs/types.md` covering the `@ArrowDecimal` precision/scale + Decimal128/256 mapping and the rounding contract (default `UNNECESSARY`, explicit opt-in). Pushed in a2785be0.

mrouvroy-floe (approved):

kwesterfeld-floe (approved):

PR #350, opened 2026-06-24
Add Floecat query diagnostics 

mrouvroy-floe:
## Summary
Add reusable telemetry diagnostics with readable RPC summary events and debug-gated store spans.
Add repository/store operation summaries for repo gets, lists, counts, writes, fallbacks, and store latency.
Add query path diagnostics for GetUserObjects, scan init, name resolution, snapshot pinning, and storage authority resolution.
Add Flight stream diagnostics and OTel context propagation across Flight worker threads.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist


mcusack-floe:
The repository diagnostics will overstate store totals because they instrument both public repository APIs and the internal helpers they call. 

For example, GenericResourceRepository.getByKey() wraps get() (service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java:59-60), while BaseResourceRepository.get() is itself wrapped (service/src/main/java/ai/floedb/floecat/service/repo/util/BaseResourceRepository.java:123-157). Likewise create() wraps putBlob() (GenericResourceRepository.java:82-120, BaseResourceRepository.java:201-236) and update() wraps getByKey() plus putBlob(). 

Because StoreOperationSummary sums every observed scope, the reported store_ms and op counts no longer approximate wall-clock time or logical repository operations, which makes the new trace summaries misleading when diagnosing latency.

mcusack-floe:
CreateSnapshot / UpdateSnapshot now acknowledge success even when the current-snapshot-id advance fails, which can leave SS_CURRENT permanently stale. SnapshotServiceImpl.maybeAdvanceCurrentSnapshot() catches and suppresses every RuntimeException, and that helper is called on the successful create/update paths. 

This is a behavioral regression from the previous code, which aborted the RPC on bounded CAS failure. After this change, clients can receive a successful snapshot mutation response while subsequent GetSnapshot(SS_CURRENT) still returns the older snapshot.

The reconcile finalizer has the same correctness hole: it reports the snapshot finalized successfully even if advancing the table’s current snapshot pointer fails. SnapshotFinalizeReconcileExecutor calls advanceCurrentSnapshot() before returning success.

mrouvroy-floe:
Move the snapshot improvement out of this for easier review process. 
See: https://github.com/eng-floe/floecat/pull/352

telemetry-hub/core/src/main/java/ai/floedb/floecat/telemetry/helpers/StoreMetrics.java
```diff
@@ -55,26 +56,40 @@ public ObservationScope observe(Tag... extraTags) {
...
       this.metricsScope = metricsScope;
       this.traceScope = traceScope;
+      this.component = component;
+      this.operation = operation;
+      this.recordInSummary = StoreOperationSummary.enterScope();
     }
 
     @Override
     public void success() {
+      error = null;
```
mcusack-floe, line 86: We record every summary scope as successful whenever error == null, but the underlying observation model explicitly treats “closed without success() or error()” as result=unknown in non-strict mode.

We should create a variable successCalled, set true in success(), and then in close() pass successCalled && error == null to StoreOperationSummary.record() instead of error == null.

Otherwise any caller that forgets to set an outcome will now emit conflicting telemetry: the timer/counter says unknown, while the new trace summary says success.

mcusack-floe:
The repository wrappers still double-count store telemetry even after StoreOperationSummary.enterScope() was added. 

For example, getByKey() now opens its own observation scope and then calls the already-instrumented base get() ([GenericResourceRepository.java (line 59)], [BaseResourceRepository.java (line 123)]. create() does the same around putBlob() ([GenericResourceRepository.java (line 82)], [BaseResourceRepository.java (line 207)].

mcusack-floe (approved):

PR #352, opened 2026-06-26
Use a dedicated current snapshot pointer for O(1) current snapshot lookup

mrouvroy-floe:
## Summary
This moves Floecat’s internal “current snapshot” state out of table properties and into a dedicated `CurrentSnapshotPointer` resource.

Current snapshot lookup now uses:

```text
table_id -> CurrentSnapshotPointer -> snapshot_id -> Snapshot
```

instead of scanning/listing snapshots or relying on table `current-snapshot-id`.

Changes:
- Adds `CurrentSnapshotPointer` proto/resource/repository.
- Updates `SnapshotRepository.getCurrentSnapshot()` to use the pointer first, with bounded fallback only when the pointer is missing or dangling.
- Updates snapshot create/update/delete/finalize paths to advance or clear the pointer safely.
- Prevents stale/downgrade updates by comparing snapshot ordering metadata and rechecking pointer state before CAS.
- Makes finalize-time pointer repair fail if the referenced snapshot row is missing.
- Stops internal service/gateway paths from treating table `current-snapshot-id` as authoritative.
- Keeps Iceberg REST metadata projection behavior separate from Floecat’s internal current snapshot authority.
- Removes managed table-property semantics for `current-snapshot-id`.
- Adds regression tests for pointer lookup, downgrade prevention, CAS races, missing snapshot repair, and Iceberg REST `main` ref resolution.

## Why
Warm query/scan paths need current snapshot resolution to be O(1). 
A table property is the wrong authority for this because current snapshot state is table-level mutable state pointing to a snapshot, not connector-owned table metadata. 
A dedicated pointer gives us explicit ownership, safer CAS semantics, and avoids stale table-property split-brain behavior.

## Validation

- `make fmt`
- `mvn -pl service -am -Dtest=SnapshotServiceImplTest,SnapshotFinalizeReconcileExecutorTest -Dsurefire.failIfNoSpecifiedTests=false test`
## Type of Change

- [x] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist


protocol-gateway/iceberg-rest/src/main/java/ai/floedb/floecat/gateway/iceberg/rest/services/table/transaction/TransactionCommitSnapshotSupport.java
```diff
@@ -218,6 +241,91 @@ SnapshotChangePlan planAtomicSnapshotChanges(
...
+    if (requestedMainSnapshotId != null) {
+      return null;
+    }
+    if (existingCurrentSnapshotId == null || existingCurrentSnapshotId <= 0) {
+      return parsed.latestAddedSnapshotId();
+    }
+    return null;
+  }
+
+  private Long currentSnapshotId(Table table) {
```
mcusack-floe: We're still deriving snapshot id from table properties. Shouldn't these come from the new pointer record for snapshot id with a call to tableSupport.loadCurrentSnapshotId(table)?

mcusack-floe:
protocol-gateway/iceberg-rest/src/main/java/ai/floedb/floecat/gateway/iceberg/rest/common/TableResponseMapper.java:61 and :89 with .../TableMetadataBuilder.java:139-153 and :212-239
Load/commit responses still build TableMetadataView from table properties plus the snapshot list, while only passing the current snapshot’s metadata location separately. When current-snapshot-id is absent from table properties, TableMetadataBuilder resolves currentSnapshotId by picking the latest snapshot from the list (lines 148-153), then synthesizes refs.main from that value (lines 226-239). On a table whose current pointer intentionally references an older snapshot, the response will therefore mix the older snapshot’s metadata location with a newer current-snapshot-id/refs.main. That is a user-visible correctness bug for REST load/commit responses and metadata reconstruction.

mcusack-floe (approved):

PR #353, opened 2026-06-27
information_schema: hack fix for allowing bad JSON parse result

kwesterfeld-floe:
## Summary

There be demons.  Some of the delta catalog in prod/staging is resulting in JSON parse errors.  Instead of failing whole query for information_schema.columns, instead mark a bad row in result data so we can research/fix by observing results.

## Type of Change

- [x] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist

## Additional Notes

Will need a more comprehensive decision on how to handle this.

### DIFF
--- a/core/catalog/src/main/java/ai/floedb/floecat/systemcatalog/informationschema/ColumnsScanner.java
+++ b/core/catalog/src/main/java/ai/floedb/floecat/systemcatalog/informationschema/ColumnsScanner.java
@@ -104,6 +104,7 @@ public final class ColumnsScanner implements SystemObjectScanner {
 
   private static final Schema ARROW_SCHEMA = ArrowSchemaUtil.toArrowSchema(SCHEMA);
   private static final int ARROW_BATCH_SIZE = 512;
+  private static final String SCHEMA_ERROR_COLUMN_NAME = "__schema_error__";
 
   @Override
   public Stream<SystemObjectRow> scan(SystemObjectScanContext ctx) {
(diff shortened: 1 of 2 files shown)


kwesterfeld-floe:
Note the offending item in the list: 
```
floe=> select
floe->     table_catalog,
floe->     table_schema,
floe->     table_name,
floe->     column_name,
floe->     data_type,
floe->     ordinal_position
floe->   from information_schema.columns
floe->   where column_name = '__schema_error__'
floe->   order by table_catalog, table_schema, table_name;
 table_catalog | table_schema |      table_name      |   column_name    |                data_type                 | ordinal_position
---------------+--------------+----------------------+------------------+------------------------------------------+------------------
 floe          | obs          | floe_prod_otel_spans | __schema_error__ | ERROR: Failed to parse Delta schema JSON |                0
(1 row)
```

mrouvroy-floe (changes requested):
I think the behavior goal is right: one broken user object should not abort an information_schema/system catalog scan. But I’m not convinced the __schema_error__ fake column row is the right way of doing it.

It makes information_schema.columns contain a row that is not actually a column, uses ordinal_position = 0 as a sentinel, and puts an operational error into data_type. That works for this one scanner only because data_type is a string. It won’t generalize cleanly to other system objects, and we already have another missed path: ConstraintScanIndex.columnsById() still calls ctx.graph().tableSchema(...), so the constraint information_schema views can still fail on the same kind of bad object.

Given we don’t have a warning/diagnostics API yet, I’d prefer the PR do this instead:

centralize best-effort system-object materialization in one helper, e.g. bestEffortTableSchema(...)
use it from all system-object schema lookups, at least ColumnsScanner and ConstraintScanIndex
on per-object failure, log a structured warning with scanner/catalog/schema/table/table_id/error
skip rows for the failed object rather than emitting fake rows
keep normal query/planning paths fail-fast
Then we can add a real user-visible scan warning/diagnostics channel in a follow-up that works for rows, Arrow, and Flight.

PR #355, opened 2026-06-28
storage/aws: refresh clients after closed pools

kwesterfeld-floe:
## Summary

- Recreate S3 and DynamoDB clients when AWS reports a closed connection pool.
- Convert closed-pool storage failures into retryable service errors.
- Add InitScan failure logging before gRPC error mapping drops the stack.

## Type of Change

- [x] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist

## Additional Notes

I've only observed the pool closed problem after floecat is left running 30+ minutes.  I believe the root cause is in the s3 library, but there was also "use-after-error" pattern in dynamodb client, so I made them match.  Further, the non-conversion of errors and retry on the query path seemed wrong in light of other code.

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/query/impl/QueryScanServiceImpl.java
+++ b/service/src/main/java/ai/floedb/floecat/service/query/impl/QueryScanServiceImpl.java
@@ -67,43 +67,50 @@ public class QueryScanServiceImpl extends BaseServiceImpl implements QueryScanSe
    */
   public Uni<InitScanResponse> initScan(InitScanRequest request) {
     var L = LogHelper.start(LOG, "InitScan");
-    return run(() -> {
-          String correlationId = principal.get().getCorrelationId();
-          authz.require(principal.get(), "catalog.read");
-          String queryId = mustNonEmpty(request.getQueryId(), "query_id", correlationId);
-          if (!request.hasTableId()) {
-            throw GrpcErrors.invalidArgument(
-                correlationId, QUERY_TABLE_ID_REQUIRED, Map.of("query_id", queryId));
-          }
-          var ctx =
-              queryStore
-                  .get(queryId)
-                  .orElseThrow(
-                      () ->
-                          GrpcErrors.notFound(
-                              correlationId, QUERY_NOT_FOUND, Map.of("query_id", queryId)));
-          ResourceId tableId = request.getTableId();
-          var pin = ctx.requireSnapshotPin(tableId, correlationId);
-          var initData = scanBundles.initScan(correlationId, tableId, pin.getSnapshotId());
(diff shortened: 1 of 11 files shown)


storage/aws/src/main/java/ai/floedb/floecat/storage/kv/dynamodb/ps/KvPointerStore.java
```diff
@@ -100,6 +104,20 @@ public boolean isEmpty() {
...
-    pointers.getKvStore().dump(header).await().indefinitely();
+    await(
+        () -> {
+          pointers.getKvStore().dump(header).await().indefinitely();
+          return null;
+        });
+  }
+
+  private static <T> T await(Supplier<T> operation) {
+    try {
```
mcusack-floe, line 116: We're now wrapping all runtime exceptions which changes the semantics of non-retryable errors thrown before dynamo is even called. The wrapper should probably be narrower, ideally just processing the closed pool path.

service/src/main/java/ai/floedb/floecat/service/query/impl/QueryScanServiceImpl.java
```diff
@@ -67,43 +67,50 @@ public class QueryScanServiceImpl extends BaseServiceImpl implements QueryScanSe
...
+                          .predicates(request.getPredicatesList())
+                          .build();
+                  ScanHandle handle = queryStore.createScanSession(correlationId, session);
+                  return InitScanResponse.newBuilder()
+                      .setHandle(handle)
+                      .setTableInfo(initData.tableInfo())
+                      .build();
+                })
+            .onFailure()
+            .invoke(t -> logInitScanFailure(request, t));
```
mcusack-floe: Here we log every initScan failure at error level before status mapping via logInitScanFailure(). That includes expected client-side outcomes like INVALID_ARGUMENT, NOT_FOUND, and auth failures from the body above, so ordinary bad requests will now emit server-error stack traces. This is a logging regression and will make operational signals noisy.

mcusack-floe:
This unit test is failing:

Error:  Errors: 
Error:    KvStoreProducerTest.produces_KvStore_with_expected_table_name:32 » NullPointer Cannot invoke "io.quarkus.arc.ArcContainer.select(java.lang.Class, java.lang.annotation.Annotation[])" because the return value of "io.quarkus.arc.Arc.container()" is null

The issue is that in unit tests Arc.container() can be null. Perhaps inject Instance<DynamoDbAsyncClientManager> and avoid static Arc lookup entirely.

mrouvroy-floe (approved):

PR #357, opened 2026-06-29
Richer connector policies

mcusack-floe:
## Summary

This PR adds support for persisting richer auto-capture policy on connector reconcile policy. A connector can now store auto_capture_policy, including capture outputs, column selectors, default column scope, and max default column count. Capture-mode manual triggers inherit that persisted policy when no trigger-time capture policy is provided, or replace it for a single run with explicit trigger-time capture policy flags such as --capture, --columns, --default-cols, and --max-default-cols. Invalid persisted policies are rejected on connector writes. For manual capture-mode triggers, if neither an explicit request policy nor a stored connector capture policy is present, the request now fails instead of falling back to an implicit default.

`connector update demo-glue 
  --policy-enabled true 
  --policy-mode incremental 
  --policy-current 
  --policy-capture stats,index 
  --policy-default-cols explicit-only`

and 

`connector trigger demo-glue 
  --incremental 
  --current 
  --mode metadata-and-capture`
  
In this example, the trigger inherits the connector’s persisted capture policy. A one-off run can still override it, for example with --capture table-stats,file-stats.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist


service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
```diff
@@ -962,6 +964,35 @@ private static void validateReconcilePolicy(ReconcilePolicy policy, String corr)
...
+    if (policy.hasAutoCapturePolicy()) {
+      validateCapturePolicy(policy.getAutoCapturePolicy(), corr, "policy.auto_capture_policy");
+    }
+  }
+
+  private static void validateCapturePolicy(CapturePolicy policy, String corr, String fieldName) {
+    if (policy == null) {
+      return;
+    }
+    if (policy.getOutputsCount() == 0) {
```
mrouvroy-floe: This should also reject `CO_UNSPECIFIED` / `UNRECOGNIZED` entries, not just an empty outputs list. A client can currently persist `auto_capture_policy.outputs = [CO_UNSPECIFIED]`; that passes connector validation here, but the scheduler later throws in `ReconcilePlannerScheduler#fromProtoCaptureOutput`, so auto-reconcile for that connector fails at enqueue time. 
We should validate every persisted output enum here before accepting the connector policy.

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/ReconcilePlannerScheduler.java
```diff
@@ -359,6 +363,56 @@ private static ReconcileSnapshotSelection effectiveSnapshotSelection(Connector c
...
+      return fromProtoCapturePolicy(connector.getPolicy().getAutoCapturePolicy());
+    }
+    return DEFAULT_CAPTURE_POLICY;
+  }
+
+  private static ReconcileCapturePolicy fromProtoCapturePolicy(CapturePolicy capturePolicy) {
+    if (capturePolicy == null) {
+      return ReconcileCapturePolicy.empty();
+    }
+    return ReconcileCapturePolicy.of(
```
mrouvroy-floe, line 388: [major] **capture-policy `properties` map silently dropped in every conversion layer**

Two independent spots in the capture-policy data flow discard the `properties` map, so engine-specific parameters (e.g. `stats.ndv.sample_fraction`) are lost before they ever reach the executor.

**ReconcilePlannerScheduler (line 379):** When building a `ReconcileCapturePolicy` from the stored connector proto, the call omits `capturePolicy.getPropertiesMap()`. Any `auto_capture_policy.properties` a user persisted via the API will appear stored, but all automatic reconcile jobs execute with an empty `ReconcileCapturePolicy.properties()` map.

**GrpcRemoteReconcileExecutorClient (line 1310):** After the PR renamed the wire type from `ai.floedb.floecat.reconciler.rpc.CapturePolicy` to `ai.floedb.floecat.capture.rpc.CapturePolicy`, the `.putAllProperties(effective.properties())` call was dropped from the builder chain in `toProtoCapturePolicy`. Even if the scheduler bug is fixed, any properties still present in `ReconcileCapturePolicy` are stripped again when the policy is serialised onto the gRPC `LeasedFileGroupExecution`.

Fix the scheduler by passing `capturePolicy.getPropertiesMap()` to the `ReconcileCapturePolicy.of(...)` overload. Fix the gRPC client by restoring `.putAllProperties(effective.properties())` in the `toProtoCapturePolicy` builder chain. Add a round-trip test that asserts a non-empty `properties` map survives persist → schedule → wire.

---

🤖 _Generated with Claude Sonnet 5 and GPT-5.5._

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
```diff
@@ -962,6 +964,43 @@ private static void validateReconcilePolicy(ReconcilePolicy policy, String corr)
     if (policy.getScope() == ReconcileSnapshotScope.RSS_LATEST_N && policy.getLatestN() <= 0) {
       throw GrpcErrors.invalidArgument(corr, null, Map.of("field", "policy.latest_n"));
     }
+    if (policy.hasAutoCapturePolicy()) {
+      validateCapturePolicy(policy.getAutoCapturePolicy(), corr, "policy.auto_capture_policy");
```
mrouvroy-floe: [major] **Secret-bearing keys accepted and stored in `auto_capture_policy.properties`**

`validateCapturePolicy` checks outputs, max columns, and column selectors, but never inspects `policy.auto_capture_policy.properties` for secret-bearing keys. An API client can submit `{"s3.secret-access-key": "..."}` inside that map, and the value will be persisted to the connector record without any rejection.

`maskConnector` only masks the `auth` field, so `get` and `list` responses can return those values in plaintext, leaking credentials to any caller with read access.

Call the existing `validateSecretBearingMap` helper (or equivalent) against `policy.getPropertiesMap()` inside `validateCapturePolicy` before the record is written. Ensure `maskConnector` also redacts sensitive entries in `auto_capture_policy.properties`. Add create/update test coverage for this path.

---

🤖 _Generated with Claude Sonnet 5 and GPT-5.5._

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
```diff
@@ -962,6 +964,43 @@ private static void validateReconcilePolicy(ReconcilePolicy policy, String corr)
...
+    }
+    for (int i = 0; i < policy.getOutputsCount(); i++) {
+      var output = policy.getOutputs(i);
+      if (output == ai.floedb.floecat.capture.rpc.CaptureOutput.CO_UNSPECIFIED
+          || output == ai.floedb.floecat.capture.rpc.CaptureOutput.UNRECOGNIZED) {
+        throw GrpcErrors.invalidArgument(
+            corr, null, Map.of("field", fieldName + ".outputs[" + i + "]"));
+      }
+    }
+    if (policy.getMaxDefaultColumns() < 0) {
```
mrouvroy-floe: [minor] **Unknown `default_column_scope` enum value accepted on create/update**

`validateCapturePolicy` rejects `UNRECOGNIZED` capture outputs but does not validate `default_column_scope`. A client that sends `setDefaultColumnScopeValue(99)` will have the request accepted and the unknown value stored.

Downstream reconcile mappers silently coerce `UNRECOGNIZED` to `FIRST_N`, so the service executes a different policy than the one persisted — undermining the PR's intent of rejecting invalid stored policies.

Reject `DefaultColumnScope.UNRECOGNIZED` (and decide whether `DCS_UNSPECIFIED` should be rejected or normalised) in `validateCapturePolicy` before storage.

---

🤖 _Generated with Claude Sonnet 5 and GPT-5.5._

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/ReconcilePlannerScheduler.java
```diff
@@ -359,6 +363,57 @@ private static ReconcileSnapshotSelection effectiveSnapshotSelection(Connector c
...
+        capturePolicy.getColumnsList().stream()
+            .map(ReconcilePlannerScheduler::fromProtoCaptureColumnPolicy)
+            .toList(),
+        capturePolicy.getOutputsList().stream()
+            .map(ReconcilePlannerScheduler::fromProtoCaptureOutput)
+            .collect(java.util.stream.Collectors.toCollection(java.util.LinkedHashSet::new)),
+        fromProtoDefaultColumnScope(capturePolicy.getDefaultColumnScope()),
+        capturePolicy.getMaxDefaultColumns(),
+        capturePolicy.getPropertiesMap());
```
mrouvroy-floe: [major] **Persist capture policy properties with jobs**

`fromProtoCapturePolicy()` now copies `auto_capture_policy.properties` into the `ReconcileScope`, so scheduled connector policies can carry engine options such as `stats.ndv.sample_fraction` into queued reconcile jobs.

That value is lost as soon as the job is persisted. `StoredJobDefinition.of()` only stores columns, outputs, default scope, and max columns; `StoredJobDefinition.toScope()` reconstructs the policy with the overload that has no properties argument, and the native/memory clone helpers have no field to copy either. Any queued job that is loaded from durable storage, projected, or leased after a restart will run without the persisted capture properties even though the connector policy included them.

Add a `capturePolicyProperties` field to `StoredJobDefinition`, populate it from `effectiveScope.capturePolicy().properties()`, pass it back through `ReconcileCapturePolicy.of(..., properties)`, update the clone helpers, and add a durable definition round-trip test.

---

🤖 _Generated with GPT-5.5._

mrouvroy-floe (approved):
**[minor] Column selectors with non-per-column outputs build columns with no enabled output** — `client-cli/src/main/java/ai/floedb/floecat/client/cli/util/CliUtils.java:295`

Each `CaptureColumnPolicy` derives its per-column flags purely from the request-level outputs: `setCaptureStats(outputs.contains(CO_COLUMN_STATS))` and `setCaptureIndex(outputs.contains(CO_PARQUET_PAGE_INDEX))`. So `--columns colA --capture table-stats` (or `file-stats`) produces a column entry where both flags are `false`.

The two call sites then diverge on how that state is handled. On connector writes the server's `validateCapturePolicy` rejects it with the opaque `"column capture policy has no enabled outputs"` error, even though the user did pass `--capture`. On the trigger/analyze path there is no equivalent per-column validation, so the useless column silently flows through to the executor.

Consider validating client-side that `--columns`/`--policy-columns` is only accepted alongside `column-stats`/`index` outputs (or emitting a clear message), so the two paths behave consistently and the error is actionable. The same derivation is duplicated in `ConnectorCliSupport.parsePolicyCaptureOptions` (around line 1183).

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/ReconcileControlImpl.java
```diff
@@ -801,21 +799,97 @@ private static void requireSpecifiedCaptureMode(
...
-  private static void requireExplicitCapturePolicy(
-      CaptureMode mode, CaptureScope scope, String correlationId) {
+  private static void validateRequestCapturePolicy(CaptureScope scope, String correlationId) {
+    if (scope == null || !scope.hasCapturePolicy()) {
+      return;
+    }
+    validateCapturePolicy(scope.getCapturePolicy(), correlationId, "scope.capture_policy");
+  }
+
+  private static void validateCapturePolicy(
```
mrouvroy-floe: [minor] **`validateCapturePolicy` duplicated verbatim across two service classes**

`ReconcileControlImpl.validateCapturePolicy` (lines 809–847) is a near-identical copy of `ConnectorsImpl.validateCapturePolicy` (lines 977–1015) — same output checks, same `default_column_scope` guard, same per-column selector/enabled-output checks, differing only in the correlation-id parameter name and the field-path prefix.

Aligning by copy-paste means the two validation paths can silently diverge the next time a rule changes — exactly the problem the "Align request-side and connector-side policy validation" commit set out to prevent.

Extract a single shared static helper (e.g. `CapturePolicyValidator.validate(CapturePolicy, String correlationId, String fieldPrefix)`) that both call sites delegate to. Both classes already share the same package tree, so no new module boundary is needed.

---

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
```diff
@@ -962,6 +969,50 @@ private static void validateReconcilePolicy(ReconcilePolicy policy, String corr)
...
```
mrouvroy-floe: [nit] **`max_default_columns == 0` accepted server-side but forbidden by CLI and silently coerced downstream**

`validateCapturePolicy` only rejects `getMaxDefaultColumns() < 0`, so a policy with `max_default_columns = 0` and `default_column_scope = DCS_FIRST_N` passes server-side validation.

This creates a three-way inconsistency: the CLI rejects it (`--policy-max-default-cols must be greater than 0`); `ReconcileCapturePolicy`'s constructor silently coerces `maxDefaultColumns <= 0` to `DEFAULT_MAX_COLUMNS` (32); and the API stores and echoes `0` while effectively applying `32` — the persisted intent and the effective behavior disagree.

Reject `<= 0` in both `ConnectorsImpl.validateCapturePolicy` and the mirrored check in `ReconcileControlImpl` to align with the CLI boundary and eliminate the surprising silent coercion.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/ReconcileControlImpl.java
```diff
@@ -801,21 +799,97 @@ private static void requireSpecifiedCaptureMode(
...
+          || output == ai.floedb.floecat.capture.rpc.CaptureOutput.UNRECOGNIZED) {
+        throw GrpcErrors.invalidArgument(
+            correlationId, null, Map.of("field", fieldName + ".outputs[" + i + "]"));
+      }
+    }
+    if (policy.getMaxDefaultColumns() < 0) {
+      throw GrpcErrors.invalidArgument(
+          correlationId, null, Map.of("field", fieldName + ".max_default_columns"));
+    }
+    if (policy.getDefaultColumnScope()
```
mrouvroy-floe: [blocker] **Capture-policy validation rejects `DCS_UNSPECIFIED`, breaking existing callers and CI**

`validateCapturePolicy` treats an unset `default_column_scope` (proto3 default `DCS_UNSPECIFIED`) as invalid and throws `INVALID_ARGUMENT`. Any client that builds a `CapturePolicy` with only `outputs` set — a previously valid request — is now rejected.

This directly causes the red CI: `ci-core` and `ci-localstack` both fail `ConnectorIT` with 13+ `INVALID_ARGUMENT` errors from `runReconcile`, plus `startCaptureNotFound` failing with `expected gRPC code NOT_FOUND got INVALID_ARGUMENT`. The test helper `scopeWithCaptureOutputs` builds a policy without setting `default_column_scope`, so every capture path through it trips this check before the connector is even resolved. External API callers that construct the same minimal policy will hit the same regression.

The rejection also contradicts the code's own mapping: `fromProtoDefaultColumnScope` / `scopeFromCaptureScope` immediately below already maps `DCS_UNSPECIFIED` to `FIRST_N`, meaning the rest of the service has always treated an omitted scope as a valid default. The validation gate should either allow `DCS_UNSPECIFIED` and let the existing conversion supply the default, or explicitly document that callers must set a non-default scope — the latter being a breaking API change.

---

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
```diff
@@ -962,6 +969,50 @@ private static void validateReconcilePolicy(ReconcilePolicy policy, String corr)
...
+      throw GrpcErrors.invalidArgument(
+          corr, null, Map.of("field", fieldName + ".max_default_columns"));
+    }
+    if (policy.getDefaultColumnScope()
```
mrouvroy-floe: [major] **Connector-write validation duplicates `DCS_UNSPECIFIED` rejection, diverging from reconcile semantics**

The connector-side `validateCapturePolicy` applies the same rejection of `DCS_UNSPECIFIED` for `auto_capture_policy.default_column_scope`. While the CLI currently always populates this field (defaulting to `DCS_FIRST_N`), a programmatic `createConnector` or `updateConnector` that sets `auto_capture_policy` with only `outputs` will be rejected even though the reconcile path treats an unset scope as `FIRST_N`.

This creates an inconsistency between the write boundary (`ConnectorsImpl`) and the read/execute boundary (`ReconcileControlImpl`): data written through one path may be invalid according to the other. Relaxing or aligning these checks should be done in lockstep with the request-side fix.

The two `validateCapturePolicy` copies are near-identical; extracting a shared helper would prevent them from drifting again in the future.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/ReconcileControlImpl.java
mcusack-floe: Fixed in e64277872 and pushed. DCS_UNSPECIFIED and max_default_columns=0 are now accepted as omitted proto defaults, preserving the existing effective FIRST_N / default-max-columns behavior; UNRECOGNIZED still fails explicitly. I also removed the test-helper injection of those defaults so ConnectorIT exercises the actual omitted-field path. Verified with make test TEST=ConnectorIT (45 passed, 3 skipped) plus the focused service/reconciler suite (78 passed).

service/src/main/java/ai/floedb/floecat/service/common/CapturePolicyValidator.java
```diff
@@ -0,0 +1,66 @@
...
+          correlationId, null, Map.of("field", fieldName + ".outputs"));
+    }
+    for (int i = 0; i < policy.getOutputsCount(); i++) {
+      var output = policy.getOutputs(i);
+      if (output == CaptureOutput.CO_UNSPECIFIED || output == CaptureOutput.UNRECOGNIZED) {
+        throw GrpcErrors.invalidArgument(
+            correlationId, null, Map.of("field", fieldName + ".outputs[" + i + "]"));
+      }
+    }
+    if (policy.getMaxDefaultColumns() < 0) {
```
mrouvroy-floe, line 43: [nit] **`getMaxDefaultColumns() < 0` is not a meaningful range check**

`max_default_columns` is a proto3 `uint32`, whose generated Java getter returns an `int`. For every realistic input (`0` .. `2^31-1`) this value is non-negative, so the guard never fires; it only trips for values above `2^31` that wrap to a negative `int` — i.e. it rejects "absurdly large" rather than "negative".

The net effect is a validation line that reads like a lower-bound check but effectively validates nothing for normal use, which is misleading to future readers.

Either drop the check, or if the intent is to reject `0`/unreasonably-large values, validate that explicitly (e.g. `Integer.toUnsignedLong(policy.getMaxDefaultColumns())` against a sane upper bound). Note the downstream `ReconcileCapturePolicy` already normalizes `<= 0` to `DEFAULT_MAX_COLUMNS`, so a `0` here silently becomes 32 rather than "zero columns" — worth confirming that is the intended semantics.

---

mrouvroy-floe (approved):

PR #358, opened 2026-06-29
Set auto-reconcile to tick every 500ms

mcusack-floe:
## Summary

Set default auto-reconcile to run every 0.5s by default. This is still overridable in the environment or properties.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist


mrouvroy-floe (dismissed):

service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java
```diff
@@ -457,27 +457,21 @@ public boolean delete(K key) {
             return false;
           }
           String blobUri = resolveBlobUriForDelete(key, canonicalPointer);
+          T currentValue =
+              getByKeyUnobserved(key)
```
mrouvroy-floe: [major] **Delete aborts on corrupt blob, leaving resources permanently stuck**

`delete()` and `deleteWithPrecondition()` now call `getByKeyUnobserved()` before issuing any pointer delete. If the canonical pointer exists but its blob is missing or unparsable, `BaseResourceRepository.read()` throws `CorruptionException`, aborting the delete before the canonical pointer is removed.

This is a regression from the previous implementation, which explicitly caught `CorruptionException` and still deleted the canonical pointer. Once a resource (table, connector, namespace, etc.) has a dangling or corrupt blob, users and GC can no longer remove it through the repository API — the resource becomes permanently stuck behind 500s/corruption errors with no recovery path.

Preserve the new atomic secondary-pointer cleanup (via `compareAndSetBatch`) for the normal case, but restore the old recovery behavior when `CorruptionException` is thrown: catch it, skip the blob-dependent steps, and still delete the canonical pointer so the resource can be cleaned up.

---

protocol-gateway/iceberg-rest/src/main/java/ai/floedb/floecat/gateway/iceberg/rest/services/table/transaction/TransactionCommitSnapshotSupport.java
```diff
@@ -153,6 +154,19 @@ SnapshotChangePlan planAtomicSnapshotChanges(
         }
       }
     }
+    if (!writesNewSnapshot
+        && resolvedMetadataLocation != null
+        && (updates == null || updates.isEmpty() || !parsed.containsSnapshotUpdates())) {
```
mrouvroy-floe: [minor] **`metadata-location` rewrite skipped for remove-snapshot-only commits — confirm intended**

The current-snapshot `metadata-location` rewrite guard now includes the clause `&& (updates == null || updates.isEmpty() || !parsed.containsSnapshotUpdates())`. Previously the block fired whenever `!writesNewSnapshot && resolvedMetadataLocation != null && !isBlank`.

Because `REMOVE_SNAPSHOTS`, `SET_SNAPSHOT_REF`, and `REMOVE_SNAPSHOT_REF` all set `containsSnapshotUpdates = true` in `CommitUpdateInspector`, a commit that supplies a new `metadata-location` alongside only remove-snapshot/ref updates (no `ADD_SNAPSHOT`) will now skip `rewriteCurrentSnapshotMetadata`. An expire-snapshots commit is exactly this shape and does produce a new `metadata.json`, so the stored `metadataLocation` on the current snapshot pointer may go stale.

Confirm this is intentional; if not, refine the guard so it only excludes commits that genuinely have no snapshot-related updates.

---

protocol-gateway/iceberg-rest/src/main/java/ai/floedb/floecat/gateway/iceberg/rest/services/table/transaction/TransactionCommitService.java
```diff
@@ -114,204 +115,233 @@ private Response commitInternal(
...
+        if (!snapshotChangePlan.txChanges().isEmpty()) {
+          txChanges.addAll(snapshotChangePlan.txChanges());
+        }
       }
-      if (!snapshotChangePlan.txChanges().isEmpty()) {
-        txChanges.addAll(snapshotChangePlan.txChanges());
+
+      Response applyResponse = transactionCommitExecutionSupport.apply(openTransaction, txChanges);
+      if (!shouldRetryTransactionConflict(attempt, applyResponse)) {
+        return applyResponse;
```
mrouvroy-floe: [minor] **Internal `floecat.retryable-transaction-conflict` header leaks to the client on exhausted retries**

`apply()` attaches the internal HTTP header `floecat.retryable-transaction-conflict: true` to conflict responses (see `TransactionCommitExecutionSupport.conflictResponse` / `retryableConflict`, lines 316–323). This header is only meaningful internally to drive the replan loop via `isRetryableConflict`.

On the final attempt (`MAX_INTERNAL_REPLAN_ATTEMPTS = 3`), `shouldRetryTransactionConflict` returns `false` and `commit()` returns `applyResponse` verbatim at line 339 — header included. The `Response` flows straight out through JAX-RS, so a commit that exhausts all retries returns a 409 to the caller with this internal header exposed.

Strip the header (or use a dedicated response builder) before returning the final conflict response to the client.

---

mrouvroy-floe (commented):
**[nit] Dead code: private `firstNonBlank(String...)` is now unused** — `protocol-gateway/iceberg-rest/src/main/java/ai/floedb/floecat/gateway/iceberg/rest/services/table/transaction/TransactionCommitExecutionSupport.java:469`

Both former call sites were removed in this PR (`beginIdempotency` now uses `trimToNull`, and `OpenTransaction` construction passes `txId` directly). The private static `firstNonBlank(String...)` in this class has no remaining callers. Other classes define their own copies; only this class's copy is orphaned. Remove it to keep the class tidy.

---

mcusack-floe:
This has crept in size because switching to 0.5s reconcile interval exposed issues in OCC conflict handling during table commit, in reconcile queue deduplication and in table metadata handling. These are all addressed now.

PR #360, opened 2026-07-02
Add refresh on connector pools for sync clients

mcusack-floe:
## Summary

Connection pools to DynamoDB, STS and Secrets manager on synchronous clients time out after a period of inactivity. This change adds a connection pool refresh.

## Type of Change

- [X] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/DynamoReconcileJobIndexBackend.java
+++ b/service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/DynamoReconcileJobIndexBackend.java
@@ -21,8 +21,11 @@ import static ai.floedb.floecat.storage.kv.KvAttributes.ATTR_PARTITION_KEY;
 import static ai.floedb.floecat.storage.kv.KvAttributes.ATTR_SORT_KEY;
 import static ai.floedb.floecat.storage.kv.KvAttributes.ATTR_VERSION;
 
+import ai.floedb.floecat.storage.aws.ClosedAwsClientDetector;
+import ai.floedb.floecat.storage.aws.DynamoDbClientManager;
 import ai.floedb.floecat.storage.kv.KvAttributes;
 import io.quarkus.arc.properties.IfBuildProperty;
+import jakarta.enterprise.inject.Instance;
 import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
 import java.util.ArrayList;
@@ -30,6 +33,9 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
 import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
@@ -46,6 +52,10 @@ import software.amazon.awssdk.services.dynamodb.model.TransactionCanceledExcepti
 @IfBuildProperty(name = "floecat.kv", stringValue = "dynamodb")
 public class DynamoReconcileJobIndexBackend implements ReconcileJobIndexBackend {
   @Inject DynamoDbClient dynamoDb;
+  @Inject Instance<DynamoDbClientManager> dynamoDbClientManager;
+  private volatile Supplier<DynamoDbClient> dynamoDbSupplier;
+  private volatile BiConsumer<DynamoDbClient, Throwable> clientFailureHandler =
+      (client, failure) -> {};
 
   @ConfigProperty(name = "floecat.kv.table", defaultValue = "floecat_pointers")
   String table = "floecat_pointers";
@@ -55,7 +65,71 @@ public class DynamoReconcileJobIndexBackend implements ReconcileJobIndexBackend 
 
   public void bind(DynamoDbClient dynamoDb, String table) {
     this.dynamoDb = dynamoDb;
+    this.dynamoDbSupplier = () -> dynamoDb;
+    this.clientFailureHandler = (client, failure) -> {};
     this.table = table;
+  }
+
+  public void bind(
+      Supplier<DynamoDbClient> dynamoDbSupplier,
+      String table,
+      BiConsumer<DynamoDbClient, Throwable> clientFailureHandler) {
+    this.dynamoDbSupplier = dynamoDbSupplier;
+    this.clientFailureHandler =
+        clientFailureHandler == null ? (client, failure) -> {} : clientFailureHandler;
+    this.table = table;
+  }
+
+  private DynamoDbClient dynamoDb() {
+    if (dynamoDbSupplier != null) {
+      return dynamoDbSupplier.get();
+    }
+    if (dynamoDbClientManager != null && dynamoDbClientManager.isResolvable()) {
+      DynamoDbClientManager manager = dynamoDbClientManager.get();
+      dynamoDbSupplier = manager::current;
+      clientFailureHandler = manager::refreshAfterFailure;
+      return dynamoDbSupplier.get();
+    }
+    return dynamoDb;
+  }
+
+  private boolean refreshAfterFailure(DynamoDbClient client, Throwable failure) {
+    if (ClosedAwsClientDetector.isConnectionPoolShutdown(failure)) {
+      clientFailureHandler.accept(client, failure);
+      return true;
+    }
+    return false;
+  }
+
+  private <T> T callDynamo(Function<DynamoDbClient, T> operation) {
+    for (int attempt = 0; ; attempt++) {
+      DynamoDbClient client = dynamoDb();
+      try {
+        return operation.apply(client);
+      } catch (RuntimeException e) {
+        boolean refreshed = refreshAfterFailure(client, e);
+        if (refreshed && attempt == 0) {
+          continue;
+        }
+        throw e;
+      }
+    }
+  }
+
+  private void callDynamoVoid(java.util.function.Consumer<DynamoDbClient> operation) {
+    for (int attempt = 0; ; attempt++) {
+      DynamoDbClient client = dynamoDb();
+      try {
+        operation.accept(client);
+        return;
+      } catch (RuntimeException e) {
+        boolean refreshed = refreshAfterFailure(client, e);
+        if (refreshed && attempt == 0) {
+          continue;
+        }
+        throw e;
+      }
+    }
   }
 
   @Override
@@ -233,19 +307,21 @@ public class DynamoReconcileJobIndexBackend implements ReconcileJobIndexBackend 
       if (exclusiveStartKey != null && !exclusiveStartKey.isEmpty()) {
         request.exclusiveStartKey(exclusiveStartKey);
       }
-      var response = dynamoDb.scan(request.build());
+      var response = callDynamo(client -> client.scan(request.build()));
       for (var item : response.items()) {
         if (item == null || item.isEmpty()) {
           continue;
         }
-        dynamoDb.deleteItem(
-            software.amazon.awssdk.services.dynamodb.model.DeleteItemRequest.builder()
-                .tableName(table)
-                .key(
-                    Map.of(
-                        ATTR_PARTITION_KEY, item.get(ATTR_PARTITION_KEY),
-                        ATTR_SORT_KEY, item.get(ATTR_SORT_KEY)))
-                .build());
+        callDynamoVoid(
+            client ->
+                client.deleteItem(
+                    software.amazon.awssdk.services.dynamodb.model.DeleteItemRequest.builder()
+                        .tableName(table)
+                        .key(
+                            Map.of(
+                                ATTR_PARTITION_KEY, item.get(ATTR_PARTITION_KEY),
+                                ATTR_SORT_KEY, item.get(ATTR_SORT_KEY)))
+                        .build()));
         deleted = true;
       }
       exclusiveStartKey = response.lastEvaluatedKey();
@@ -283,7 +359,10 @@ public class DynamoReconcileJobIndexBackend implements ReconcileJobIndexBackend 
       }
     }
     try {
-      dynamoDb.transactWriteItems(TransactWriteItemsRequest.builder().transactItems(tx).build());
+      callDynamoVoid(
+          client ->
+              client.transactWriteItems(
+                  TransactWriteItemsRequest.builder().transactItems(tx).build()));
       return true;
     } catch (TransactionCanceledException e) {
       return false;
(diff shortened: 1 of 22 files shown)


mcusack-floe:
Refresh sync and async DynamoDB clients by failed client identity so stale
closed-pool failures cannot replace a newer healthy client. Retry sync durable
DynamoDB backend operations once after refreshing the client, and ensure the
ready-queue backend uses the manager-backed production path instead of pinning
the injected client.

Add regression coverage for stale failed-client refresh handling and ready-queue
closed-pool retry behavior.

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/DynamoReconcileJobIndexBackend.java
```diff
@@ -55,9 +65,73 @@ public DynamoReconcileJobIndexBackend() {}
...
+    this.table = table;
+  }
+
+  private DynamoDbClient dynamoDb() {
+    if (dynamoDbSupplier != null) {
+      return dynamoDbSupplier.get();
+    }
+    if (dynamoDbClientManager != null && dynamoDbClientManager.isResolvable()) {
+      DynamoDbClientManager manager = dynamoDbClientManager.get();
+      dynamoDbSupplier = manager::current;
```
mrouvroy-floe: [minor] **Lazy init publishes `dynamoDbSupplier` before `clientFailureHandler`, allowing a concurrent failure to retry without refreshing**

In the lazy fallback of `dynamoDb()`, `dynamoDbSupplier` is assigned before `clientFailureHandler`. Both fields are `volatile` but the two writes are not atomic together, so a thread that observes the manager-backed supplier while `clientFailureHandler` is still the no-op default will call `refreshAfterFailure`, get `true` (the detector matched), but invoke a handler that refreshes nothing. The retry then calls `manager.current()`, receives the same stale client, fails again, and re-throws — defeating the entire healing mechanism.

The same shape exists in the legacy `bind(DynamoDbClient, String)` path, where the handler is permanently a no-op: `refreshAfterFailure` still returns `true` and the loop retries a permanently dead client.

Assign `clientFailureHandler` before (or atomically with) `dynamoDbSupplier`, or use a single compound object so both are visible together. The legacy `bind` path should either set a real handler or stop signalling `true` from `refreshAfterFailure`.

---

🤖 _Generated with Fable 5 and GPT-5.5._

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/DynamoReconcileLeaseBackend.java
```diff
@@ -61,9 +71,73 @@ public DynamoReconcileLeaseBackend() {}
...
+    if (dynamoDbClientManager != null && dynamoDbClientManager.isResolvable()) {
+      DynamoDbClientManager manager = dynamoDbClientManager.get();
+      dynamoDbSupplier = manager::current;
+      clientFailureHandler = manager::refreshAfterFailure;
+      return dynamoDbSupplier.get();
+    }
+    return dynamoDb;
+  }
+
+  private boolean refreshAfterFailure(DynamoDbClient client, Throwable failure) {
```
mrouvroy-floe: [minor] **~70 lines of refresh/retry plumbing copy-pasted verbatim across three backends**

`bind(Supplier, String, BiConsumer)`, `dynamoDb()`, `refreshAfterFailure()`, `callDynamo()`, and `callDynamoVoid()` are duplicated identically in `DynamoReconcileJobIndexBackend`, `DynamoReconcileLeaseBackend`, and `DynamoReconcileReadyQueueBackend`.

This triples the surface for the ordering bug noted above — a fix applied to one backend can silently miss the other two. Current test coverage reflects that risk: only the ready-queue backend has a closed-pool retry regression test.

Extract a small helper (e.g. `RefreshingDynamoCaller` holding the supplier + failure handler with `call`/`callVoid` methods) that all three backends delegate to, so the retry semantics live and are tested in exactly one place.

---

🤖 _Generated with Fable 5 and GPT-5.5._

mrouvroy-floe (commented):
**[major] STS and Secrets Manager sync clients not refreshed — part of the production failure remains unfixed** — `storage/aws/src/main/java/ai/floedb/floecat/storage/aws/secrets/ProdSecretsManager.java:174`

The PR description states the closed-pool timeout affects DynamoDB, STS, and Secrets Manager sync clients, but only the DynamoDB client gets refresh handling. `ProdSecretsManager` still holds `secretsClient`, `stsClient`, and every entry in `perAccountClients` as long-lived final fields created once at construction time.

When any of those clients' connection pools idle out and close, `put`, `get`, `update`, `delete`, and the per-account `StsAssumeRoleCredentialsProvider` path will keep reusing the same stale client and throw `Connection pool shut down` indefinitely — there is no recovery path. `StorageAuthorityResolver` is exempt because it constructs its `StsClient` per call, but `ProdSecretsManager` is not.

Add the same refresh-on-closed-pool behavior (manager or supplier + failure handler) for `SecretsManagerClient` and `StsClient`, and thread it through `perAccountClients` so the per-account assume-role credential providers also pick up fresh clients after a pool shutdown.

---

🤖 _Generated with Fable 5 and GPT-5.5._

mrouvroy-floe (changes requested):

storage/aws/src/main/java/ai/floedb/floecat/storage/aws/secrets/ProdSecretsManager.java
```diff
@@ -200,6 +230,50 @@ private SecretsManagerClient clientForAccount(String accountId) {
...
+      return;
+    }
+    if (!perAccountClients.remove(accountId, failedClient)) {
+      return;
+    }
+    closeQuietly(failedClient);
+    StsClient previousSts = stsClient.get();
+    StsClient nextSts = awsClients.stsClient();
+    if (stsClient.compareAndSet(previousSts, nextSts)) {
+      closeQuietly(previousSts);
```
mrouvroy-floe: [major] **Refreshing one per-account client closes the shared `StsClient` still referenced by every other cached provider**

In the role-ARN path, `refreshAfterClosedPool` swaps `stsClient` and closes the previous instance whenever **any** per-account secrets client hits a closed pool — even when the failure came from that client's own Secrets Manager connection pool and the STS client is healthy. Every other entry in `perAccountClients` was built with a `StsAssumeRoleCredentialsProvider` that captured `stsClient.get()` at `computeIfAbsent` time (line 214), and those cached entries are not invalidated.

The next credential refresh on each of those providers now fails with `Connection pool shut down` against the deliberately-closed STS client, which triggers *another* STS swap-and-close, re-breaking any providers rebuilt in the meantime — a cascade across accounts. The single retry in `withClientRefresh` masks most of this, but it can still escape to callers: thread B rebuilds its client capturing the current STS, thread C then swaps and closes that STS before B's assume-role call runs, and B's one retry fails with no further recovery. With N active accounts, one idle-timeout event produces O(N) induced failures and error-log noise even in the happy case.

Concrete options, in rough order of preference:

- When swapping the STS client, also remove and close **all** `perAccountClients` entries (they are all bound to the old STS), so nothing keeps referencing the closed client.
- Give each per-account provider its own `StsClient`, closed together with its secrets client, so refresh is contained to one account.
- Only refresh STS when the STS client is actually implicated in the failure, rather than on every per-account pool shutdown.

---

🤖 _Generated with Fable 5 and GPT-5.5._

storage/aws/src/main/java/ai/floedb/floecat/storage/aws/secrets/ProdSecretsManager.java
```diff
@@ -163,25 +194,24 @@ private static List<Tag> buildTags(String accountId, String secretName) {
...
-    clientForAccount(accountId)
-        .tagResource(TagResourceRequest.builder().secretId(secretName).tags(tags).build());
+    client.tagResource(TagResourceRequest.builder().secretId(secretName).tags(tags).build());
   }
 
   private SecretsManagerClient clientForAccount(String accountId) {
     String role = roleArn.map(String::trim).filter(value -> !value.isBlank()).orElse("");
     if (role.isEmpty()) {
-      return secretsClient;
+      return secretsClient.get();
```
mrouvroy-floe: [minor] **NPE instead of a clear error when operations race `@PreDestroy` shutdown**

`shutdown()` now does `secretsClient.getAndSet(null)` / `stsClient.getAndSet(null)`. A `get`/`put` racing shutdown (or arriving after it) gets `null` back from `clientForAccount` in the no-role path and fails with a bare `NullPointerException` from `operation.run(null)`; in the role path `stsClient.get()` returns `null` into the provider builder.

Before this change the same race surfaced as a closed-client exception from the AWS SDK, which was at least self-explanatory. The NPE also doesn't match `ClosedAwsClientDetector`, so it propagates without any refresh attempt.

Consider throwing an explicit `IllegalStateException("secrets manager is shut down")` when the reference is `null`, or leaving the closed client in place instead of nulling the reference.

---

🤖 _Generated with Fable 5 and GPT-5.5._

storage/aws/src/main/java/ai/floedb/floecat/storage/kv/dynamodb/DynamoDbKvStore.java
```diff
@@ -38,18 +40,23 @@
...
   static final int DELETE_BATCH_LIMIT = Integer.getInteger("floedb.floecat.delete.batch.size", 25);
   private final Supplier<DynamoDbAsyncClient> ddb;
-  private final Consumer<Throwable> clientFailureHandler;
+  private final BiConsumer<DynamoDbAsyncClient, Throwable> clientFailureHandler;
+  private final boolean supportsClientRefresh;
   private final String table;
 
   public DynamoDbKvStore(DynamoDbAsyncClient ddb, String table) {
-    this(() -> ddb, table, failure -> {});
+    this(() -> ddb, table, (client, failure) -> {});
```
mrouvroy-floe: [nit] **Fixed-client constructor enables a pointless closed-pool retry**

The convenience constructor delegates with a non-null no-op handler, so `supportsClientRefresh` becomes `true` and a closed-pool failure triggers a retry against the exact same fixed client — a guaranteed second failure with no refresh in between.

Passing `null` instead lets the main constructor's null-handling set `supportsClientRefresh = false` (with the same no-op handler), which matches the intended semantics of the flag.

```suggestion
    this(() -> ddb, table, null);
```

---

🤖 _Generated with Fable 5 and GPT-5.5._

storage/aws/src/main/java/ai/floedb/floecat/storage/aws/secrets/ProdSecretsManager.java
```diff
@@ -163,41 +194,109 @@ private static List<Tag> buildTags(String accountId, String secretName) {
...
+      }
+      closeQuietly(failedClient);
+      StsClient previousSts = stsClient.get();
+      if (previousSts == null) {
+        return;
+      }
+      StsClient nextSts = awsClients.stsClient();
+      if (stsClient.compareAndSet(previousSts, nextSts)) {
+        closeQuietly(previousSts);
+        closePerAccountClients();
```
mrouvroy-floe: [minor] **Any single per-account closed pool tears down all clients, failing concurrent in-flight retries**

In the role-ARN path, a closed-pool failure on one account's Secrets Manager client always swaps the shared `StsClient` and calls `closePerAccountClients()`, even when the failure came from that client's own Secrets Manager pool and STS is healthy. Every other account's client is closed while it may be serving a request on another thread.

Those threads do recover via `withClientRefresh`'s single retry, so this self-heals in the common case. But a request that is already on its retry attempt when the teardown lands fails permanently with `Connection pool shut down` — the retry budget is one, and the teardown consumes it. There's also churn cost: N healthy clients plus their assume-role sessions are discarded, and all role-path operations stall on `perAccountClientLock` while the closes run.

If the STS swap is intentional because a closed-pool failure can't be attributed to the secrets pool vs. the STS pool underneath `StsAssumeRoleCredentialsProvider`, consider attributing it by inspecting the cause chain (an STS-originated failure surfaces through credential resolution, e.g. wrapped in `SdkClientException` from credential loading) and only swapping STS + clearing the cache in that case. At minimum a comment documenting the deliberate blast radius would help the next reader.

---

🤖 _Generated with Fable 5 and GPT-5.5._

storage/aws/src/main/java/ai/floedb/floecat/storage/aws/DynamoDbClientManager.java
```diff
@@ -0,0 +1,81 @@
...
+@ApplicationScoped
+public class DynamoDbClientManager {
+
+  private static final Logger LOG = Logger.getLogger(DynamoDbClientManager.class);
+
+  @Inject AwsClients awsClients;
+
+  private final AtomicReference<DynamoDbClient> current = new AtomicReference<>();
+
+  public DynamoDbClient current() {
```
mrouvroy-floe, line 37: [minor] **`current()` resurrects a client after `@PreDestroy` close, leaking it**

`close()` sets the `AtomicReference` to `null`, but `current()` treats `null` as "not yet created" and happily builds a fresh client via `awsClients.newDynamoDbClient()`. A reconcile operation racing container shutdown will recreate a client after `@PreDestroy` ran, and nothing ever closes it.

`ProdSecretsManager` handles the same situation explicitly with `requireOpen(...)` throwing `IllegalStateException("secrets manager is shut down")`. Doing the equivalent here (a `closed` flag checked in `current()`, or a sentinel) would keep the two shutdown behaviors consistent and avoid the leak.

The same pattern exists in the pre-existing `DynamoDbAsyncClientManager` and `S3ClientManager`, so this could also be a small follow-up covering all three.

---

🤖 _Generated with Fable 5 and GPT-5.5._

storage/aws/src/main/java/ai/floedb/floecat/storage/aws/DynamoDbAsyncClientManager.java
```diff
@@ -47,15 +47,17 @@ public DynamoDbAsyncClient current() {
     return current.get();
   }
 
-  public void refreshAfterFailure(Throwable failure) {
+  public void refreshAfterFailure(DynamoDbAsyncClient failedClient, Throwable failure) {
```
mrouvroy-floe, line 61: [minor] **`S3ClientManager.refreshAfterFailure` still has the stale-identity bug fixed here**

This PR adds a `failedClient` identity guard to both DynamoDB managers so a stale closed-pool failure can't evict and close a newer healthy client. `S3ClientManager.refreshAfterFailure(Throwable)` (untouched by this PR) still uses the old shape: it reads `current.get()` as `previous` and swaps it unconditionally, so a delayed failure report from an already-replaced client closes the healthy replacement out from under in-flight S3 operations.

Since `S3BlobStore` wires it exactly the way `DynamoDbKvStore` wires the async manager, the same production scenario applies. Not blocking here because it's pre-existing code outside this PR's stated scope, but it's the identical bug class this PR exists to fix — worth a fast-follow that aligns `S3ClientManager.refreshAfterFailure` to the `(failedClient, failure)` signature.

---

🤖 _Generated with Fable 5 and GPT-5.5._

mrouvroy-floe (approved):

PR #366, opened 2026-07-06
fix: enforce system-object immutability and cross-kind relation-name uniqueness

mrouvroy-floe:
## Summary

Makes relation naming and system-object protection structural rather than best-effort. Replaces the query-time ambiguity check with a write-time guarantee.

## What's in here

- Guard indirect system-table writes — snapshot/index/statistics/transaction-apply paths now reject mutations targeting system objects, closing the gaps the direct-RPC guards didn't cover.
- Remove query-time catalog ambiguity checks — resolution no longer detects table/view name collisions after the fact.
- Shared relation-name claim pointer — tables and views reserve one kind-agnostic /relations/by-name/<name> pointer, so a table and a view can never hold the same name in a namespace; whichever is created (or renamed) second loses the atomic reservation.
- Drop the redundant requireRelationNameWriteEligible pre-check — the atomic claim plus the existing NameConflictException → ALREADY_EXISTS mapping now cover it on all create/rename paths.
- Resolver cleanup — dedupe catalog/namespace resolution, remove the dead view-relation resolver.
- Non-idempotent CreateView maps cross-kind conflicts to VIEW_ALREADY_EXISTS.
- System views listed via the Catalog Surface report the requested (user-facing) catalog id, matching tables.
- Transaction table write-eligibility maps transient failures to retryable instead of a terminal conflict.

## Type of Change

- [x] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist


service/src/main/java/ai/floedb/floecat/service/metagraph/resolver/NameResolver.java
```diff
@@ -167,75 +167,91 @@ public Optional<ResourceId> resolveViewId(String cid, String accountId, NameRef
...
 
-  public Optional<ResolvedRelation> resolveTableRelation(String accountId, NameRef ref) {
+  /**
+   * Kind-agnostic name resolution: resolves catalog and namespace once, then probes tables and (on
+   * miss) views. Unlike calling {@link #resolveTableRelation} followed by {@link
+   * #resolveViewRelation}, this does not re-resolve the catalog and namespace for the view probe. A
+   * relation name is unique across kinds (enforced by the shared relation-name claim), so the
+   * table-first order is deterministic.
+   */
+  public Optional<ResourceId> resolveRelationId(String accountId, NameRef ref) {
```
jchauzi-floe, line 176: **Ambiguity check removal has no migration path for existing collisions**

Pre-PR, `TABLE` and `VIEW` had independent per-kind name pointers (`tablePointerByName` / `viewPointerByName`) — nothing prevented a table and a view from already sharing a name in production data. The removed `QUERY_INPUT_AMBIGUOUS` check was the only thing that ever caught that.

The new `/relations/by-name/<name>` claim only stops *new* collisions. There's no backfill/reconciliation for anything that collided before this deploys — `resolveRelationId` now silently returns the table and the view becomes unreachable by name, with no error and no signal.

Also worth noting: `UserGraphTest.resolveNameReturnsTableWhenDuplicateRelationNamesExist` seeds a collision directly and asserts "returns the table" as the expected behavior, rather than treating it as a case to detect.

If pre-existing collisions are known not to exist in prod, fine as-is. If not, worth a one-time scan before/after rollout.

service/src/main/java/ai/floedb/floecat/service/transaction/impl/TransactionIntentApplierSupport.java
```diff
@@ -701,6 +825,53 @@ private ApplyOutcome buildNameUpsertOp(
...
+      List<PointerStore.CasOp> ops) {
+    var ptr = pointerStore.get(key).orElse(null);
+    if (ptr == null) {
+      Pointer created = PointerReferences.blobPointer(key, nextBlobUri, 1L, owner, "");
+      return addOp(new PointerStore.CasUpsert(key, 0L, created), key, touchedKeys, ops);
+    }
+    ResourceId held = ptr.getResourceId();
+    if (!Objects.equals(held.getId(), owner.getId())) {
+      return ApplyOutcome.conflict(
+          "RELATION_NAME_CONFLICT",
```
jchauzi-floe, line 848: **Same collision surfaces as 3 different error codes depending on the path**

- Direct RPC (`ViewServiceImpl.createView`, ~line 229): `VIEW_ALREADY_EXISTS`
- Transaction-apply (here): `RELATION_NAME_CONFLICT`
- Same-kind rename conflict, same file (`buildNameUpsertOp`): `NAME_POINTER_CONFLICT`

Same underlying condition (name already claimed by the other kind), three different codes depending on which of the two independent claim implementations detected it. Since this file hand-rolls the CAS ops instead of going through `Schemas.java` / `GenericResourceRepository` (the commit message says it "bypasses the generic repo"), the two implementations can drift beyond just the error code too.

service/src/main/java/ai/floedb/floecat/service/catalog/impl/ViewServiceImpl.java
```diff
@@ -224,7 +224,19 @@ public Uni<CreateViewResponse> createView(CreateViewRequest request) {
                               "catalog_id", spec.getCatalogId().getId(),
                               "namespace_id", spec.getNamespaceId().getId()));
                     }
-                    viewRepo.create(view);
+                    try {
+                      viewRepo.create(view);
+                    } catch (BaseResourceRepository.NameConflictException nce) {
```
jchauzi-floe, line 230: See the counterpart check in `TransactionIntentApplierSupport.buildRelationClaimUpsertOp` — same cross-kind collision, but that path throws `RELATION_NAME_CONFLICT` instead of `VIEW_ALREADY_EXISTS`. Same invariant, two error codes depending on which write path catches it.

jchauzi-floe (commented):
Automated review pass on this diff. Inline comments below, most important first:

1. Ambiguity-check removal has no migration path for pre-existing collisions (`NameResolver.java`)
2. Same cross-kind conflict surfaces as 3 different error codes across 2 independent claim implementations
3. `GetView` wasn't updated to match the `ListViews` catalog-id fix
4. `CatalogSurfaceWritePolicy` enforcement is opt-in at 14+ call sites, not structural
5. Duplicated `systemNamespacePathMatch` logic instead of reusing the existing copy
6. Transaction-apply write-eligibility fails open if `overlay` is null
7. `INTENDED_BLOB_URI` skips catalog/namespace prepare-time validation that `TABLE` payloads get
8. `validateTableWriteEligibility` / `validateExistingTableWriteEligibility` are near-duplicate methods

service/src/main/java/ai/floedb/floecat/service/catalog/impl/surface/CatalogSurfaceViewPageSource.java
```diff
@@ -91,6 +91,9 @@ public String systemRelativeKey(ViewNode node) {
 
   @Override
   public View mapSystemNode(ViewNode node) {
-    return CatalogSurfaceViews.viewFromSystemNode(node);
+    // Rewrite to the requested (user-facing) catalog so system views listed through the Catalog
+    // Surface report the same catalog id as sibling tables (the "symlink" model), rather than the
+    // raw system node's catalog id.
+    return CatalogSurfaceViews.viewFromSystemNode(node).toBuilder().setCatalogId(catalogId).build();
```
jchauzi-floe, line 97: **`GetView` doesn't get the same fix**

This rewrites the catalog id for `ListViews`, matching tables. `CatalogSurfaceViews.getView()` (`viewFromOverlayNodeOrRepo` → `viewFromSystemNode`) wasn't touched by this PR and still returns the raw system catalog id for the same view.

Concretely: list a system view under catalog `cat` → `catalog_id: "cat"`. `GetView` the same `view_id` → `catalog_id: "<raw system catalog>"`. Same resource, two different catalog ids depending on which RPC you call.

service/src/main/java/ai/floedb/floecat/service/catalog/impl/NamespaceServiceImpl.java
```diff
@@ -400,6 +400,10 @@ public Uni<UpdateNamespaceResponse> updateNamespace(UpdateNamespaceRequest reque
 
                   var mask = normalizeMask(request.getUpdateMask());
                   var desired = applyNamespaceSpecPatch(current, request.getSpec(), mask, corr);
+                  var desiredPath = new ArrayList<>(desired.getParentsList());
+                  desiredPath.add(desired.getDisplayName());
+                  catalogSurfaceWritePolicy()
+                      .requireNamespacePathWriteEligible(desired.getCatalogId(), desiredPath, corr);
```
jchauzi-floe, line 406: **`CatalogSurfaceWritePolicy` is opt-in, not enforced**

This 4-line addition is itself evidence of the problem: `createNamespace` already called `requireNamespacePathWriteEligible`, but `updateNamespace` didn't until this patch added it. `CatalogSurfaceWritePolicy` is a plain class instantiated ad hoc at 14+ call sites across the codebase — nothing (base class, interceptor, repo-layer check) forces a writer to call it. This PR fixes the one gap that's been found; nothing stops the next one from being just as easy to miss.

jchauzi-floe, line 406: **Duplicates `systemNamespacePathMatch` instead of reusing it**

`requireNamespacePathWriteEligible` in `CatalogSurfaceWritePolicy` re-implements this class's own `systemNamespacePathMatch` / `listSystemNamespaces` (same `PATH_DELIM`, same `SystemPathMatch` enum, same matching logic) — `createNamespace` still calls the local copy, `updateNamespace` (this diff) calls the policy-class copy. Two copies of the same system-path-collision rule that now have to be kept in sync by hand.

service/src/main/java/ai/floedb/floecat/service/transaction/impl/TransactionIntentApplierSupport.java
```diff
@@ -608,6 +659,79 @@ private ApplyOutcome planConnectorIntentOps(
...
 
+  private ApplyOutcome validateTableWriteEligibility(Table table, boolean existingTable) {
+    if (table == null || !table.hasResourceId()) {
+      return ApplyOutcome.conflict(
+          "TABLE_INTENT_INVALID_PAYLOAD", "table payload is missing resource_id", null, null, null);
+    }
+    if (SystemResourceIdGenerator.isSystemId(table.getResourceId())) {
+      return tableImmutableConflict(table.getResourceId().getId());
+    }
+    if (overlay == null) {
```
jchauzi-floe, line 683: **Fails open if `overlay` is ever null**

```java
if (overlay == null) {
  return ApplyOutcome.applied();
}
```

Treats the write as allowed rather than rejecting/retrying — skips exactly the catalog/namespace/table-writability checks this PR adds (the `SystemResourceIdGenerator.isSystemId` check above it still runs independently). Same pattern repeated at line 707 in `validateExistingTableWriteEligibility`.

`overlay` is `@Inject`-ed on an `@ApplicationScoped` bean so CDI should guarantee non-null in production — but a guard whose entire purpose is closing gaps probably shouldn't default to "allow" on its own null case.

service/src/main/java/ai/floedb/floecat/service/transaction/impl/TransactionsServiceImpl.java
```diff
@@ -1549,6 +1559,102 @@ && isTableByIdPointer(pointerKey)) {
...
+        String blobUri = change.getIntendedBlobUri().trim();
+        if (isDeleteSentinelBlobUri(accountId, txId, target.pointerKey(), blobUri)) {
+          catalogSurfaceWritePolicy()
+              .requireWritableTableForDelete(
+                  tableId,
+                  correlationId(),
+                  currentVersion > 0L || hasMeaningfulPrecondition(change.getPrecondition()));
+          return;
+        }
+        if (currentVersion > 0L) {
```
jchauzi-floe, line 1594: **`INTENDED_BLOB_URI` skips catalog/namespace validation at prepare time; `TABLE` doesn't**

For `case INTENDED_BLOB_URI` with `currentVersion == 0L` (non-delete), only `rejectSystemTableId(tableId)` runs above this — no catalog/namespace check. `case TABLE` always calls `requireWritableTablePayload`, which checks both regardless of `currentVersion`.

Likely unavoidable here (blob content isn't readable until apply), and `TransactionIntentApplierSupport.validateTableWriteEligibility` does catch it at apply time — so it's late rejection, not a bypass. But `PrepareTransaction` can succeed for a create that's going to fail at commit, only for this one payload shape. Worth a comment explaining why, if intentional.

service/src/main/java/ai/floedb/floecat/service/transaction/impl/TransactionIntentApplierSupport.java
```diff
@@ -608,6 +659,79 @@ private ApplyOutcome planConnectorIntentOps(
     return ApplyOutcome.applied();
   }
 
+  private ApplyOutcome validateTableWriteEligibility(Table table, boolean existingTable) {
```
jchauzi-floe: **`validateTableWriteEligibility` / `validateExistingTableWriteEligibility` are ~90% identical**

Same null/resource-id check, same `isSystemId` check, same `overlay == null` short-circuit, same `catch (StatusRuntimeException)` → conflict / `catch (RuntimeException)` → retryable. Only real difference: this one also validates catalog/namespace (needed since update can move a table; delete can't).

Collapses into one method with a flag, e.g.:

```java
private ApplyOutcome validateTableWriteEligibility(Table table, boolean validateScope) { ... }
```

service/src/main/java/ai/floedb/floecat/service/catalog/impl/NamespaceServiceImpl.java
mrouvroy-floe: NamespaceServiceImpl` has no local `systemNamespacePathMatch`/`SystemPathMatch` — both `createNamespace` (line 200) and `updateNamespace` (line 406) call `writePolicy.requireNamespacePathWriteEligible`. The extraction that removed the local copy is commit `9e92b56d4 "extract namespace catalog surface policy"`.
 So there's one implementation, both call sites use it.

service/src/main/java/ai/floedb/floecat/service/metagraph/resolver/NameResolver.java
mrouvroy-floe: No collisions on existing systems so no backward migration checks required right now

jchauzi-floe (approved):

PR #367, opened 2026-07-06
fix: correct Catalog Surface pagination (precise cursors, phased system/user listing)

mrouvroy-floe:
## Summary

Fixes several ways paginated listings could silently drop rows at page boundaries, and adds the store primitive that makes precise resumption possible.

## What's in here

- Namespace pager now runs the repo phase for continuation tokens and scopes page tokens to their phase (user vs system), then resumes precisely after the last emitted row rather than after the over-fetched batch — via a new PointerStore.pageTokenAfterKey / KvStore.pageTokenAfterKey SPI implemented for the in-memory and DynamoDB stores.
- Prefix listing (listTablesByPrefix / listViewsByPrefix) is now phase-scoped so system rows can neither hide nor overflow past user rows; reported totals combine system + user.
- In-memory stores resume a prefix scan positionally when the cursor row was deleted between pages, matching DynamoDB's exclusiveStartKey semantics.
- Guards the namespace pager against non-positive page sizes.
- Adds real-store coverage for pageTokenAfterKey (contract tests + a Dynamo token-format pin + an end-to-end 150-namespace drain) and mixed-page / views-path / malformed-token prefix tests.

## Type of Change

- [x] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/catalog/impl/surface/CatalogSurfaceNamespacePager.java
+++ b/service/src/main/java/ai/floedb/floecat/service/catalog/impl/surface/CatalogSurfaceNamespacePager.java
@@ -101,11 +101,16 @@ final class CatalogSurfaceNamespacePager {
             .sorted(Comparator.comparing(SysItem::rel))
             .toList();
 
+    // A full page is not the same as having another row to return: only advertise a continuation
+    // when the loop actually stopped before an un-emitted system namespace. Otherwise a page that
+    // exactly consumes the last system row would hand out a token whose next page is empty.
+    boolean hasMoreSystem = false;
     for (var it : sysItems) {
       if (!resumeSystemRel.isBlank() && it.rel().compareTo(resumeSystemRel) <= 0) {
         continue;
       }
       if (out.size() >= want) {
+        hasMoreSystem = true;
         break;
       }
       out.add(source.mapSystemNode(it.namespace()));
@@ -113,7 +118,7 @@ final class CatalogSurfaceNamespacePager {
     }
 
     String nextToken =
-        out.size() == want && !lastSystemRel.isBlank()
+        hasMoreSystem && !lastSystemRel.isBlank()
             ? CatalogSurfaceSupport.encodeToken(NS_TOKEN_PREFIX, lastSystemRel)
             : "";
 
--- a/service/src/main/java/ai/floedb/floecat/service/catalog/impl/surface/CatalogSurfaceNamespacePager.java
+++ b/service/src/main/java/ai/floedb/floecat/service/catalog/impl/surface/CatalogSurfaceNamespacePager.java
@@ -27,60 +27,76 @@ import java.util.Map;
 
 final class CatalogSurfaceNamespacePager {
 
+  // System-phase resume token: the carried relative name is a *system* namespace name.
   private static final String NS_TOKEN_PREFIX = "ns:";
+  // Repo-phase resume token: the carried relative name is a *user* namespace name.
+  private static final String NSU_TOKEN_PREFIX = "nsu:";
 
   private CatalogSurfaceNamespacePager() {}
 
   static Page list(int want, String pageToken, Source source, String corr) {
 
     final int batch = Math.max(want * 4, 64);
-    final boolean isServiceToken = pageToken != null && pageToken.startsWith(NS_TOKEN_PREFIX);
-    final String resumeAfterRel =
-        isServiceToken ? CatalogSurfaceSupport.decodeToken(NS_TOKEN_PREFIX, pageToken, corr) : "";
-    String cursor = isServiceToken ? "" : pageToken;
+    final boolean isSystemToken = pageToken != null && pageToken.startsWith(NS_TOKEN_PREFIX);
+    final boolean isUserToken = pageToken != null && pageToken.startsWith(NSU_TOKEN_PREFIX);
+    // Resume keys are scoped to their own phase: a user relative name must never filter system
+    // namespaces (and vice versa), since the two are independent, independently-ordered keyspaces.
+    final String resumeSystemRel =
+        isSystemToken ? CatalogSurfaceSupport.decodeToken(NS_TOKEN_PREFIX, pageToken, corr) : "";
+    final String resumeUserRel =
+        isUserToken ? CatalogSurfaceSupport.decodeToken(NSU_TOKEN_PREFIX, pageToken, corr) : "";
+    // Raw repo cursors carry no service prefix; service tokens restart the scan and resume by name.
+    String cursor = (isSystemToken || isUserToken) ? "" : pageToken;
 
     var out = new ArrayList<Namespace>(want);
-    String lastEmittedRel = "";
+    String lastUserRel = "";
+    String lastSystemRel = "";
 
-    // The repo phase must run for continuation tokens too: a batch that exhausts the cursor can
-    // hold
-    // more matches than fit on one page, so later pages re-scan from the start and skip already-
-    // emitted namespaces by name via resumeAfterRel. Skipping the scan for service tokens would
-    // drop
-    // every remaining user namespace once a page-filling batch exhausted the cursor.
-    while (out.size() < want) {
-      var next = new StringBuilder();
-      final List<Namespace> scanned;
-      try {
-        scanned = source.listRepo(batch, cursor, next);
-      } catch (IllegalArgumentException badToken) {
-        throw GrpcErrors.invalidArgument(corr, PAGE_TOKEN_INVALID, Map.of("page_token", cursor));
-      }
-
-      for (var ns : scanned) {
-        if (!matches(ns, source)) {
-          continue;
+    // Repo (user) phase. Runs for the first page, raw-cursor pages, and user-phase resume tokens;
+    // it
+    // is skipped only once we have advanced into the system phase (system-phase token). A batch
+    // that
+    // exhausts the cursor can hold more matches than fit on one page, so a user-phase resume token
+    // re-scans from the start and skips already-emitted rows by name via resumeUserRel.
+    if (!isSystemToken) {
+      while (out.size() < want) {
+        var next = new StringBuilder();
+        final List<Namespace> scanned;
+        try {
+          scanned = source.listRepo(batch, cursor, next);
+        } catch (IllegalArgumentException badToken) {
+          throw GrpcErrors.invalidArgument(corr, PAGE_TOKEN_INVALID, Map.of("page_token", cursor));
         }
 
-        var rel = relativeQualifiedName(ns, source.parentPath());
-        if (!resumeAfterRel.isBlank() && rel.compareTo(resumeAfterRel) <= 0) {
-          continue;
+        for (var ns : scanned) {
+          if (!matches(ns, source)) {
+            continue;
+          }
+
+          var rel = relativeQualifiedName(ns, source.parentPath());
+          if (!resumeUserRel.isBlank() && rel.compareTo(resumeUserRel) <= 0) {
+            continue;
+          }
+
+          out.add(ns);
+          lastUserRel = rel;
+          if (out.size() >= want) {
+            break;
+          }
         }
 
-        out.add(ns);
-        lastEmittedRel = rel;
-        if (out.size() >= want) {
+        cursor = next.toString();
+        if (cursor.isBlank() || out.size() >= want) {
           break;
         }
       }
-
-      cursor = next.toString();
-      if (cursor.isBlank() || out.size() >= want) {
-        break;
-      }
     }
 
-    if (cursor.isBlank() && out.size() < want) {
+    // System phase: only once the repo cursor is exhausted, or when resuming directly into it. When
+    // transitioning here from the repo phase, resumeSystemRel is blank, so system namespaces are
+    // emitted from the beginning regardless of where the user phase left off.
+    boolean enteredSystemPhase = false;
+    if ((isSystemToken || cursor.isBlank()) && out.size() < want) {
       record SysItem(NamespaceNode namespace, String rel) {}
 
       var sysItems =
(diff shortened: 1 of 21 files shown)


service/src/main/java/ai/floedb/floecat/service/metagraph/overlay/MetaGraph.java
```diff
@@ -465,19 +456,89 @@ public ResolveResult batchResolveViews(
...
+    return new ResolveResult(
+        merged, systemAll.size() + user.totalSize(), encodeUserToken(user.nextToken()));
+  }
 
-    return mergePrefixResults(system, user, max);
+  /**
+   * User-relation total for combined counts on system-phase pages. The one-row probe's rows and
+   * cursor are discarded; the user phase later starts from a blank cursor, so nothing is skipped.
+   */
+  private int userTotalByPrefix(String correlationId, NameRef prefix, boolean tables) {
```
jchauzi-floe, line 550: **CONFIRMED** — this does a real backend round trip just to get a count.

`userGraph.resolveTables`/`resolveViews(prefix, 1, "")` runs a full `catalogRepository.getByName` + `namespaceRepository.getByPath` + `tableRepository.list(limit=1)` + `tableRepository.count(...)`, and the fetched row/cursor are thrown away. It's called once per system-phase page (see the call sites a few lines up), so for N system relations at page size P this fires ~N/P times.

`NamespaceRepository.count` / `CatalogSurfaceRelationPager.countRepo()` already wrap a plain `repo.countByPrefix(...)` with no row fetch — use that instead.

service/src/main/java/ai/floedb/floecat/service/metagraph/overlay/MetaGraph.java
```diff
@@ -465,19 +456,89 @@ public ResolveResult batchResolveViews(
...
-        firstPage ? collectSystemRelationsInNamespace(prefix, ctx, false, max) : List.of();
-    int userLimit = Math.max(1, max - system.size());
+    final boolean sysToken = token != null && token.startsWith(SYS_TOKEN_PREFIX);
+    final String sysResume = sysToken ? decodeSysToken(correlationId, token) : "";
+    final boolean userPhase = !sysToken && token != null && !token.isBlank();
+    final String userToken = userPhase ? decodeUserToken(token) : "";
+
+    // Collected on every page (bounded, in-memory) so totals can include the system rows.
+    List<CatalogOverlay.QualifiedRelation> systemAll =
+        new ArrayList<>(collectSystemRelationsInNamespace(prefix, ctx, tables, Integer.MAX_VALUE));
```
jchauzi-floe: **CONFIRMED** — `collectSystemRelationsInNamespace(..., Integer.MAX_VALUE)` + `sort` runs on *every* page, even pages fully in the user phase (`userPhase == true`) where the result is only used for `.size()` further down.

Bounded to in-memory `SystemGraph` lookups so no I/O, but it's still real work (per-relation `tableName`/`viewName` resolution + `NameRef` construction) repeated across the whole pagination sequence instead of once. Worth gating behind `!userPhase`, or caching the count separately from the sorted list.

storage/memory/src/main/java/ai/floedb/floecat/storage/memory/InMemoryKvStore.java
```diff
@@ -91,10 +91,21 @@ public Uni<Page> queryByPartitionKeyPrefix(
           break;
         }
       }
-      if (index < 0) {
+      if (index >= 0) {
+        start = index + 1;
+      } else if (token.startsWith(skPrefix)) {
```
jchauzi-floe: **PLAUSIBLE** — not exploitable through any current call site, but the malformed-token check is vacuous when `sortKeyPrefix` is blank.

`token.startsWith(skPrefix)` is always `true` when `skPrefix` is `""`, so a full scan (no sort-key prefix) will silently accept *any* garbage token as a valid resume position instead of throwing `IllegalArgumentException`. Every current caller happens to use a non-blank prefix (routed through `Keys.*Prefix()`), so this doesn't fire today — but nothing enforces that, and it's untested against a blank prefix.

storage/memory/src/main/java/ai/floedb/floecat/storage/memory/InMemoryKvStore.java
```diff
@@ -110,6 +121,12 @@ public Uni<Page> queryByPartitionKeyPrefix(
     return Uni.createFrom().item(new Page(List.copyOf(items), nextToken));
   }
 
+  @Override
+  public String pageTokenAfterKey(Key key) {
```
jchauzi-floe, line 128: **PLAUSIBLE**, currently dead code — `encodeToken("")` returns `""`, which is indistinguishable from the "no more pages" sentinel used everywhere else in this codebase. If a row's sort key is ever empty, this would silently truncate pagination.

Traced every caller: nothing reaches this in production today — the memory-profile `PointerStore` bypasses `KvStore.pageTokenAfterKey` entirely (`InMemoryPointerStore.pageTokenAfterKey` just returns the raw key), and the DynamoDB path encodes `pk + "\n" + sk` together so it can't degenerate to `""`. Flagging because it's a shipped SPI method with a real gap, not because it fires anywhere yet.

storage/memory/src/main/java/ai/floedb/floecat/storage/memory/InMemoryPointerStore.java
```diff
@@ -77,10 +77,16 @@ public List<Pointer> listPointersByPrefix(
     int start = 0;
     if (pageToken != null && !pageToken.isEmpty()) {
       int idx = Collections.binarySearch(keys, pageToken);
-      if (idx < 0) {
+      if (idx >= 0) {
+        start = idx + 1;
+      } else if (pageToken.startsWith(pfx)) {
```
jchauzi-floe: **PLAUSIBLE** — same issue as `InMemoryKvStore.queryByPartitionKeyPrefix`: `pageToken.startsWith(pfx)` is vacuously `true` when `pfx` is `""`, so a blank-prefix scan can no longer reject a malformed token — it just resumes from wherever the binary search inserts it. Not reachable from any current caller (all go through non-blank `Keys.*Prefix()` values), but untested and silently wrong if that ever changes.

service/src/main/java/ai/floedb/floecat/service/metagraph/overlay/MetaGraph.java
```diff
@@ -679,6 +742,25 @@ private static String decodeUserToken(String token) {
     return token;
   }
 
+  private static String encodeSysName(String name) {
```
jchauzi-floe: **CONFIRMED** — this duplicates `CatalogSurfaceSupport.encodeToken`/`decodeToken`, which already implements "prefix + base64url(name), throw `PAGE_TOKEN_INVALID` on bad input" and is shared by `CatalogSurfaceNamespacePager` and `CatalogSurfaceRelationPager`.

Also, within this file the two phases don't encode consistently: `sys:` tokens go through `encodeSysName` (base64), but `u:` tokens (`encodeUserToken`, line 764) just concatenate the raw delegate cursor with no encoding. Not a bug by itself since the two payloads are different kinds of string, but there are now two independent "phase token" implementations in the codebase that can drift apart.

service/src/main/java/ai/floedb/floecat/service/metagraph/overlay/MetaGraph.java
```diff
@@ -465,19 +456,89 @@ public ResolveResult batchResolveViews(
...
+  /**
+   * Phased prefix listing: system relations first (sorted by canonical name), then user relations.
+   *
+   * <p>Tokens are phase-scoped so neither cursor can advance past rows the page did not emit. A
+   * {@code sys:} token resumes the (bounded, in-memory) system list after the carried name; the
+   * user-phase token is the user graph's own cursor. The user graph is only consulted for rows once
+   * the system rows are fully emitted, so a page filled by system rows never advances — and can
+   * never drop — user results. Reported totals combine both sources.
+   */
+  private ResolveResult listRelationsByPrefix(
```
jchauzi-floe, line 476: **PLAUSIBLE** (architecture cleanup, not a bug) — `listRelationsByPrefix` reimplements the same "system phase, then user phase, phase-scoped resume token" shape that `CatalogSurfaceRelationPager` already generalizes via its `Source<P,N>` abstraction.

Checked whether the un-fixed sibling pager has the same row-skipping bug this PR fixes elsewhere — it doesn't: `CatalogSurfaceRelationPager` never post-filters its scanned batch, so its raw batch-end cursor is safe. But the duplication itself is real: a future fix to one phased pager (like this PR) won't automatically apply to the other.

storage/aws/src/main/java/ai/floedb/floecat/storage/kv/dynamodb/DynamoDbKvStore.java
```diff
@@ -153,6 +153,15 @@ static Optional<Map<String, AttributeValue>> decodeToken(Optional<String> token)
     return Optional.of(Map.of(ATTR_PARTITION_KEY, S(parts[0]), ATTR_SORT_KEY, S(parts[1])));
   }
 
+  @Override
+  public String pageTokenAfterKey(Key key) {
```
jchauzi-floe, line 164: **CONFIRMED** — this rebuilds the `(pk, sk)` map inline instead of calling the existing `keyMap(Key)` helper (line 73), which does the exact same thing and is used by every other method in this class.

```java
return encodeToken(keyMap(key)).orElseThrow();
```

would do it and keep this in sync if `keyMap` ever changes.

jchauzi-floe (commented):
Reviewed the pagination-correctness fix. Core logic (`CatalogSurfaceNamespacePager`, the `pageTokenAfterKey` SPI, `MetaGraph.listRelationsByPrefix`) is correct for the bug class described here — traced the fill/exhaust/phase-boundary cases by hand and found no flaw, and it's backed by solid tests (150-namespace real-store drain, mixed-page, malformed-token).

All findings below are in supporting code, not the headline fix. Two labels:
- **CONFIRMED** — verified against this code, reproducible today.
- **PLAUSIBLE** — mechanism is real but not reachable through any current call site (latent, untested, or dead code). Worth fixing, not urgent.

One more, not inline since `RepoTestPointerStores.java` isn't touched by this PR: **PLAUSIBLE** — `service/src/test/java/ai/floedb/floecat/service/repo/impl/RepoTestPointerStores.java:32`, `DelegatingPointerStore` is documented as forwarding every `PointerStore` call to its delegate, but it doesn't override the new `pageTokenAfterKey`. Any subclass (`FailingBatchPointerStore`, `CountingPrefixScanPointerStore`, etc.) will throw `UnsupportedOperationException` if ever wrapped around a store that's actually capable of it. No current test hits this, so nothing fails today — but it's a trap for the next person adding a pagination test through one of these decorators.

service/src/main/java/ai/floedb/floecat/service/catalog/impl/surface/CatalogSurfaceNamespacePager.java
```diff
@@ -74,35 +81,42 @@ static Page list(int want, String pageToken, Source source, String corr) {
...
+    record SysItem(NamespaceNode namespace, String rel) {}
+
+    var sysItems =
+        source.systemNamespaces().stream()
+            .filter(ns -> matches(ns, source))
+            .map(ns -> new SysItem(ns, relativeQualifiedName(ns, source.parentPath())))
+            .sorted(Comparator.comparing(SysItem::rel))
+            .toList();
+
+    for (var it : sysItems) {
```
mcusack-floe, line 108: The current code emits an ns: continuation whenever the page fills on a system namespace, but a full page is not the same as having another row to return. If the last available system namespace exactly fills the page, the token points after that final row, so the next request produces an empty page. Tracking whether the loop actually stopped before an un-emitted system row preserves the intended continuation behavior and avoids advertising a next page that does not exist.

`boolean hasMoreSystem = false;
for (var it : sysItems) {
  if (!resumeSystemRel.isBlank() && it.rel().compareTo(resumeSystemRel) <= 0) {
    continue;
  }
  if (out.size() >= want) {
    hasMoreSystem = true;
    break;
  }
  out.add(source.mapSystemNode(it.namespace()));
  lastSystemRel = it.rel();
}

String nextToken =
    hasMoreSystem && !lastSystemRel.isBlank()
        ? CatalogSurfaceSupport.encodeToken(NS_TOKEN_PREFIX, lastSystemRel)
        : "";`

service/src/main/java/ai/floedb/floecat/service/metagraph/overlay/MetaGraph.java
```diff
@@ -465,19 +455,102 @@ public ResolveResult batchResolveViews(
...
+          // System rows remain: continue the system phase next page; the user cursor is untouched.
+          return new ResolveResult(
+              merged,
+              systemTotal + userTotalByPrefix(correlationId, prefix, tables),
+              PageTokens.encode(SYS_TOKEN_PREFIX, lastSystemName));
+        }
+        merged.add(rel);
+        lastSystemName = name;
+      }
+      if (merged.size() >= max) {
```
mcusack-floe, line 520: A boundary bug appears when system rows exactly fill the page and there are no user rows. The code unconditionally returns USER_TOKEN_PREFIX after the system phase fills exactly, assuming there may be user results. If userTotalByPrefix(...) is 0, that token only leads to an empty follow-up page. The handoff token should be emitted only when there are user rows to continue into.

`if (merged.size() >= max) {
  int userTotal = userTotalByPrefix(correlationId, prefix, tables);
  return new ResolveResult(
      merged,
      systemTotal + userTotal,
      userTotal > 0 ? USER_TOKEN_PREFIX : "");
}`

jchauzi-floe (approved):

PR #368, opened 2026-07-06
perf: cut redundant storage reads on the query resolution path

mrouvroy-floe:
## Summary

Removes repeated pointer/blob reads on the cold query path (name → id → pin → hydrate); most of these paths had no cache at all, so the wins apply warm too.
Effect: on a typical N-table-by-name query, roughly a 60–65% reduction in storage ops on the resolve+pin segment (view resolution benefits most).

## What's in here

- snapshotPinFor answers the "is this a system table?" check from the in-memory system graph instead of hydrating the full user node from storage.
- Metadata/hydration read trimming: metaFor* no longer re-reads the pointer it just fetched; graph consumers use a pointer-only meta (no S3 HEAD, no etag); node hydration fetches the blob the meta already names instead of re-reading the pointer via getById. Cold hydration drops from 3 pointer + HEAD + GET to 1 pointer + GET.
- Batch scope memoization: fully-qualified list resolution and query-input name resolution resolve each catalog/namespace once per call instead of once per name (new CatalogOverlay.resolveNames, default-safe for other overlays).
- Kind-agnostic name resolution answers from the shared relation-name claim (1 pointer, no blob) instead of a table-then-view probe.
- Folds the batchResolveTables/batchResolveViews twins into one shared method.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/metagraph/loader/NodeLoader.java
+++ b/service/src/main/java/ai/floedb/floecat/service/metagraph/loader/NodeLoader.java
@@ -79,25 +79,23 @@ public class NodeLoader {
     return catalogRepository.listIds(accountId);
   }
 
+  // These fetch the pointer-only meta once and hydrate from the blob it names (via load), rather
+  // than reading the canonical pointer twice — once in getById and again in pointerMetaForSafe.
+  // Called per-item in listing loops, so the extra pointer read added up.
+
   public Optional<NamespaceNode> namespace(ResourceId id) {
     if (id.getKind() != ResourceKind.RK_NAMESPACE) return Optional.empty();
-    return namespaceRepository
-        .getById(id)
-        .map(ns -> toNamespaceNode(ns, namespaceRepository.pointerMetaForSafe(id)));
+    return mutationMeta(id).flatMap(meta -> load(id, meta)).map(NamespaceNode.class::cast);
   }
 
   public Optional<UserTableNode> table(ResourceId id) {
     if (id.getKind() != ResourceKind.RK_TABLE) return Optional.empty();
-    return tableRepository
-        .getById(id)
-        .map(t -> toTableNode(t, tableRepository.pointerMetaForSafe(id)));
+    return mutationMeta(id).flatMap(meta -> load(id, meta)).map(UserTableNode.class::cast);
   }
 
   public Optional<ViewNode> view(ResourceId id) {
     if (id.getKind() != ResourceKind.RK_VIEW) return Optional.empty();
-    return viewRepository
-        .getById(id)
-        .map(v -> toViewNode(v, viewRepository.pointerMetaForSafe(id)));
+    return mutationMeta(id).flatMap(meta -> load(id, meta)).map(ViewNode.class::cast);
   }
 
   /**
--- a/service/src/main/java/ai/floedb/floecat/service/metagraph/loader/NodeLoader.java
+++ b/service/src/main/java/ai/floedb/floecat/service/metagraph/loader/NodeLoader.java
@@ -83,28 +83,35 @@ public class NodeLoader {
     if (id.getKind() != ResourceKind.RK_NAMESPACE) return Optional.empty();
     return namespaceRepository
         .getById(id)
-        .map(ns -> toNamespaceNode(ns, namespaceRepository.metaForSafe(id)));
+        .map(ns -> toNamespaceNode(ns, namespaceRepository.pointerMetaForSafe(id)));
   }
 
   public Optional<UserTableNode> table(ResourceId id) {
     if (id.getKind() != ResourceKind.RK_TABLE) return Optional.empty();
-    return tableRepository.getById(id).map(t -> toTableNode(t, tableRepository.metaForSafe(id)));
+    return tableRepository
+        .getById(id)
+        .map(t -> toTableNode(t, tableRepository.pointerMetaForSafe(id)));
   }
 
   public Optional<ViewNode> view(ResourceId id) {
     if (id.getKind() != ResourceKind.RK_VIEW) return Optional.empty();
-    return viewRepository.getById(id).map(v -> toViewNode(v, viewRepository.metaForSafe(id)));
+    return viewRepository
+        .getById(id)
+        .map(v -> toViewNode(v, viewRepository.pointerMetaForSafe(id)));
   }
 
-  /** Loads the mutation metadata for the provided resource. */
+  /**
+   * Loads the mutation metadata for the provided resource. Graph consumers only use the pointer
+   * version (cache key) and timestamps, so this is a pointer-only read — no blob HEAD, blank etag.
+   */
   public Optional<MutationMeta> mutationMeta(ResourceId id) {
     try {
       ResourceKind kind = id.getKind();
       return switch (kind) {
-        case RK_CATALOG -> Optional.of(catalogRepository.metaForSafe(id));
-        case RK_NAMESPACE -> Optional.of(namespaceRepository.metaForSafe(id));
-        case RK_TABLE -> Optional.of(tableRepository.metaForSafe(id));
-        case RK_VIEW -> Optional.of(viewRepository.metaForSafe(id));
+        case RK_CATALOG -> Optional.of(catalogRepository.pointerMetaForSafe(id));
+        case RK_NAMESPACE -> Optional.of(namespaceRepository.pointerMetaForSafe(id));
+        case RK_TABLE -> Optional.of(tableRepository.pointerMetaForSafe(id));
+        case RK_VIEW -> Optional.of(viewRepository.pointerMetaForSafe(id));
         default -> Optional.empty();
       };
     } catch (StorageNotFoundException snf) {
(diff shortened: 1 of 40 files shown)


jchauzi-floe (commented):
Reviewed the query-resolution perf changes. Two bugs in the new batching/hydration logic, one backward-compat gap that undercuts the stated savings, and a handful of cleanup items — all left as inline comments.

**Correctness bugs (2):** stale cross-instance reads in `NodeLoader.load()`, and scope-memo cache poisoning in `NameResolver` when a blank-name ref shares a batch with a valid one.

**Backward-compat gap (1):** `relationNameClaim` is a net regression for any table/view created before the claim mechanism existed (no backfill), and for every not-found lookup — worth reflecting in the PR's savings numbers.

**Cleanup (4):** a missed double pointer-read in `NodeLoader`, a dead branch in `pointerMetaForSafe`, two independently-built memoization schemes for the same batching problem, and `getByBlobUri` living on the generic repository base for what's really a `NodeLoader`-only concern.

One more not tied to a single line: none of the new fast paths (`getByBlobUri`, `relationNameClaim`) are actually exercised by the test suite — the `Fake*Repository` test doubles bypass the real blob store, so `getByBlobUri` always falls through to the old path in tests, and `relationNameClaim` is never populated. That's how the two bugs above ship without a failing test.

service/src/main/java/ai/floedb/floecat/service/metagraph/loader/NodeLoader.java
```diff
@@ -83,43 +83,71 @@ public Optional<NamespaceNode> namespace(ResourceId id) {
...
 
-  /** Rehydrates the relation node for the provided metadata snapshot. */
+  /**
+   * Rehydrates the relation node for the provided metadata snapshot. The metadata already names the
+   * blob, so the blob is fetched directly — skipping the pointer re-read {@code getById} would do —
+   * and the node content stays consistent with the metadata's pointer version. Falls back to a
+   * pointer-based read when the blob has moved (e.g. updated and garbage-collected in between).
+   */
   public Optional<GraphNode> load(ResourceId id, MutationMeta meta) {
+    String blobUri = meta.getBlobUri();
```
jchauzi-floe, line 127: **Bug: this can serve stale content across instances, not just a stale version tag.**

`meta` here comes from `GraphCacheManager`'s meta cache (`UserGraph.resolve()` calls `cache.getMeta(id)` and only refreshes it after the TTL expires, default 2s). That cache is a plain in-process Caffeine cache — `invalidate()` only runs in the writer's own process, there's no cross-instance invalidation.

Before this change, `load()` always called `repo.getById(id)`, i.e. a fresh pointer read, so a stale `meta` only mislabeled the version — the actual content returned was always current. Now it fetches the blob named by `meta.getBlobUri()` directly. Since CAS blobs aren't GC'd until `min-age-ms` after becoming orphaned (30s default), way longer than the meta TTL, the old blob is still readable.

Concretely: instance A updates a table. Within the ~2s TTL window, a request on instance B with a still-cached (pre-update) `meta` hits `load()`, and gets the **old** blob back — served as if it were current. That's a real content-staleness regression, not just an eventually-consistent version counter.

```java
String blobUri = meta.getBlobUri();
```

No test catches this: `Fake*Repository.put()` in the test doubles never populates the underlying blob store, so `getByBlobUri` always misses in tests and silently falls through to `getById()` — the fast path this PR adds is never actually exercised.

service/src/main/java/ai/floedb/floecat/service/metagraph/resolver/NameResolver.java
```diff
@@ -175,16 +178,49 @@ public Optional<ResourceId> resolveViewId(String cid, String accountId, NameRef
...
+  private Optional<ResourceId> resolveRelationId(
+      String accountId, NameRef ref, Map<String, Optional<Scope>> scopeMemo) {
     return diagnose(
         "resolve_relation_id",
         "",
         accountId,
         () ->
-            resolveScope(accountId, ref)
+            scopeMemo
+                .computeIfAbsent(scopeKey(ref), k -> resolveScope(accountId, ref))
```
jchauzi-floe: **Bug: cache poisoning when a blank/invalid name shares scope with a valid one in the same batch.**

`scopeKey(ref)` only encodes `catalog` + `path` — not `name`:

```java
private static String scopeKey(NameRef ref) {
  return ref.getCatalog() + "\u001F" + String.join("\u001F", ref.getPathList());
}
```

But `resolveScope` short-circuits to `Optional.empty()` when `validName(ref)` fails (blank name). Since that result gets cached under the name-independent key via `computeIfAbsent`, the first blank-name ref in a batch poisons the scope entry for every other ref sharing that catalog/path.

`UserGraph.resolveNames`'s validation only rejects a blank catalog, not a blank name, so this is reachable — a batch like `[NameRef(catalog="c1", name=""), NameRef(catalog="c1", name="orders")]` will report `orders` as not-found even though it exists, purely because the empty-name entry got processed first and cached `Optional.empty()` for the `"c1"` scope key.

service/src/main/java/ai/floedb/floecat/service/metagraph/resolver/NameResolver.java
```diff
@@ -175,16 +178,49 @@ public Optional<ResourceId> resolveViewId(String cid, String accountId, NameRef
...
                 .flatMap(
                     scope -> {
                       String catalogId = scope.catalog().getResourceId().getId();
                       String namespaceId = scope.namespace().getResourceId().getId();
+                      // One pointer read (no blob fetch) answers both kind and id via the shared
+                      // relation-name claim written by every table/view create and rename.
+                      Optional<ResourceId> claimed =
+                          tableRepository.relationNameClaim(
```
jchauzi-floe: **Backward-compatibility note, not a correctness issue — but worth flagging since it undercuts the perf claim.**

The relation-name claim this reads was introduced 3 days before this PR with no backfill. For every table/view created before that (i.e. presumably most rows in any existing account today), this probe misses and falls through to the exact same table-then-view dual-probe as before — so those lookups now do the claim read **plus** the old work, not instead of it. Same for every not-found lookup (typo'd or nonexistent name): permanent extra pointer read, not a transitional cost.

So this is a net regression for existing data and for miss lookups until a backfill runs — and there's no backfill job in this PR or planned. Worth calling out explicitly in the PR description's savings numbers, since they don't seem to account for this.

service/src/main/java/ai/floedb/floecat/service/metagraph/loader/NodeLoader.java
```diff
@@ -83,43 +83,71 @@ public Optional<NamespaceNode> namespace(ResourceId id) {
     if (id.getKind() != ResourceKind.RK_NAMESPACE) return Optional.empty();
     return namespaceRepository
         .getById(id)
-        .map(ns -> toNamespaceNode(ns, namespaceRepository.metaForSafe(id)));
+        .map(ns -> toNamespaceNode(ns, namespaceRepository.pointerMetaForSafe(id)));
```
jchauzi-floe: **Missed optimization — same redundancy this PR fixes elsewhere.**

`getById(id)` already does a pointer read internally; `pointerMetaForSafe(id)` right after it does a second, independent pointer read for the same id. Same pattern at lines 93 and 100 (`table()`, `view()`).

This PR already knows how to avoid exactly this: `BaseResourceRepository.readMetaOrDefault(Optional<Pointer>, ...)` was added so `metaFor`/`metaForSafe` can reuse a pointer the caller already fetched instead of re-reading it. These three methods (called per-item in listing loops, e.g. `UserGraph.java:482,488,510,515`) would benefit from the same treatment.

service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java
```diff
@@ -603,19 +604,44 @@ public MutationMeta metaForSafe(K key, Timestamp nowTs) {
...
+   * precondition checks.
+   */
+  public MutationMeta pointerMetaForSafe(K key) {
+    return observeRepository(
+        "pointer_meta_for_safe",
+        () -> {
+          Timestamp nowTs = Timestamps.fromMillis(clock.millis());
+          String canonical = schema.canonicalPointerForKey.apply(key);
+          var ptrOpt = pointerStore.get(canonical);
+          String blobUri = ptrOpt.isEmpty() && schema.casBlobs ? "" : blobUriFor(key, ptrOpt);
```
jchauzi-floe: **Dead branch — this line is equivalent to just calling `blobUriFor(key, ptrOpt)`.**

```java
String blobUri = ptrOpt.isEmpty() && schema.casBlobs ? "" : blobUriFor(key, ptrOpt);
```

`blobUriFor` (below) already returns `""` for exactly the `ptrOpt.isEmpty() && schema.casBlobs` case — that's its `casBlobs` branch when `ptrOpt.isPresent()` is false. So both arms of this ternary produce what `blobUriFor(key, ptrOpt)` alone would produce. `metaForSafe` right above (this same PR) does it the simple way:

```java
String blobUri = blobUriFor(key, ptrOpt);
```

Suggest matching that here.

service/src/main/java/ai/floedb/floecat/service/metagraph/resolver/FullyQualifiedResolver.java
```diff
@@ -196,18 +199,38 @@ public ResolveResult resolveViewsByPrefix(
   // Internal helpers (canonical entry resolution)
   // ----------------------------------------------------------------------
 
-  private Optional<QualifiedRelation> resolveTableEntry(String cid, String accountId, NameRef ref) {
+  /**
+   * Per-call scope memo: a batch of names typically shares one catalog.namespace, so resolve each
+   * distinct scope once instead of re-reading catalog and namespace for every name.
+   */
+  private final class ScopeMemo {
```
jchauzi-floe: **Two different memoization schemes added in this same PR for the same problem.**

This `ScopeMemo` caches catalog and namespace in separate maps (namespace keyed by catalog **id** + path). `NameResolver.resolveRelationIds` (added in the same PR, `NameResolver.java` around line 190) memoizes catalog+namespace as one combined `Scope`, keyed by catalog **display name** + path.

Different granularity (this one dedupes catalog reads across namespaces in the same catalog; `NameResolver`'s doesn't) and different key composition, for what the PR description calls the same optimization. A future fix to one (TTL, negative-result caching, the `\u001F`-key-collision edge case) won't apply to the other. Might be worth one shared memo abstraction instead of two call-site-specific ones.

service/src/main/java/ai/floedb/floecat/service/repo/util/BaseResourceRepository.java
```diff
@@ -399,6 +407,33 @@ protected MutationMeta readMetaOrDefault(String pointerKey, String blobUri, Time
         .build();
   }
 
+  /**
+   * Fetches and parses a resource directly by blob URI, skipping the pointer read. Intended for
+   * callers that already resolved the pointer (e.g. graph hydration from cached metadata). Returns
+   * empty when the blob is absent — callers should fall back to a pointer-based read, since the
+   * pointer may have moved to a new blob in the meantime.
+   */
+  public Optional<T> getByBlobUri(String blobUri) {
```
jchauzi-floe: **Generic placement for a caller-specific policy.**

`getByBlobUri` (and `pointerMetaForSafe` in `GenericResourceRepository`) land on the shared base every repository composes, so `AccountRepository`, `SnapshotRepository`, `TransactionRepository`, etc. all pick this up even though none of them use it — only `NodeLoader` does, through 4 of the 13 repositories. The "fetch by blob URI, fall back to fetch by pointer" decision is `NodeLoader.load()`'s concern; putting the primitive here means nothing marks it as scoped to that one caller, so it's easy for someone to reach for it on a repo where "read by blob content regardless of current pointer" isn't actually safe (e.g. one with GC semantics tied to pointer state).

mcusack-floe (approved):

jchauzi-floe (approved):

Issue #361, opened 2026-07-03
[Bug]: principal-context loss in GetUserObjects can resolve tables as NOT_FOUND instead of failing (silent 42P01 in engine)


jchauzi-floe:
The known principal-context propagation race (context lost between interceptor and handler while the MDC on the same log line is fully populated) has a second, quieter failure mode: instead of `PERMISSION_DENIED: missing permission: catalog.read`, `GetUserObjects` completes OK and resolves the requested table as `RESOLUTION_STATUS_NOT_FOUND`. The engine turns that into `42P01 relation "sys.const" does not exist` on a plain `SELECT $1::int4` — a wrong-answer-shaped error with no floecat-side trace.

## Evidence

core develop run [28621661587](https://github.com/eng-floe/core/actions/runs/28621661587) (linux-arm test-sql), floecat pin 157722a3:

```
2026-07-02 22:50:58,875 INFO  [UserObjectsServiceImpl] op=GetUserObjects start
2026-07-02 22:50:58,883 INFO  [UserObjectsServiceImpl] op=GetUserObjects ok query_id=860653811620909056 correlation_id= tables=1 dispatchMs=0.1 outcome=completed elapsedMs=8.1
```

`correlation_id=` is empty, yet the MDC of that same line has everything: `correlation_id=b78f1ccf-889e-44db-863e-1c56279724dd, floecat_subject=dev-user, floecat_account_id=5eaa9cd5-…`. So the request metadata arrived intact; the io.grpc.Context was lost between the interceptor and the handler. Every neighboring `GetUserObjects` (~100ms on either side) has a proper correlation id and resolves the same `sys.const` candidate fine.

Same signature in run [28577593280](https://github.com/eng-floe/core/actions/runs/28577593280) 12h earlier. Original PERMISSION_DENIED form reproduced with AUTHDIAG in core run 27714312054 (principal at `Authorizer.require` was `accountId='' subject='' grantedPermissions=[]`). Sibling: #335.

## Mechanism

*(rewritten 2026-07-03 after full RCA — see the comment below for the verified details; the original text here wrongly blamed a blank-account authz fallback)*

1. The io.grpc.Context is lost in the multi-hop relay (event loop → duplicated context → blocking workers → Mutiny executor): under Quarkus, attach/detach on a duplicated context is a read/write of one shared last-writer-wins slot, several threads of the same call are associated with that context concurrently, and every relay point re-captures `Context.current()` instead of passing the resolved context by reference.
[...]

PR #372, opened 2026-07-06
Fix silent call-context loss: carry the whole ResolvedCallContext on one write-once channel

mrouvroy-floe:
Part 1 of 2 for #361 (part 2: context-loss hardening, stacked on this branch).

## Problem

One request's context rides three channels with very different reliability. MDC and the principal survive Quarkus's executor hops via Vert.x duplicated-context locals, but the **correlation id and engine context lived only on `io.grpc.Context` keys** — which, under Quarkus, are one shared last-writer-wins slot that the call's *own* threads race (event loop, blocking-wrap workers, Mutiny emitter). Mirroring only the principal (the earlier partial fix) moved the failure past authz: instead of a loud `PERMISSION_DENIED`, the engine context silently reads back empty, the engine-keyed system graph resolves nothing, and `GetUserObjects` completes OK with `sys.const` → `RESOLUTION_STATUS_NOT_FOUND` — surfacing as a wrong-answer-shaped `42P01` in the engine with no floecat-side trace (see the RCA in #361).

## Fix

**Commit 1 — test first.** `QueryContextPropagationIT` gains two silent-form detectors: every streaming attempt resolves `sys.const`, an engine-gated system table served by the test-only `TestCatalogExtension` (mirroring how engine extensions ship system catalogs in production; only resolvable while the request's engine context is visible), and a new scenario asserts the `NOT_FOUND` error payload echoes the request correlation id — that value is read *inside the service body* from the fragile channel, so it probes exactly the read that went empty in CI.

**Commit 2 — one write-once carrier.** New `ResolvedCallContexts` stores the whole `ResolvedCallContext` (principal + correlation + engine + query id + auth header values) once per call on the same duplicated-context local that carries MDC, plus an explicit `callWith`/`runWith` scope for handing it by reference across executor hops. All readers now go carrier-first, grpc-keys last:
- `PrincipalProvider.get()` (replaces the principal-only mirror)
- `EngineContextProvider.engineContext()` (previously grpc-keys only — the resolution-side hole)
- `GrpcResolvedCallContexts` (Flight path)
- `OutboundContextClientInterceptor` (outbound RPCs from executor threads no longer drop engine/correlation headers)
- `BaseServiceImpl.run` (every unary body)

The four streaming services (`UserObjects`, `QueryScan`, `QuerySystemScan`, `PlannerStats`) read the resolved context **once at method entry**, before any hop, and carry it by reference into their bodies. `InboundCallContextHelper` also stamps the correlation id onto the `PrincipalContext` — the field existed but was never set.

**Commit 3 — engine by value through resolution.** `CatalogOverlay` gains engine-explicit `resolveName`/`resolve` overloads (default-delegating, so other implementers are unaffected); the bundle iterator passes the engine captured at construction instead of re-reading the request context per lookup (`resolve(id)` had the same per-lookup hole as `resolveName` — fixing only names would have moved the failure to the node stage). `MetaGraph.resolveName` now WARNs when a lookup misses both graphs with an empty engine instead of silently returning NOT_FOUND.

## Verification

- 926 unit tests + 29 ITs green after every commit; re-verified after rebasing onto `origin/main`.
- Stress: 9,600 calls at 32 threads across the three IT scenarios — zero context losses, zero silent NOT_FOUNDs, zero correlation mismatches.
- Honest caveat: the race does not reproduce locally even pre-fix (same reason the original PERMISSION_DENIED IT went quiet), so the new tests are armed detectors validated against the fixed code. The acceptance signal is the two CI signatures from #361 staying gone on core test-sql runs: empty `correlation_id=` on `GetUserObjects` ok-lines, and `sys.*` NOT_FOUND on constant-only statements.

Engine-side follow-up (bare `42P01` without the `ResolutionFailure` payload) stays tracked in eng-floe/core#1692.

### DIFF
--- a/core/scanner/src/main/java/ai/floedb/floecat/scanner/spi/CatalogOverlay.java
+++ b/core/scanner/src/main/java/ai/floedb/floecat/scanner/spi/CatalogOverlay.java
@@ -30,6 +30,7 @@ import ai.floedb.floecat.metagraph.model.TypeNode;
 import ai.floedb.floecat.metagraph.model.UserTableNode;
 import ai.floedb.floecat.query.rpc.SchemaColumn;
 import ai.floedb.floecat.query.rpc.SnapshotPin;
+import ai.floedb.floecat.scanner.utils.EngineContext;
 import com.google.protobuf.Timestamp;
 import java.util.List;
 import java.util.Optional;
@@ -48,6 +49,18 @@ public interface CatalogOverlay {
 
   /** Resolves any graph node for the given resource. Engine context is resolved implicitly. */
   Optional<GraphNode> resolve(ResourceId id);
+
+  /**
+   * Resolves any graph node for the given resource using an explicit engine context.
+   *
+   * <p>Prefer this overload wherever the caller already holds the request's engine context (e.g.
+   * from a {@link MetadataResolutionContext}): re-reading the engine from the request context on
+   * every lookup is fragile across executor hops, and a silently empty engine makes engine-gated
+   * system objects unresolvable.
+   */
+  default Optional<GraphNode> resolve(ResourceId id, EngineContext engineContext) {
+    return resolve(id);
+  }
 
   /**
    * Lists every relation under the requested catalog (namespaces, tables, views, plus system
@@ -173,6 +186,17 @@ public interface CatalogOverlay {
       out.computeIfAbsent(ref, r -> resolveName(correlationId, r));
     }
     return out;
+  }
+
+  /**
+   * Resolves a relation (table or view) by name reference using an explicit engine context.
+   *
+   * <p>Prefer this overload wherever the caller already holds the request's engine context — see
+   * {@link #resolve(ResourceId, EngineContext)}.
+   */
+  default Optional<ResourceId> resolveName(
+      String correlationId, NameRef ref, EngineContext engineContext) {
+    return resolveName(correlationId, ref);
   }
 
   /** Resolves a system table name without involving the user graph. */
--- a/core/scanner/src/main/java/ai/floedb/floecat/scanner/spi/CatalogOverlay.java
+++ b/core/scanner/src/main/java/ai/floedb/floecat/scanner/spi/CatalogOverlay.java
@@ -57,6 +57,11 @@ public interface CatalogOverlay {
    * from a {@link MetadataResolutionContext}): re-reading the engine from the request context on
    * every lookup is fragile across executor hops, and a silently empty engine makes engine-gated
    * system objects unresolvable.
+   *
+   * <p><b>Implementers beware:</b> this default delegates to {@link #resolve(ResourceId)} and
+   * <em>ignores</em> the passed engine context. Any implementation that serves engine-gated objects
+   * must override it, or callers passing an explicit engine silently lose it (test doubles included
+   * — a fake inheriting this default cannot catch engine-threading regressions).
    */
   default Optional<GraphNode> resolve(ResourceId id, EngineContext engineContext) {
     return resolve(id);
@@ -192,7 +197,8 @@ public interface CatalogOverlay {
    * Resolves a relation (table or view) by name reference using an explicit engine context.
    *
    * <p>Prefer this overload wherever the caller already holds the request's engine context — see
-   * {@link #resolve(ResourceId, EngineContext)}.
+   * {@link #resolve(ResourceId, EngineContext)}, including its note that this default ignores the
+   * passed engine context for implementations that do not override it.
    */
   default Optional<ResourceId> resolveName(
       String correlationId, NameRef ref, EngineContext engineContext) {
--- a/service/src/main/java/ai/floedb/floecat/service/common/BaseServiceImpl.java
+++ b/service/src/main/java/ai/floedb/floecat/service/common/BaseServiceImpl.java
@@ -25,6 +25,8 @@ import ai.floedb.floecat.common.rpc.Precondition;
 import ai.floedb.floecat.common.rpc.ResourceId;
 import ai.floedb.floecat.common.rpc.ResourceKind;
 import ai.floedb.floecat.connector.rpc.NamespacePath;
+import ai.floedb.floecat.flight.context.ResolvedCallContext;
+import ai.floedb.floecat.service.context.impl.ResolvedCallContexts;
 import ai.floedb.floecat.service.error.impl.GrpcErrors;
 import ai.floedb.floecat.service.repo.util.BaseResourceRepository;
 import ai.floedb.floecat.service.security.impl.PrincipalProvider;
@@ -85,16 +87,23 @@ public abstract class BaseServiceImpl {
 
   protected <T> Uni<T> run(Supplier<T> body) {
     GrpcContextUtil grpcCtx = GrpcContextUtil.capture();
+    // Read the resolved call context at method entry — before any executor hop — and carry it by
+    // reference into the body. The captured io.grpc.Context alone is unreliable across the hop
+    // (eng-floe/floecat#361).
+    ResolvedCallContext callCtx = ResolvedCallContexts.currentOrNull();
     Context otelCtx = Context.current();
     return Uni.createFrom()
         .item(
             () ->
                 grpcCtx.call(
-                    () -> {
-                      try (Scope ignored = otelCtx.makeCurrent()) {
-                        return body.get();
-                      }
-                    }));
+                    () ->
+                        ResolvedCallContexts.callWith(
+                            callCtx,
+                            () -> {
+                              try (Scope ignored = otelCtx.makeCurrent()) {
+                                return body.get();
+                              }
+                            })));
   }
 
   protected <T> Uni<T> runWithRetry(Supplier<T> body) {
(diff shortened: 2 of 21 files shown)


service/src/main/java/ai/floedb/floecat/service/context/impl/OutboundContextClientInterceptor.java
```diff
@@ -61,14 +62,30 @@ public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
...
-            Optional.ofNullable(InboundContextInterceptor.QUERY_KEY.get())
+            Optional.ofNullable(
+                    firstNonBlank(
+                        resolved != null ? resolved.queryId() : null,
+                        InboundContextInterceptor.QUERY_KEY.get()))
                 .orElseGet(() -> Baggage.current().getEntryValue("query_id"));
 
-        EngineContext engineContext = InboundContextInterceptor.ENGINE_CONTEXT_KEY.get();
+        EngineContext engineContext =
+            resolved != null && resolved.engineContext().hasEngineKind()
```
jchauzi-floe: Falls back to the stale `io.grpc.Context` key whenever `resolved`'s own field is legitimately empty, not just when `resolved` is absent:

```java
EngineContext engineContext =
    resolved != null && resolved.engineContext().hasEngineKind()
        ? resolved.engineContext()
        : InboundContextInterceptor.ENGINE_CONTEXT_KEY.get();
```

Worker thread handles call A with `engine=trino`, gets reused for an engine-less call B. `resolved.engineContext()` correctly says "no engine" for B, but since `hasEngineKind()` is false this falls through to `ENGINE_CONTEXT_KEY.get()` — which can still hold A's stale value via the `ContextStorageOverride` thread-local fallback. Same conflation for `queryId` (line 78) and `correlationId` (line 107) via `firstNonBlank(resolved.xxx(), LEGACY_KEY.get())`.

`PrincipalProvider.get()` a few files over does this right: `if (resolved != null) return resolved.principalContext();` — trust it once non-null, don't inspect the field.

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -1812,7 +1812,11 @@ private Optional<ResourceId> resolveNameCached(NameRef ref) {
       }
       long startNs = System.nanoTime();
       try {
-        Optional<ResourceId> resolved = overlay.resolveName(correlationId, ref);
+        // Pass the engine captured at iterator construction: re-reading it from the request
+        // context per lookup is fragile across executor hops, and an empty engine silently
+        // un-resolves engine-gated system objects (eng-floe/floecat#361).
+        Optional<ResourceId> resolved =
+            overlay.resolveName(correlationId, ref, resolutionContext.engineContext());
```
jchauzi-floe, line 1822: Fixed here, but `buildRelation` (line 563) has the identical problem and wasn't touched:

```java
EngineContext ctx = engineContext.engineContext();
```

That's a live re-read of the request-scoped provider, not `resolutionContext.engineContext()` — the same fragile-across-executor-hops read this PR is fixing everywhere else. If it diverges, `decorationRequired(ctx)` silently goes false and engine-specific decoration gets skipped with no log line. `resolutionContext` is already a parameter to `buildRelation`, so it's a one-line change.

service/src/main/java/ai/floedb/floecat/service/metagraph/overlay/MetaGraph.java
```diff
@@ -308,12 +321,30 @@ public Optional<ResourceId> resolveView(String correlationId, NameRef ref) {
...
+  public Optional<ResourceId> resolveName(
+      String correlationId, NameRef ref, EngineContext engineContext) {
+    EngineContext ctx = orRequestEngine(engineContext, this::engineContext);
+    Optional<ResourceId> resolved =
+        resolveWithAmbiguityCheck(
+            correlationId,
+            ref,
+            () -> systemGraph.resolveName(ref, ctx),
+            () -> userGraph.resolveName(correlationId, ref));
+    if (resolved.isEmpty() && !ctx.hasEngineKind()) {
```
jchauzi-floe, line 327: This WARNs on every miss with an empty engine context, not just real context loss. `client-cli`'s `query begin table <name>` never sends an engine header, so a plain typo in a table name from the CLI hits this on the resolution hot path. Only `GetSystemObjects` actually requires `engine_kind` — nothing stops an engine-less caller from routinely tripping this.

core/scanner/src/main/java/ai/floedb/floecat/scanner/spi/CatalogOverlay.java
```diff
@@ -49,6 +50,18 @@ public interface CatalogOverlay {
...
 
+  /**
+   * Resolves any graph node for the given resource using an explicit engine context.
+   *
+   * <p>Prefer this overload wherever the caller already holds the request's engine context (e.g.
+   * from a {@link MetadataResolutionContext}): re-reading the engine from the request context on
+   * every lookup is fragile across executor hops, and a silently empty engine makes engine-gated
+   * system objects unresolvable.
+   */
+  default Optional<GraphNode> resolve(ResourceId id, EngineContext engineContext) {
```
jchauzi-floe, line 66: Default silently drops the passed engine context for any implementer that doesn't override it (same pattern for `resolveName` below, line 183):

```java
default Optional<GraphNode> resolve(ResourceId id, EngineContext engineContext) {
  return resolve(id);
}
```

`MetaGraph` is the only overrider. `FakeCatalogOverlay` — used by `UserObjectBundleServiceTest`, the test covering this exact fix — inherits the default, so that test can't catch a regression that stops threading the engine context through. Doesn't hit prod today (`MetaGraph` is the only CDI binding) but worth at least a javadoc warning that non-overriders silently ignore the param.

service/src/main/java/ai/floedb/floecat/service/common/BaseServiceImpl.java
```diff
@@ -85,16 +87,23 @@ protected long idempotencyTtlSeconds() {
 
   protected <T> Uni<T> run(Supplier<T> body) {
```
jchauzi-floe, line 93: This generalizes the capture-and-wrap fix for `Uni`-returning unary RPCs. There's no equivalent for `Multi` streaming RPCs — it's hand-rolled identically at 6 call sites across `PlannerStatsServiceImpl`, `QueryScanServiceImpl`, `QuerySystemScanServiceImpl`, `UserObjectsServiceImpl`. Nothing stops the next streaming method from skipping this and reintroducing #361. Might be worth a streaming equivalent here (or on `GrpcContextUtil`) so it's structural instead of copy-paste.

service/src/main/java/ai/floedb/floecat/service/common/BaseServiceImpl.java
```diff
@@ -207,6 +216,10 @@ protected static boolean hasMeaningfulPrecondition(Precondition precondition) {
   }
 
   protected String correlationId() {
+    ResolvedCallContext resolved = ResolvedCallContexts.currentOrNull();
+    if (resolved != null && !resolved.correlationId().isBlank()) {
```
jchauzi-floe: This fallback (`callCtx.correlationId()` if non-blank, else `principalContext.getCorrelationId()`) is reimplemented at 4 places: here, `PlannerStatsServiceImpl` (twice), `UserObjectsServiceImpl`, and `QueryScanServiceImpl` (which pulled it into a private `correlationIdOf` helper). Worth putting on `ResolvedCallContext` as `effectiveCorrelationId()` — this exact logic already drifted once before this PR (`UserObjectsServiceImpl` used to source its fallback from `CORR_KEY` instead of the principal).

service/src/main/java/ai/floedb/floecat/service/context/impl/ResolvedCallContexts.java
```diff
@@ -0,0 +1,189 @@
...
+   * plumbing bug and is logged loudly rather than silently masked.
+   */
+  public static void storeOnDuplicatedContext(ResolvedCallContext resolved) {
+    Objects.requireNonNull(resolved, "resolved");
+    io.vertx.core.Context ctx = Vertx.currentContext();
+    if (ctx == null || !VertxContext.isDuplicatedContext(ctx)) {
+      return;
+    }
+    Object existing = ctx.getLocal(CONTEXT_LOCAL);
+    if (existing != null && !existing.equals(resolved)) {
```
jchauzi-floe, line 82: `existing.equals(resolved)` degrades to reference identity for `engineContext` — `EngineContext` has no `equals()`/`hashCode()` and `EngineContext.of(...)` never interns non-empty instances. Not reachable today (this has exactly one call site, invoked once per call), but if a second path to `storeOnDuplicatedContext` ever gets added, two value-identical contexts will spuriously trip the "plumbing bug" WARN below. Cheap to close now: add `equals()`/`hashCode()` to `EngineContext`.

jchauzi-floe (commented):
Reviewed the context-carrier changes end to end (traced every `callWith`/`runWith` call site and cross-checked against production callers). The core fix holds up. 7 issues below, most severe first.

jchauzi-floe (approved):

PR #373, opened 2026-07-06
Context-loss hardening: loud failures, disagreement detectors, deterministic interceptor ordering

### LINKED ISSUE #361: [Bug]: principal-context loss in GetUserObjects can resolve tables as NOT_FOUND instead of failing (silent 42P01 in engine)

mrouvroy-floe:
Part 2 of 2 for #361, stacked on #372 (the fix). Closes #361.

Every context channel without a loud-failure guard fails **silently** when the next propagation bug arrives — that is the lesson of the `PERMISSION_DENIED` → silent `NOT_FOUND` shift documented in the #361 RCA. This PR adds the guards, the detection, and the deterministic interceptor ordering so a future loss is loud, countable, and diagnosable.

## Commit 1 — loud failures + channel-disagreement detectors

- **`Authorizer`**: a principal with no subject and no permissions is not a caller lacking a grant — it is a caller whose identity never arrived. Report `UNAUTHENTICATED` (with the attempted permission in the description) instead of the misleading `PERMISSION_DENIED`. Real authz failures (subject present, grant missing) still get `PERMISSION_DENIED`.
- **`ResolvedCallContexts`**: WARN when MDC carries a correlation id but no context channel delivers a resolved call context. MDC and the carrier are written together in the inbound interceptor, so any disagreement is a propagation loss — turned into a countable signal.
- **`EngineContextProvider`**: the analogous WARN when MDC says the request declared an engine but resolution sees an empty engine context — the exact condition that silently un-resolves engine-gated system objects.
- `QueryContextPropagationIT` treats `UNAUTHENTICATED` in the stress loop as a context-loss failure alongside `PERMISSION_DENIED`.

Both WARNs should sit at flat zero in CI logs; any occurrence is a propagation regression, caught before it becomes a wrong answer.

## Commit 2 — deterministic interceptor ordering + relay test

Quarkus's gRPC interceptor comparator only honors the CDI `Prioritized` **interface** (verified against the Quarkus 3.31.4 sources); the `@Priority` annotations on our three global interceptors were silently ignored, so they all tied at 0 with nondeterministic relative order — and the `@Priority(0) // Must run before other interceptors` comment had the sort semantics backwards either way (ascending sort, last element outermost per `ServerInterceptors.intercept`).

- Implement `Prioritized` on all four global interceptors with the intended, now-guaranteed order (outermost first): inbound context → telemetry → rpc logging → error localization. Quarkus's built-ins near `Integer.MAX_VALUE` stay outermost of all.
- Priorities and the sort semantics are documented in one place, `GrpcInterceptorPriorities`.
- Add the relay-level test from the RCA: inbound interceptor (production key wiring + carrier) → `CtxPropagatingBlockingWrap` → a streaming-shaped body that reads the resolved context at entry and re-emits on a bare executor, under concurrent listener events — asserting principal, correlation id, and engine context integrity at both the service and emitter layers, including on threads where only the explicit `runWith` scope can carry the context.

## Verification

926 unit tests + 29 ITs green (including with the new deterministic interceptor order); propagation IT stress at 32×100 clean. Re-verified after rebasing onto `origin/main`.

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/context/EngineContextProvider.java
+++ b/service/src/main/java/ai/floedb/floecat/service/context/EngineContextProvider.java
@@ -22,6 +22,8 @@ import ai.floedb.floecat.service.context.impl.InboundContextInterceptor;
 import ai.floedb.floecat.service.context.impl.ResolvedCallContexts;
 import jakarta.enterprise.context.ApplicationScoped;
 import java.util.Optional;
+import org.jboss.logging.Logger;
+import org.jboss.logging.MDC;
 
 /**
  * Resolves the engine context declared by the current call.
@@ -32,6 +34,8 @@ import java.util.Optional;
  */
 @ApplicationScoped
 public final class EngineContextProvider {
+
+  private static final Logger LOG = Logger.getLogger(EngineContextProvider.class);
 
   public String engineKind() {
     return engineContext().engineKind();
@@ -69,6 +73,27 @@ public final class EngineContextProvider {
     String kind = Optional.ofNullable(InboundContextInterceptor.ENGINE_KIND_KEY.get()).orElse("");
     String version =
         Optional.ofNullable(InboundContextInterceptor.ENGINE_VERSION_KEY.get()).orElse("");
-    return EngineContext.of(kind, version);
+    EngineContext fallback = EngineContext.of(kind, version);
+    if (!fallback.hasEngineKind()) {
+      warnIfRequestDeclaredEngine();
+    }
+    return fallback;
+  }
+
+  /**
+   * Channel-disagreement detector: MDC carrying an engine kind proves the request declared an
+   * engine, so resolving an empty engine context here is a propagation loss — the exact condition
+   * that silently un-resolves engine-gated system objects (eng-floe/floecat#361). This should never
+   * fire; any occurrence is a regression in context propagation.
+   */
+  private static void warnIfRequestDeclaredEngine() {
+    Object mdcEngineKind = MDC.get("floecat_engine_kind");
+    if (mdcEngineKind instanceof String engineKind && !engineKind.isBlank()) {
+      LOG.warnf(
+          "engine-context channels disagree: MDC floecat_engine_kind=%s is populated but no"
+              + " channel carries an engine context — the call context was lost on the way to"
+              + " this thread (correlation_id=%s)",
+          engineKind, MDC.get("correlation_id"));
+    }
   }
 }
--- a/service/src/main/java/ai/floedb/floecat/service/context/EngineContextProvider.java
+++ b/service/src/main/java/ai/floedb/floecat/service/context/EngineContextProvider.java
@@ -22,8 +22,6 @@ import ai.floedb.floecat.service.context.impl.InboundContextInterceptor;
 import ai.floedb.floecat.service.context.impl.ResolvedCallContexts;
 import jakarta.enterprise.context.ApplicationScoped;
 import java.util.Optional;
-import org.jboss.logging.Logger;
-import org.jboss.logging.MDC;
 
 /**
  * Resolves the engine context declared by the current call.
@@ -34,8 +32,6 @@ import org.jboss.logging.MDC;
  */
 @ApplicationScoped
 public final class EngineContextProvider {
-
-  private static final Logger LOG = Logger.getLogger(EngineContextProvider.class);
 
   public String engineKind() {
     return engineContext().engineKind();
@@ -75,25 +71,12 @@ public final class EngineContextProvider {
         Optional.ofNullable(InboundContextInterceptor.ENGINE_VERSION_KEY.get()).orElse("");
     EngineContext fallback = EngineContext.of(kind, version);
     if (!fallback.hasEngineKind()) {
-      warnIfRequestDeclaredEngine();
+      // MDC carrying an engine kind proves the request declared an engine, so an empty engine
+      // context here is the exact condition that silently un-resolves engine-gated system
+      // objects (eng-floe/floecat#361).
+      ResolvedCallContexts.warnOnChannelDisagreement(
+          "floecat_engine_kind", "no channel carries an engine context");
     }
     return fallback;
-  }
-
-  /**
-   * Channel-disagreement detector: MDC carrying an engine kind proves the request declared an
-   * engine, so resolving an empty engine context here is a propagation loss — the exact condition
-   * that silently un-resolves engine-gated system objects (eng-floe/floecat#361). This should never
-   * fire; any occurrence is a regression in context propagation.
-   */
-  private static void warnIfRequestDeclaredEngine() {
-    Object mdcEngineKind = MDC.get("floecat_engine_kind");
-    if (mdcEngineKind instanceof String engineKind && !engineKind.isBlank()) {
-      LOG.warnf(
-          "engine-context channels disagree: MDC floecat_engine_kind=%s is populated but no"
-              + " channel carries an engine context — the call context was lost on the way to"
-              + " this thread (correlation_id=%s)",
-          engineKind, MDC.get("correlation_id"));
-    }
   }
 }
--- a/service/src/main/java/ai/floedb/floecat/service/context/impl/ResolvedCallContexts.java
+++ b/service/src/main/java/ai/floedb/floecat/service/context/impl/ResolvedCallContexts.java
@@ -24,6 +24,7 @@ import io.vertx.core.Vertx;
 import java.util.Objects;
 import java.util.concurrent.Callable;
 import org.jboss.logging.Logger;
+import org.jboss.logging.MDC;
 
 /**
  * Write-once carrier for the fully resolved inbound call context (principal, correlation id, engine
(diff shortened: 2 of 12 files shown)


service/src/main/java/ai/floedb/floecat/service/security/impl/Authorizer.java
```diff
@@ -38,8 +36,21 @@ public void require(PrincipalContext principalContext, List<String> permissions)
...
+  }
+
+  /**
+   * A principal with no subject and no permissions is not a caller that lacks a grant — it is a
+   * caller whose identity never arrived (interceptor bypassed, or the call context lost before
+   * authorization; eng-floe/floecat#361). Reporting that as UNAUTHENTICATED instead of
+   * PERMISSION_DENIED keeps the next propagation bug diagnosable.
+   */
+  private static RuntimeException denied(PrincipalContext principalContext, String detail) {
+    if (principalContext.getSubject().isBlank() && principalContext.getPermissionsCount() == 0) {
```
jchauzi-floe: Non-blocking. This infers "identity never arrived" from principal *shape* (blank subject + zero permissions), not an explicit signal from wherever context actually gets lost. It's safe today: OIDC throws before returning a blank subject, dev mode always sets permissions, and unauthenticated-allowed endpoints never reach `require()`. But that's a property of current callers, not something this code enforces.

If a future auth mode ever produces a legitimately-denied (not lost) principal with a blank subject and no permissions, it'll silently get `UNAUTHENTICATED` instead of `PERMISSION_DENIED` here, and nothing will catch the drift.

service/src/main/java/ai/floedb/floecat/service/context/impl/ResolvedCallContexts.java
```diff
@@ -104,7 +105,28 @@ public static ResolvedCallContext currentOrNull() {
...
+    if (fromKeys == null) {
+      warnIfMdcDisagrees();
+    }
+    return fromKeys;
+  }
+
+  /**
+   * Channel-disagreement detector: MDC carrying a correlation id proves this thread is inside a
+   * resolved request, so a miss on every context channel is a propagation loss — the failure mode
+   * of eng-floe/floecat#361, turned from a silent wrong answer into a countable signal. This should
```
jchauzi-floe: Non-blocking. This and `EngineContextProvider.warnIfRequestDeclaredEngine` are the same shape, copy-pasted:

```java
Object v = MDC.get(key);
if (v instanceof String s && !s.isBlank()) {
  LOG.warnf("... disagree ...", s, ...);
}
```

Given this PR's own thesis ("every channel needs a guard"), a third detector is plausible soon, and it'll likely copy-paste a third time. Worth pulling out a shared `warnOnChannelDisagreement(mdcKey, template)` before that happens.

service/src/main/java/ai/floedb/floecat/service/context/EngineContextProvider.java
```diff
@@ -69,6 +73,27 @@ public EngineContext engineContext() {
...
+    return fallback;
+  }
+
+  /**
+   * Channel-disagreement detector: MDC carrying an engine kind proves the request declared an
+   * engine, so resolving an empty engine context here is a propagation loss — the exact condition
+   * that silently un-resolves engine-gated system objects (eng-floe/floecat#361). This should never
+   * fire; any occurrence is a regression in context propagation.
+   */
+  private static void warnIfRequestDeclaredEngine() {
```
jchauzi-floe: Non-blocking. Same pattern as `ResolvedCallContexts.warnIfMdcDisagrees` — see that comment.

jchauzi-floe:
Reviewed the diff (8 angles: correctness x3, cleanup x3, altitude, conventions — no repo CLAUDE.md exists, so nothing there).

Verified the core claim — `Prioritized` vs. the silently-ignored `@Priority` — by decompiling the actual `quarkus-grpc-3.31.4` / `grpc-api-1.81.0` classes: `Interceptors$1`'s comparator only checks `instanceof Prioritized`, and `ServerInterceptors`' wiring makes the last sorted element outermost. `INBOUND_CONTEXT = 1_000_000` sorting last, and Quarkus's built-ins near `Integer.MAX_VALUE` staying outermost of all, checks out. Also traced every `Authorizer.require` call site and every test constructing a `PrincipalContext` — none currently hit the blank-subject path, so no test flips from `PERMISSION_DENIED` to `UNAUTHENTICATED`.

No correctness issues found. Left 3 non-blocking comments inline. One more, not inline since the file isn't touched by this diff so GitHub won't anchor a line comment there:

`service/src/test/java/ai/floedb/floecat/service/context/impl/TestEngineVersionCaptureInterceptor.java:36` still declares `@Priority(1000)` instead of implementing `Prioritized` — the exact defect class this PR fixes for the four production interceptors. Harmless today (it reads `x-engine-version`/`x-correlation-id` straight off `Metadata`, no dependency on interceptor ordering), but it's a live instance of the gap: if this one ever becomes order-sensitive, it silently ties at priority 0 with no compiler or test signal.

None of the above should hold up merging.

jchauzi-floe (approved):

Issue #220, opened 2026-03-16
[Feature]: Develop Discovery API for exploring, filtering, and selecting connector assets (namespaces/tables).


kwesterfeld-floe:
## Problem Statement

he Floecat API must implement a discovery endpoint to allow users to explore a connected catalog's assets during connector configuration. The API should provide available catalogs and namespaces, with support for filtering at both the namespace and table level. The exploration data returned must include key metrics such as table counts, file counts, and file sizes, if possible. This API will support the Floe UI functionality for selecting specific assets for reconciliation and verifying the connection scope and role privileges.

## Proposed Solution

New API methods to the existing connectors service.

## Alternatives Considered

None

## Scope and Impact

- Affected component(s):connectors service
- Backward compatibility considerations: new methods should not impact existing methods on this service.  New grpc message structs should be added that do not affect current APIs.
- Operational/performance considerations: wherever possible, APIs should favor subsecond passthrough APIs vs. aggregation of numerous round trip APIs to catalog endpoints, to facilitate better user experience at the consumer side of these APIs (ie. Floe UI)

## Checklist

- [x] I have searched existing issues and discussions for similar requests.
- [x] This request is aligned with project goals and architecture.
- [x] I am willing to contribute an implementation (optional).

PR #374, opened 2026-07-07
Connector Discovery service

arivarola-floe:
## Summary

Initial implementation of the Discovery API (#220), introducing endpoints to retrieve catalogs, namespaces, tables, and views from a connector. These endpoints will be used by the web application during the connector creation flow to discover and present available data sources.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist

## Additional Notes

For now, any errors during discovery will be returned as a string, this will be improved in a follow-up PR, as well as the pagination/filtering capabilities, depending on the UI requirements.


core/connectors/spi/src/main/java/ai/floedb/floecat/connector/spi/FloecatConnector.java
```diff
@@ -100,6 +100,9 @@ enum SnapshotSelectionKind {
 
   ConnectorFormat format();
 
+  List<String> listCatalogs();
```
mrouvroy-floe: [major] **SPI compatibility break**

Adding `listCatalogs()` as an abstract method to `FloecatConnector` breaks every external connector implementation compiled against the documented SPI. The PR marks no breaking changes, but a third-party connector jar will fail linkage/source compilation until it implements the new method.

The discovery API can still be introduced without forcing all connector authors to update immediately. Make this a `default` method, for example `default List<String> listCatalogs() { return List.of(); }`, or introduce a separate optional capability interface for connectors that can enumerate catalogs.

---

🤖 _Generated with GPT-5.5._

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsDiscoveryImpl.java
```diff
@@ -0,0 +1,377 @@
...
+
+                  var nsSegments = request.getNamespaceSegmentsList();
+                  String catalogName =
+                      request.hasCatalogName() ? request.getCatalogName().trim() : "";
+                  if (nsSegments.isEmpty() && catalogName.isEmpty()) {
+                    throw GrpcErrors.invalidArgument(
+                        corr, null, Map.of("field", "namespace_segments"));
+                  }
+                  String namespaceFq =
+                      nsSegments.isEmpty() ? catalogName : String.join(".", nsSegments);
```
mrouvroy-floe: [major] **Dropped catalog qualifier**

`DiscoverObjectsRequest` has both `catalog_name` and `namespace_segments`, but this branch ignores `catalog_name` whenever any namespace segment is present. A client using the API hierarchically will naturally call `DiscoverObjects` with `catalog_name = "prod_catalog"` and `namespace_segments = ["orders"]`; this code sends only `orders` to `connector.listTables()`/`listViews()`.

Connectors such as Unity Catalog and Iceberg REST expect the fully qualified namespace (`prod_catalog.orders`), so that request returns the wrong objects or an empty result even though the request carries the catalog.

Build `namespaceFq` from both fields when both are present, or normalize `DiscoverNamespaces` so `namespace_segments` always excludes the catalog and then prefix `catalog_name` here. Also populate `catalog_name` on the returned table/view `DiscoveryObject`s so clients can round-trip the selection.

---

🤖 _Generated with GPT-5.5._

mcusack-floe:
@arivarola-floe run `make fmt` before committing/pushing to avoid the ci-core action failure here.

core/connectors/spi/src/main/java/ai/floedb/floecat/connector/spi/FloecatConnector.java
arivarola-floe: Unless I'm missing something, I don't think this is a problem since we don't have any third-party connector jar

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsDiscoveryImpl.java
```diff
@@ -0,0 +1,393 @@
...
+                  String catalogFilter =
+                      request.hasCatalogName() ? request.getCatalogName().trim() : "";
+                  String query = request.getQuery().trim().toLowerCase(java.util.Locale.ROOT);
+
+                  try (var connector = ConnectorFactory.create(cfg)) {
+                    var namespaces = connector.listNamespaces();
+
+                    var refs =
+                        namespaces.stream()
+                            .filter(
```
mrouvroy-floe: [major] **Unity-style `catalog.schema` naming assumed throughout discovery — breaks Iceberg and Glue connectors**

`ConnectorsDiscoveryImpl` assumes every connector returns namespaces as `catalog.schema` strings (Unity style) in all three discovery methods, but Iceberg REST/Glue connectors do not conform to this contract.

**`discoverNamespaces` (lines 141–156):** The catalog filter retains only namespaces that equal `catalogFilter` or start with `catalogFilter + "."`. Iceberg `listNamespaces()` returns bare strings such as `orders`, not `rest_catalog.orders`, so every namespace is silently filtered out and the caller receives an empty list even when namespaces exist. Simultaneously, `segments.get(0)` mislabels the first dot-segment as the catalog name, so a multi-level namespace `sales.orders` is split into catalog=`sales` / namespace=`orders` rather than reflecting the real Iceberg catalog.

**`discoverObjects` (line 208):** `catalog_name` is prepended to `namespace_segments` before calling `connector.listTables(namespaceFq)` and `connector.listViews(namespaceFq)`. Iceberg REST expects a bare namespace (`orders`); passing `rest_catalog.orders` makes REST query the wrong namespace and Glue query a non-existent database, so object discovery returns no tables after a user selects a catalog.

Fix by defining a clear SPI contract for `FloecatConnector`: either all implementations return catalog-qualified namespace strings, or the service treats the catalog as opaque metadata and does not embed it in namespace strings passed back to connectors. Update `discoverNamespaces` filter logic and `discoverObjects` argument construction to match whichever contract is chosen.

---

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsDiscoveryImpl.java
```diff
@@ -0,0 +1,393 @@
...
+import org.jboss.logging.Logger;
+
+// Note: the connector SPI (listCataogs, listNamespaces, listTables, listViews) always returns a
+// full list with no cursor support, so pagination cannot be pushed to the source. Page parameters
+// in these RPCs are ignored and all results are returned in a single response.
+
+@GrpcService
+public class ConnectorsDiscoveryImpl extends BaseServiceImpl implements ConnectorDiscovery {
+  @Inject ConnectorRepository connectorRepo;
+  @Inject CatalogRepository catalogRepo;
```
mrouvroy-floe: [minor] **Unused injected repositories**

`catalogRepo`, `namespaceRepo`, `tableRepo`, and `idempotencyStore` are injected via CDI but never referenced anywhere in `ConnectorsDiscoveryImpl`.

Unused injections add container startup cost and mislead readers into thinking discovery reads from the local catalog store — it does not; all data flows through live connector calls.

Remove the four fields until they are actually needed.

---

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsDiscoveryImpl.java
```diff
@@ -0,0 +1,393 @@
...
+
+  private ConnectorConfig resolveCredentials(ConnectorConfig base, AuthConfig auth) {
+    if (hasAuthCredentials(auth)) {
+      AuthResolutionContext context = AuthResolutionContexts.fromInboundContext();
+      return CredentialResolverSupport.apply(base, auth.getCredentials(), context);
+    }
+    return base;
+  }
+
+  private static DiscoveryStatus buildErrorStatus(Exception e) {
```
mrouvroy-floe, line 269: [minor] **`buildErrorStatus` exposes raw exception chain to callers**

`buildErrorStatus` walks the full `Throwable` cause chain and concatenates each exception's class name and message into `DiscoveryStatus.summary`, which is returned to the API caller.

This can surface internal implementation details — driver/SDK class names, connection URIs, and occasionally credential-adjacent messages — to external clients.

Before this reaches an external UI, ensure the summary is scrubbed of sensitive content. At minimum, log the full chain server-side and expose only a sanitized, user-friendly message in the response.

---

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsDiscoveryImpl.java
```diff
@@ -0,0 +1,393 @@
...
+    throw GrpcErrors.invalidArgument(corr, null, Map.of("field", "target"));
+  }
+
+  private ConnectorConfig buildConnectorConfigFromSpec(ConnectorSpec spec, String corr) {
+    var kind = resolveKind(spec.getKind(), corr);
+    var auth = toConnectorAuth(spec.getAuth());
+    var cfg =
+        new ConnectorConfig(
+            kind,
+            spec.getDisplayName() != null ? spec.getDisplayName() : "",
```
mrouvroy-floe: [nit] **Dead null-check on proto3 string getter**

`spec.getDisplayName() != null ? spec.getDisplayName() : ""` — generated proto3 string accessors never return `null`; an unset field returns `""`. The ternary is dead code and can be simplified to `spec.getDisplayName()`.

---

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsDiscoveryImpl.java
arivarola-floe: I removed discoverCatalogs and the catalog_name filters from DiscoverNamespaces/DiscoverObjects entirely, so namespaces are the only thing being discovered now, and the catalog (if any) is just the first segment. This matches how listing already works today.
We can revisit later when the Discovery API is expanded to include more information about each object type

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsDiscoveryImpl.java
arivarola-floe: This is intended for now, will expand on this on a follow-up PR

connectors/catalogs/delta/src/main/java/ai/floedb/floecat/connector/delta/uc/impl/GlueDeltaCatalog.java
```diff
@@ -34,6 +35,25 @@ final class GlueDeltaCatalog implements AutoCloseable {
     this.glue = glue;
   }
 
+  List<String> deltaCatalogs() {
```
mrouvroy-floe: [minor] **`deltaCatalogs()` is dead code and misuses `databaseHasDelta`**

`deltaCatalogs()` has no callers anywhere in the repo — a repo-wide search finds only this definition. It appears to be a straggler from the removed catalog-discovery methods (see commit `c817cb02` "Remove list/discover catalogs methods" and the commented-out `catalog_name` fields in `discovery.proto`).

Beyond being unused, it's also incorrect: it iterates Glue *catalog* names from `getCatalogs()` and passes each into `databaseHasDelta(name)`, which issues a `GetTablesRequest` with `databaseName(<catalog>)`. A catalog is not a database, so this check would not do what the name implies if it were ever wired up.

Remove the method (and the now-unused `GetCatalogsRequest` import) so the dead code doesn't get resurrected with the latent bug.

---

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsDiscoveryImpl.java
```diff
@@ -0,0 +1,305 @@
...
+    ensureKind(connectorId, ResourceKind.RK_CONNECTOR, "connector_id", corr);
+    return connectorId.toBuilder().setAccountId(accountId).build();
+  }
+
+  private static ConnectorConfig.Auth toConnectorAuth(AuthConfig auth) {
+    return new ConnectorConfig.Auth(
+        auth.getScheme(), auth.getPropertiesMap(), auth.getHeaderHintsMap());
+  }
+
+  private static boolean hasAuthCredentials(AuthConfig auth) {
```
mrouvroy-floe: [minor] **Auth/config helpers duplicated from `ConnectorsImpl`**

`hasAuthCredentials`, `resolveCredentials`, `toConnectorAuth`, and the `ConnectorKind` -> `Kind` switch are copy-pasted verbatim from `ConnectorsImpl` (`hasAuthCredentials` at `ConnectorsImpl.java:769`, `resolveCredentials` at `:972`, `toConnectorAuth` at `:980`, the kind switch at `:696`).

This matters because credential handling is security-sensitive: if the logic in one impl is fixed or hardened later, the copy here can silently drift out of sync.

Consider extracting these into a shared helper (e.g. a `ConnectorConfigSupport` used by both services) rather than maintaining two copies.

---

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsDiscoveryImpl.java
```diff
@@ -0,0 +1,270 @@
...
+            auth);
+    return ConnectorConfigSupport.resolveCredentials(cfg, authForResolve);
+  }
+
+  private ResourceId scopedConnectorId(String accountId, ResourceId connectorId, String corr) {
+    ensureKind(connectorId, ResourceKind.RK_CONNECTOR, "connector_id", corr);
+    return connectorId.toBuilder().setAccountId(accountId).build();
+  }
+
+  private static DiscoveryStatus buildErrorStatus(Exception e) {
```
mrouvroy-floe, line 269: [major] **Raw exception chain exposed in `buildErrorStatus` response**

`buildErrorStatus` walks the full `Throwable` cause chain and concatenates each exception's class name and message into `DiscoveryStatus.summary`, which is returned verbatim to the API caller.

This is not merely an implementation-detail leak. Existing connector code can put upstream response bodies and full request URLs into exception messages during discovery — for example, Unity Catalog list failures include the URL and `resp.body()`. Those messages can contain host paths, object names, service-specific diagnostics, and credential-adjacent text that should stay server-side. The endpoint feeds the web connector-creation UI, making this a user-facing exposure surface.

Log the full exception chain server-side and return only a sanitized, user-friendly message in the response. If the UI needs structured diagnostics, pass only allowlisted fields or explicitly redacted messages rather than raw exception text.

---

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsDiscoveryImpl.java
```diff
@@ -0,0 +1,270 @@
...
+import io.smallrye.mutiny.Uni;
+import jakarta.inject.Inject;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.jboss.logging.Logger;
+
+// Note: the connector SPI (listNamespaces, listTables, listViews) always returns a
```
mrouvroy-floe, line 53: [minor] **Pagination ignored — unbounded response with no client-visible cap**

Both RPCs accept a `PageRequest` and return a `PageResponse`, but the page parameter is ignored and the response `page` field is never populated — every match is returned in a single response.

For a connector pointed at a catalog with thousands of tables or views, this materializes the entire list into one gRPC message, which can be a latency and memory concern and gives clients no way to bound the result set. Message size may also exceed gRPC limits.

At minimum, apply an application-level cap: truncate results and signal truncation via `PageResponse` or `DiscoveryStatus` so callers know the list is incomplete. Full cursor-based pagination can follow once the SPI supports it.

---

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsDiscoveryImpl.java
```diff
@@ -0,0 +1,270 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ai.floedb.floecat.service.connector.impl;
+
+import ai.floedb.floecat.common.rpc.ResourceId;
+import ai.floedb.floecat.common.rpc.ResourceKind;
+import ai.floedb.floecat.connector.rpc.*;
```
mrouvroy-floe: [nit] **Wildcard import diverges from production convention**

`import ai.floedb.floecat.connector.rpc.*;` is the only wildcard import in this area of production code — every other production class (e.g. `ConnectorsImpl`) lists `...connector.rpc` types explicitly, one per import.

Wildcard imports obscure which generated types a class actually depends on and can mask future name collisions as the proto package grows. Expand to explicit imports to match the surrounding codebase style.

---

PR #375, opened 2026-07-07
Support multiple source/destination mappings per connector

arivarola-floe:
## Summary

Update the connector configuration to support a list of source/destination pairs, enabling multiple namespace/table mappings to be configured from a single connector, as required by the web application. The existing source and destination fields are retained for backward compatibility with existing tools.

This change also allows connectors to be created without an initial source or destination. In this case, the connector is created in the CS_CREATING state and should not be reconciled until its configuration is complete.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [ ] No breaking changes

## Checklist

### DIFF
--- a/core/proto/src/main/proto/floecat/connector/connector.proto
+++ b/core/proto/src/main/proto/floecat/connector/connector.proto
@@ -68,8 +68,10 @@ message Connector {
   string display_name = 2;
   optional string description = 3;
   ConnectorKind kind = 4;
-  SourceSelector source = 10;
-  DestinationTarget destination = 11;
+  // Legacy storage only — normalized into mappings on read; never written.
+  SourceSelector source = 10 [deprecated = true];
+  // Legacy storage only — normalized into mappings on read; never written.
+  DestinationTarget destination = 11 [deprecated = true];
   repeated SourceMapping mappings = 12;
   string uri = 20;
   AuthConfig auth = 22;
@@ -198,8 +200,8 @@ message ConnectorSpec {
   optional string display_name = 1;
   optional string description = 2;
   ConnectorKind kind = 3;
-  SourceSelector source = 10;
-  DestinationTarget destination = 11;
+  reserved 10, 11;
+  reserved "source", "destination";
   repeated SourceMapping mappings = 12;
   optional string uri = 20;
   AuthConfig auth = 21;
--- a/core/proto/src/main/proto/floecat/connector/connector.proto
+++ b/core/proto/src/main/proto/floecat/connector/connector.proto
@@ -58,6 +58,11 @@ message DestinationTarget {
   }
 }
 
+message SourceMapping {
+  SourceSelector source = 1;
+  DestinationTarget destination = 2;
+}
+
 message Connector {
   ai.floedb.floecat.common.ResourceId resource_id = 1;
   string display_name = 2;
@@ -65,6 +70,7 @@ message Connector {
   ConnectorKind kind = 4;
   SourceSelector source = 10;
   DestinationTarget destination = 11;
+  repeated SourceMapping mappings = 12;
   string uri = 20;
   AuthConfig auth = 22;
   ReconcilePolicy policy = 30;
@@ -76,9 +82,10 @@ message Connector {
 
 enum ConnectorState {
   CS_UNSPECIFIED = 0;
-  CS_ACTIVE = 1;
-  CS_PAUSED = 2;
-  CS_DELETING = 3;
+  CS_CREATING = 1;
+  CS_ACTIVE = 2;
+  CS_PAUSED = 3;
+  CS_DELETING = 4;
 }
 
 message AuthConfig {
@@ -193,6 +200,7 @@ message ConnectorSpec {
   ConnectorKind kind = 3;
   SourceSelector source = 10;
   DestinationTarget destination = 11;
+  repeated SourceMapping mappings = 12;
   optional string uri = 20;
   AuthConfig auth = 21;
   ReconcilePolicy policy = 30;
--- a/core/proto/src/main/proto/floecat/connector/connector.proto
+++ b/core/proto/src/main/proto/floecat/connector/connector.proto
@@ -82,10 +82,10 @@ message Connector {
 
 enum ConnectorState {
   CS_UNSPECIFIED = 0;
-  CS_CREATING = 1;
-  CS_ACTIVE = 2;
-  CS_PAUSED = 3;
-  CS_DELETING = 4;
+  CS_ACTIVE = 1;
+  CS_PAUSED = 2;
+  CS_DELETING = 3;
+  CS_CREATING = 4;
 }
 
 message AuthConfig {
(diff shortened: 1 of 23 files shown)


core/proto/src/main/proto/floecat/connector/connector.proto
```diff
@@ -76,9 +82,10 @@ message Connector {
 
 enum ConnectorState {
   CS_UNSPECIFIED = 0;
-  CS_ACTIVE = 1;
-  CS_PAUSED = 2;
-  CS_DELETING = 3;
+  CS_CREATING = 1;
```
mrouvroy-floe: [blocker] **`ConnectorState` values renumbered**

This inserts `CS_CREATING` at numeric value `1` and shifts the existing values for `CS_ACTIVE`, `CS_PAUSED`, and `CS_DELETING`.

`ConnectorRepository` persists `Connector` messages as protobuf bytes via `Connector::toByteArray` and reads them with `Connector::parseFrom`. Existing stored active connectors therefore have `state = 1`; after this schema change they decode as `CS_CREATING`, and the scheduler/reconciler paths that require `CS_ACTIVE` will skip or reject them.

Keep the existing enum numbers stable and add the new state at the next unused value, e.g. leave `CS_ACTIVE = 1`, `CS_PAUSED = 2`, `CS_DELETING = 3`, and assign `CS_CREATING = 4`.

---

🤖 _Generated with GPT-5.5._

mrouvroy-floe (commented):
**[major] `mappings[].source.columns` ignored** — `reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReconcilerService.java:327`

Table planning now iterates `effectiveMappings(active)`, but snapshot planning still derives `defaultColumnSelectors` from `active.source().getColumnsList()`.

For connectors created with more than one mapping, `ConnectorsImpl` stores the mapping list but does not populate the legacy top-level `source`, so any per-mapping `source.columns` selectors are lost once the planned `ReconcileTableTask` reaches snapshot planning and execution. That changes capture completeness checks and capture execution from the configured column subset to the default/all-column behavior.

Carry the effective mapping's column selectors through the planned table task, or resolve them from the connector mapping that matches the task's source/destination before computing `defaultColumnSelectors`. The queued execution paths that use `active.source().getColumnsList()` need the same treatment.

---

🤖 _Generated with GPT-5.5._

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/QueuedReconcileWorkerSupport.java
```diff
@@ -826,7 +829,11 @@ private Result reconcileDiscoveryTableTask(
         reconcilerService.connectorOpener.open(active.resolvedConfig())) {
       ResourceId destCatalogId = active.destination().getCatalogId();
```
mrouvroy-floe: [major] **Discovery execution uses legacy `destination` instead of mapping-aware destination for multi-mapping connectors**

**Problem:** `executeDiscoveryTable` derives `destCatalogId` from `active.destination().getCatalogId()` and passes `active.destination()` to `resolveOrCreateDiscoveryNamespaceId`. For connectors with more than one mapping, `ConnectorsImpl` only populates the legacy top-level `destination` when there is exactly one mapping (see `ConnectorsImpl.java:304-313`); for genuine multi-mapping connectors `active.destination()` is the default instance and `destCatalogId` is an empty `ResourceId`.

**Impact:** Table planning was correctly made mapping-aware (`effectiveColumnSelectors`, per-mapping destination namespace on the task), but execution is blind to it. Table lookup, `ensureTable`, and metadata updates run against the wrong catalog or fail outright, even though the planned `ReconcileTableTask` carries valid per-mapping destination data.

**Fix:** Either carry the mapping destination catalog ID through the planned `ReconcileTableTask` and read it here, or resolve the effective `SourceMapping` for the task at execution time and use that mapping's `DestinationCatalog` instead of the legacy `active.destination()`.

---

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
```diff
@@ -1211,6 +1252,19 @@ private Connector applyConnectorSpecPatch(
       b.setDestination(db.build());
     }
 
+    if (maskTargets(mask, "mappings")) {
```
mrouvroy-floe, line 1101: [major] **Update path does not clear stale legacy `source`/`destination` when mappings change to multiple entries**

**Problem:** Create enforces mutual exclusion between the legacy `source`/`destination` fields and `mappings`, and only mirrors a single mapping into the legacy fields. The update path does neither: when a single-mapping connector is patched to multiple `mappings`, the branch replaces the repeated `mappings` list but only rewrites the legacy fields when the new list has exactly one item — leaving the old single-mapping `source`/`destination` persisted alongside a multi-entry `mappings` list. Additionally, a client can patch both `source` (or `destination`) and `mappings` in the same mask with no post-check enforcing exclusivity.

**Impact:** API callers receive misleading legacy fields. Any legacy reconciler path that reads `active.source()` or `active.destination()` will use the previous single-mapping values rather than the current configuration. The stale fields also leak into the connector fingerprint and diverge from create-time invariants.

**Fix:** Mirror create behavior in the update path: after replacing `mappings`, set legacy `source`/`destination` only when the new list has exactly one entry and explicitly clear (zero/null) them for zero or multiple mappings. Add a post-patch check that rejects a mask specifying both legacy fields and `mappings`.

---

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
```diff
@@ -250,117 +252,39 @@ public Uni<CreateConnectorResponse> createConnector(CreateConnectorRequest reque
...
+                  boolean hasLegacySourceOrDest = spec.hasSource() || spec.hasDestination();
+                  var mappingsIn = spec.getMappingsList();
+                  boolean hasMappings = !mappingsIn.isEmpty();
+
+                  if (hasLegacySourceOrDest && hasMappings) {
                     throw GrpcErrors.invalidArgument(
-                        corr,
-                        GeneratedErrorMessages.MessageKey.CONNECTOR_MISSING_DESTINATION_CATALOG,
-                        Map.of("id", "destination.catalog_id|catalog_display_name"));
+                        corr, null, Map.of("field", "source|destination|mappings"));
```
mrouvroy-floe: [minor] **New validation errors pass `null` `MessageKey`, producing client-facing errors with no descriptive message-key suffix**

**Problem:** Several new argument validations — the mappings/legacy exclusivity error here, the `policy.enabled` restriction at lines 271 and 1384, and the `mappings[i].source.table` error in `resolveMapping` — pass `null` as the `MessageKey` argument to `GrpcErrors.invalidArgument`.

**Impact:** `GrpcErrors` tolerates a `null` key without throwing, but every other validation in this file supplies a `GeneratedErrorMessages.MessageKey`. These new errors reach the client with only a `field` param and no descriptive/localized message-key suffix, breaking consistency with the rest of connector validation and making observability harder.

**Fix:** Define dedicated `MessageKey` constants for each of these new error cases and pass them to `GrpcErrors.invalidArgument` in the same way existing validations do.

---

PR #376, opened 2026-07-08
FM3: Add per-account snapshot sequence number

mcusack-floe:
## Summary

Adds an atomically updated sequence number that increments each time a snapshot is created in an account.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/repo/impl/SnapshotRepository.java
+++ b/service/src/main/java/ai/floedb/floecat/service/repo/impl/SnapshotRepository.java
@@ -19,8 +19,10 @@ package ai.floedb.floecat.service.repo.impl;
 import ai.floedb.floecat.catalog.rpc.CurrentSnapshotPointer;
 import ai.floedb.floecat.catalog.rpc.Snapshot;
 import ai.floedb.floecat.common.rpc.MutationMeta;
+import ai.floedb.floecat.common.rpc.Pointer;
 import ai.floedb.floecat.common.rpc.ResourceId;
 import ai.floedb.floecat.service.repo.model.Keys;
+import ai.floedb.floecat.service.repo.model.PointerReferences;
 import ai.floedb.floecat.service.repo.model.Schemas;
 import ai.floedb.floecat.service.repo.model.SnapshotKey;
 import ai.floedb.floecat.service.repo.util.BaseResourceRepository;
@@ -28,11 +30,14 @@ import ai.floedb.floecat.service.repo.util.GenericResourceRepository;
 import ai.floedb.floecat.storage.spi.BlobStore;
 import ai.floedb.floecat.storage.spi.PointerStore;
 import ai.floedb.floecat.telemetry.StoreOperationSummary;
+import ai.floedb.floecat.types.Hashing;
 import com.google.protobuf.Timestamp;
 import com.google.protobuf.util.Timestamps;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import java.time.Clock;
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.ThreadLocalRandom;
@@ -51,6 +56,8 @@ public class SnapshotRepository {
   }
 
   private final GenericResourceRepository<Snapshot, SnapshotKey> repo;
+  private final PointerStore pointerStore;
+  private final SnapshotCreateSequenceStore createSequences;
   private final TableRepository tableRepo;
   private final CurrentSnapshotPointerRepository currentPointerRepo;
   private final Clock clock;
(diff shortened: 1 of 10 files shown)


service/src/main/java/ai/floedb/floecat/service/repo/impl/SnapshotRepository.java
```diff
@@ -387,4 +436,66 @@ private Optional<Snapshot> latestSnapshotByTime(ResourceId tableId) {
...
+    keys.add(
+        Keys.snapshotPointerByTime(
+            snapshot.getTableId().getAccountId(),
+            snapshot.getTableId().getId(),
+            snapshot.getSnapshotId(),
+            upstreamCreatedAtMs));
+    return List.copyOf(keys);
+  }
+
+  private void classifySnapshotCreateConflict(String blobUri, List<String> pointerKeys) {
```
mrouvroy-floe, line 471: [major] **Reimplemented `create` drops the `PARTIAL_STATE` metric/log on corruption**

`create` no longer delegates to `GenericResourceRepository.create`; instead it hand-rolls the batch write plus `classifySnapshotCreateConflict`, which is a near-verbatim copy of the generic `classifyCreateConflict`.

The generic path routes partial-state anomalies through `partialStateAnomaly(...)`, which logs an `errorf` and increments the `ServiceMetrics.Storage.PARTIAL_STATE` counter so the (rare, non-retryable) corruption is visible for out-of-band reconciliation. The new snapshot path throws a bare `CorruptionException` with no log and no metric — so a stored snapshot inconsistency now vanishes into a generic 500 with nothing to alert on. This is an observability regression specifically for snapshot creates, and the duplicated invariant logic (name-conflict vs partial-state vs transient-retryable classification) is now maintained in two places and will drift.

Consider extending `GenericResourceRepository.create`/`createIfAbsent` to accept a small list of extra CAS ops (the sequence upsert) so this path can reuse the audited implementation, or at minimum route the corruption case through the same `partialStateAnomaly` logging + counter.

---

mrouvroy-floe, line 471: [major] **Reimplemented `create` drops the `PARTIAL_STATE` metric/log on corruption**

`create` no longer delegates to `GenericResourceRepository.create`; instead it hand-rolls the batch write plus `classifySnapshotCreateConflict`, which is a near-verbatim copy of the generic `classifyCreateConflict`.

The generic path routes partial-state anomalies through `partialStateAnomaly(...)`, which logs an `errorf` and increments the `ServiceMetrics.Storage.PARTIAL_STATE` counter so the rare, non-retryable corruption is visible for out-of-band reconciliation. The new snapshot path throws a bare `CorruptionException` with no log and no metric — so a stored snapshot inconsistency now vanishes into a generic 500 with nothing to alert on.

Fix by calling `partialStateAnomaly(...)` (or an equivalent helper) from `classifySnapshotCreateConflict` before throwing, and consolidate the duplicated invariant logic (name-conflict vs partial-state vs transient) to avoid future drift.

---

service/src/main/java/ai/floedb/floecat/service/transaction/impl/TransactionIntentApplierSupport.java
```diff
@@ -312,6 +342,19 @@ public ApplyOutcome applyTransactionAtomically(
       }
     }
 
+    ApplyOutcome counterOutcome = appendSnapshotCreateCounterOps(intents, touchedKeys, ops);
```
mrouvroy-floe, line 345: [minor] **Snapshot sequence CAS op not counted during transaction prepare, causing spurious `POINTER_TXN_TOO_LARGE` at commit**

`applyTransactionAtomically` appends an extra `SnapshotCreateSequenceStore` CAS op for snapshot creates after `TransactionsServiceImpl.prepareTransaction` has already admitted the transaction using its fixed pointer-op estimate.

A transaction accepted at exactly the 100-op limit becomes 101 ops once the sequence update is added. Commit then returns `POINTER_TXN_TOO_LARGE` and marks the prepared transaction as a terminal apply conflict, even though the request should have been rejected before durable intents were created.

Fix by updating the prepare-time estimator to reserve one additional pointer op for each account whose transaction includes a new `snapshots/by-id` create, matching the logic in `appendSnapshotCreateSequenceOps`, so oversized snapshot transactions fail during prepare rather than at commit.

---

service/src/main/java/ai/floedb/floecat/service/repo/impl/SnapshotRepository.java
```diff
@@ -69,6 +76,8 @@ public SnapshotRepository(
             Snapshot::parseFrom,
             Snapshot::toByteArray,
             "application/x-protobuf");
+    this.pointerStore = pointerStore;
+    this.createCounters = new SnapshotCreateCounterStore(pointerStore);
```
mrouvroy-floe, line 80: [minor] **`SnapshotCreateSequenceStore` is a CDI bean but constructed with `new`**

`SnapshotCreateSequenceStore` is annotated `@ApplicationScoped`, yet `SnapshotRepository` constructs its own instance with `new SnapshotCreateSequenceStore(pointerStore)` in both constructors, while `TransactionIntentApplierSupport` `@Inject`s it (and further falls back to `new` when the field is null).

The store is stateless so there is no correctness impact today, but the mixed wiring is confusing: the `@ApplicationScoped` annotation implies a managed singleton that is in fact bypassed here, and any future per-instance state (caching, metrics) would silently diverge across the three construction sites.

Pick one strategy: either inject the bean everywhere via CDI, or drop `@ApplicationScoped` and treat it as a plain helper constructed on demand.

---

PR #377, opened 2026-07-08
fix: Harden AWS client refresh and expose reconcile planning failures

mcusack-floe:
## Summary

This change protects Floecat AWS SDK call paths from stale closed connection pools by refreshing affected clients and retrying once when Connection pool shut down is detected.

Key changes:
- Added refresh/retry wrappers for Glue, STS, Delta S3, protocol-gateway S3, S3 blob storage, DynamoDB KV, DynamoDB reconcile job stores, Secrets Manager, and storage authority credential vending.
- Added low-impact warning logs only on closed-pool refresh paths, including module/function or operation context.
- Surfaced underlying remote planning failures in Floecat logs for PLAN_CONNECTOR and PLAN_TABLE execution paths.
- Ensured Delta connector S3 and common AWS wrapper refresh by failed-client identity to avoid replacing a newer client with a stale failed one.
- Tightened DynamoDB KV refresh handling so non-closed-pool async failures are not sent to the refresh handler.

## Type of Change

- [X] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist


reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteDefaultReconcileExecutor.java
```diff
@@ -175,8 +175,51 @@ private ExecutionResult executeTable(ExecutionContext context, RemoteLeasedJob r
...
+          retryClassOf(classified),
+          blankToEmpty(classified.getMessage()),
+          rootCauseMessage(classified));
+      workerClient.submitPlanTableFailure(
+          remoteLease,
+          failureKindOf(classified),
+          retryDispositionOf(classified),
+          retryClassOf(classified),
+          blankToEmpty(classified.getMessage()));
+      return ExecutionResult.failure(
```
mrouvroy-floe, line 234: [major] **`PLAN_TABLE` failure discards classified retry disposition/kind in returned `ExecutionResult`**

The new catch block computes `failureKindOf(classified)`, `retryDispositionOf(classified)`, and `retryClassOf(classified)` and forwards them to `workerClient.submitPlanTableFailure(...)`, but the `ExecutionResult` it returns uses the short `ExecutionResult.failure(...)` overload which hardcodes `FailureKind.INTERNAL`, `RetryDisposition.RETRYABLE`, and `RetryClass.TRANSIENT_ERROR`.

That returned result is authoritative for the local completion path: `RemoteReconcileExecutorPoller.completeForOutcome(...)` consumes `result.retryDisposition` / `result.retryClass` and `classify(result)` to submit job completion. The carefully computed classification is sent to the server but ignored locally, so the poller always treats `PLAN_TABLE` failures as transient/retryable regardless of the real classification.

Return an `ExecutionResult` constructed with the same `failureKind`, `retryDisposition`, and `retryClass` values that were forwarded to `submitPlanTableFailure`, so the local and remote completion paths agree.

---

core/connectors/common/src/main/java/ai/floedb/floecat/connector/common/aws/RefreshingAwsClient.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+    this.factory = factory;
+  }
+
+  public <R> R call(ThrowingOperation<T, R> operation) throws IOException {
+    for (int attempt = 0; ; attempt++) {
+      T client = current();
+      try {
+        return operation.call(client);
+      } catch (RuntimeException e) {
```
mrouvroy-floe, line 51: [major] **Checked `IOException`s bypass closed-pool detection and refresh**

`call` accepts a `ThrowingOperation` and is used around S3 range reads where `ResponseInputStream.read(...)` can throw `IOException`. The retry loop only catches `RuntimeException`, so checked `IOException`s escape without triggering closed-pool detection.

This leaves a stale, closed S3 client installed and surfaces the original read failure instead of refreshing and retrying once — the Delta and manifest S3 streaming paths remain vulnerable to the exact condition this PR aims to harden.

Handle `IOException` the same way as `RuntimeException`: check `isConnectionPoolShutdown(e)`, refresh and retry on a match, and rethrow non-matching `IOException`s unchanged. A focused test where a `ThrowingOperation` throws `new IOException("Connection pool shut down")` would lock this down.

---

storage/aws/src/main/java/ai/floedb/floecat/storage/kv/dynamodb/DynamoDbKvStore.java
```diff
@@ -213,15 +216,23 @@ private <T> Uni<T> dynamo(
           .onFailure()
           .recoverWithUni(
               failure -> {
-                clientFailureHandler.accept(client, failure);
                 if (shouldRefreshAndRetry(failure, attempt)) {
+                  LOG.warnf(
+                      failure,
+                      "DynamoDB async connection pool shutdown in %s; refreshing client",
```
mrouvroy-floe, line 224: [minor] **`callerContext()` on the async recover path likely resolves to `unknown`/Mutiny internals**

In the `recoverWithUni` callback the failure is delivered on the SDK/event-loop thread that completed the underlying `CompletableFuture`, not the thread that invoked the KV operation. `callerContext()` walks `Thread.currentThread().getStackTrace()`, so the real caller frame is not on the stack at that point — the log will show Mutiny/SDK internals or `unknown` instead of the actual operation site.

The synchronous `dynamoBlocking` path is unaffected; only the async `recoverWithUni` branch produces misleading attribution, undermining the diagnostic value of this log line.

Capture the caller context eagerly at the call site (before going async) and thread it into the `recoverWithUni` lambda, rather than deriving it inside the async callback.

---

storage/aws/src/main/java/ai/floedb/floecat/storage/kv/dynamodb/DynamoDbKvStore.java
```diff
@@ -596,14 +633,9 @@ void listTableIfExists(String header) {
 
       System.out.println("=== END TABLE DUMP ===\n");
 
-    } catch (ExecutionException e) {
-      clientFailureHandler.accept(client, e);
+    } catch (RuntimeException e) {
```
mrouvroy-floe, line 638: [minor] **`listTableIfExists` no longer restores interrupt status**

The rewrite folds `client.scan(req).get()` into `dynamoBlocking`, which wraps any checked exception — including `InterruptedException` — into a `RuntimeException`. That wrapped exception is then swallowed by the new `catch (RuntimeException e)` block as "Table not found", silently clearing the interrupt flag.

The previous code had a dedicated `catch (InterruptedException)` that called `Thread.currentThread().interrupt()`. Impact is low since this is a debug/dump helper (`System.out`), but silently swallowing interrupt status is a latent correctness issue if this code ever runs on a pooled or managed thread.

Inspect the wrapped cause: if it is an `InterruptedException`, call `Thread.currentThread().interrupt()` before proceeding.

---

PR #380, opened 2026-07-10
fix: Harden durable reconcile queue leasing and maintenance

mcusack-floe:
## Summary

This branch reworks the durable reconcile queue around more explicit indexes, maintenance loops, and remote completion handling:
- Adds multi-index ready queue pointers for global, execution class, execution lane, job kind, and pinned executor leasing, with shared ReadyQueueKeys logic to avoid duplicated key construction.
- Moves stale ready-pointer cleanup out of the hot lease scan and into lease maintenance, which now reclaims expired leases and prunes leaked/superseded ready entries.
- Adds dedicated maintenance services/schedulers for lease recovery, projection refresh, cancellation cleanup, and ready-index repair.
- Adds ready-index migration/repair for queued jobs, including correct write-item accounting for canonical updates plus ready upserts/deletes.
- Improves queue selection fairness by rotating through eligible pinned and unpinned ready slices instead of repeatedly scanning the same slice first.
- Tightens cancellation handling for queued/running trees, including cancellation cleanup markers, lease shortening on cancel, and parent rollup behavior for cancelling/cancelled children.
- Improves parent projection maintenance with dirty-parent markers, requested/applied generations, invalid marker cleanup, and obsolete marker deletion.
- Chunks remote planner child-job submissions for table and snapshot planning so large result sets do not exceed gRPC/message limits.
- Hardens remote lease completion behavior around precondition failures, heartbeat rejection, final lease validation, and handled completions to avoid duplicate terminal submissions.
- Updates DynamoDB/memory backends to support ready queue mutations, pointer deletes, full ready scans, and combined CAS batches needed by the new maintenance paths.
- Adds/updates coverage for durable queue behavior, maintenance services, ready queue budgets, remote executor completion, gRPC chunking, and lease/poller edge cases.

## Type of Change

- [X] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [ ] No breaking changes

## Checklist


reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/GrpcRemoteReconcileExecutorClient.java
```diff
@@ -1089,10 +1205,57 @@ private static List<SubmitLeasedSnapshotFinalizeResultRequest.Chunk> chunkSnapsh
...
+      out.add(List.copyOf(current));
+    }
+    return List.copyOf(out);
+  }
+
   private static int estimatedChunkItemBytes(MessageLite message) {
     return Math.max(1, message.getSerializedSize()) + 32;
   }
 
+  private static RuntimeException leasePreconditionOrOriginal(
```
mrouvroy-floe, line 1262: [major] **`leasePreconditionOrOriginal` swallows all `FAILED_PRECONDITION`, defeating server integrity guards**

`leasePreconditionOrOriginal` converts *every* `StatusRuntimeException` with code `FAILED_PRECONDITION` into a `RemoteLeasePreconditionFailedException`, and the executors treat that as a benign lost lease — e.g. `RemoteSnapshotPlanningReconcileExecutor` returns `ExecutionResult.cancelled(...)` ("Lease no longer valid"), and `RemoteDefaultReconcileExecutor` calls `leaseNoLongerValid(...)`.

But the server (`LeasedPlannerWorkerService`) uses `FAILED_PRECONDITION` for genuine data-integrity failures, not just lease loss. The clearest case is the file-group count check in `persistPlanSnapshotSuccess` (`LeasedPlannerWorkerService.java:523`): when the declared `file_group_count` disagrees with the number of staged file-group jobs it throws `Status.FAILED_PRECONDITION`. This guard exists specifically to fail loud — it is one of the "disagreement detectors" this PR advertises. Because the client rewrites it to a lease-invalid exception, a real inconsistency is silently downgraded to a job cancellation: the snapshot's file-group children are dropped and the plan is left quietly incomplete, with no error surfaced.

Note the table path is inconsistent with the snapshot path here: `requireNoStagedPlanTableChunkAt` throws `IllegalArgumentException` (→ `INVALID_ARGUMENT`, which is *not* swallowed), while the snapshot count mismatch uses `FAILED_PRECONDITION` (which *is* swallowed).

Fix: give the server a way to distinguish lease-invalidity from integrity failures — either a distinct status code / error detail for the true lease-precondition case, or only convert that specific case — so the integrity guards propagate as real errors instead of being masked as cancellations.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/projection/ReconcileJobProjector.java
```diff
@@ -467,11 +484,13 @@ public JobProjection projectExecFileGroup(ReconcileFileGroupTask fileGroupTask,
         failedFiles = plannedFiles;
       }
     }
+    long completedFileGroups =
```
mrouvroy-floe, line 487: [major] **Exec-file-group projection counts failed groups as both completed and failed**

`completedFileGroups` is now `"JS_SUCCEEDED".equals(state) || completedFiles > 0L || failedFiles > 0L ? 1L : 0L`, while `failedFileGroups` right below is `("JS_FAILED".equals(state) || "JS_CANCELLED".equals(state)) ? 1L : 0L`. These are no longer mutually exclusive.

For a `JS_FAILED`/`JS_CANCELLED` group with any file activity the single group is reported as `completedFileGroups = 1` **and** `failedFileGroups = 1`. This is the normal failure shape: lines 483-484 force `failedFiles = plannedFiles` for a failed group with no per-file results, so `failedFiles > 0` is essentially always true on failure.

`ReconcileAncestorRollupService` sums `completedFileGroups` and `failedFileGroups` independently up the tree, and both are surfaced to the API in `ReconcileControlImpl.fileGroupCounts` (`FileGroupsCompleted` / `FileGroupsFailed`). A one-group plan that fails then reports `total=1, completed=1, failed=1` — a failed group shown as completed, and inflated completed counts across any partially-failed reconcile.

Control flow (finalization eligibility) keys off rollup *state*, not these counts, so this is reporting corruption rather than a scheduling bug — but it reliably misreports every failure. Make the two counts disjoint (a group is completed only when it is not failed/cancelled).

```suggestion
    long completedFileGroups =
        !("JS_FAILED".equals(state) || "JS_CANCELLED".equals(state))
                && ("JS_SUCCEEDED".equals(state) || completedFiles > 0L || failedFiles > 0L)
            ? 1L
            : 0L;
```

---

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/queue/ReconcileReadyIndexMaintenanceService.java
```diff
@@ -0,0 +1,384 @@
...
+              readyWrites + flush.readyWrites(),
+              chunks + flush.chunks(),
+              failedChunks + flush.failedChunks());
+        }
+        scanned++;
+        ReadyIndexRepairWork work = buildRepairWork(record);
+        if (work == null) {
+          continue;
+        }
+        if (work.writeItems() > MAX_READY_REPAIR_WRITE_ITEMS) {
```
mrouvroy-floe, line 116: [minor] **Ready-index repair permanently skips oversized jobs and re-warns every tick**

When a single QUEUED job's `work.writeItems() > MAX_READY_REPAIR_WRITE_ITEMS` (100), it is logged and `continue`d without ever advancing its `readyIndexVersion`.

Because the maintenance scan selects jobs by `state = JS_QUEUED` with a stale `readyIndexVersion`, that same job re-qualifies on every subsequent tick, so it is re-scanned and re-warned indefinitely (until it leaves QUEUED) and its ready index is never repaired.

It is bounded per page (not a CPU spin), but it is a permanent no-op plus log spam for that job. Consider a backoff/marker or splitting the repair writes across ticks so an oversized job can still make progress.

---

mrouvroy-floe (commented):
**[minor] Cancellation commit can emit a single >100-item DynamoDB transaction** — `service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/DurableReconcileJobStore.java:2480`

The chunking guard in `commitCancellationMutations` is `if (!chunk.isEmpty() && chunkWriteItems + mutationWriteItems > CANCELLATION_BATCH_WRITE_ITEM_LIMIT)`.

If a single mutation's own write-item count exceeds 100, the `!chunk.isEmpty()` clause skips the split, the mutation is placed alone into a fresh chunk, and it is then committed as one `TransactWriteItems` with more than 100 items — which DynamoDB rejects permanently, so cancellation cleanup for that root can never complete.

The parallel repair path in `ReconcileReadyIndexMaintenanceService` explicitly guards this case (skips oversized jobs); the cancellation path has no equivalent guard. In practice a single job produces only a few pointer writes so this is an edge case, but the asymmetry looks like an oversight — worth an explicit guard/assert for the single-oversized-mutation case.

---

mrouvroy-floe (commented):
**[major] Connector/view planner submit methods do not convert lease-precondition errors, so executors take the wrong completion path** — `reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/GrpcRemoteReconcileExecutorClient.java:506`

`submitPlanConnectorSuccess`, `submitPlanConnectorFailure`, `submitPlanViewSuccess`, and `submitPlanViewFailure` call `invokeWorkerControlMutationOnce` directly without wrapping the result through `leasePreconditionOrOriginal(...)`. The table and snapshot equivalents were updated to perform this conversion, but these four paths were missed.

The service uses `ReconcileLeaseGrpcStatus.leasePreconditionFailed(...)` when `requireLeasedJob(...)` rejects a stale or shortened lease. Without the client-side conversion, the executors see a generic `StatusRuntimeException` instead of `RemoteLeasePreconditionFailedException` and do not enter the "lease no longer valid" handled-completion path. A cancelled or expired `PLAN_CONNECTOR`/`PLAN_VIEW` job is therefore logged and treated as an execution failure rather than a clean cancellation, and the duplicate-terminal-submission guard described in the snapshot executor finding is never reached.

Fix: wrap the return value of `invokeWorkerControlMutationOnce` through `leasePreconditionOrOriginal(...)` (or an equivalent re-throw helper) in all four connector/view submit methods, consistent with the table/snapshot paths.

**[minor] In-memory test store ignores the cancellation-blocking predicate, masking regressions in the cancellation lease path** — `service/src/test/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/inmemory/InMemoryReconcileReadyQueueStore.java:65`

`bind(...)` still accepts a `blockedByCancellation` predicate (line 65) but the scan loop no longer consults it — the field is unused. The production `NativeReconcileReadyQueueStore` enforces the check via `ReadyQueuePruneSupport.readyEntryPruneReason(...CANCELLATION_BLOCKED...)` (lines 203–209). The two stores are intended to be behaviour-equivalent for tests.

As written, tests exercising cancellation-blocked leasing against the in-memory store will lease a job that production would skip, masking regressions in the cancellation path.

Fix: apply the same `CANCELLATION_BLOCKED` short-circuit in the in-memory scan loop so the test double matches production behaviour.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotPlanningReconcileExecutor.java
```diff
@@ -221,19 +212,35 @@ public ExecutionResult execute(ExecutionContext context) {
...
+            remoteLease,
+            failureKindOf(classified),
+            retryDispositionOf(classified),
+            retryClassOf(classified),
+            failureDetail);
+      } catch (RemoteLeasePreconditionFailedException leaseRejected) {
+        LOG.infof(
+            "Snapshot planning failure submission ignored because reconcile lease is no longer valid jobId=%s tableId=%s snapshotId=%d",
+            lease.jobId, task.tableId(), task.snapshotId());
+        return ExecutionResult.cancelled(0, 0, 0, 0, 0, 0, 0, "Lease no longer valid");
```
mrouvroy-floe, line 244: [major] **Precondition-failed snapshot failure path skips beforeHandledCompletion, causing a duplicate terminal submission**

When `submitPlanSnapshotFailure(...)` throws `RemoteLeasePreconditionFailedException`, the catch block returns `ExecutionResult.cancelled(...)` without first invoking `context.beforeHandledCompletion().run()`. That callback is what sets `completionStarted` in the poller.

Without it, `RemoteReconcileExecutorPoller.runLease` reaches the guard `if (result.cancelled && completionStarted.get())` with `completionStarted == false`, falls through to `leaseStillCompletable(...)` (which returns `true` because `leaseInvalid` was never set and the wall-clock lease has not expired), and then calls `completeForOutcome(..., OBSOLETE, ...)` — submitting a second terminal `complete` against a lease whose precondition already failed server-side. This is the exact duplicate/invalid terminal submission the PR intends to prevent.

Fix: call `context.beforeHandledCompletion().run()` before returning `ExecutionResult.cancelled(...)` in the `RemoteLeasePreconditionFailedException` handler, mirroring the pattern used in the table-submit path.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/NativeReconcileReadyQueueStore.java
```diff
@@ -97,251 +94,318 @@ public boolean matchesLeaseRequest(StoredReconcileJob record, LeaseRequest reque
...
-        var leased =
-            leaseStore.leaseCanonical(
-                candidate.canonicalPointerKey(),
-                candidate.readyPointerKey(),
-                nowMs,
-                canonicalSnapshot,
-                record);
-        if (leased.isPresent()) {
-          return leased;
+        rememberScanPageCursor(slice, nextToken);
```
mrouvroy-floe: [major] **Shared per-slice scan cursor can permanently starve a job kind behind another's backlog**

`scanPageCursors` is keyed on `ReadyQueueSlice` (index type + filter value), not on the requesting worker, so every worker scanning the same slice shares a single cursor. With `MAX_PAGES_PER_LEASE_SELECTION = 1`, each lease call advances the cursor by one 16-entry page whenever the page yields no leasable-for-this-requester entry — including pages filled entirely with `request_mismatch` rows that are intentionally left in place (lines 217–219). The cursor resets to head on any successful lease (line 235).

This creates a cross-requester starvation window: if worker A (jobKind X) continuously leases jobs from the head pages of a lane slice, each success resets the shared cursor to head, so worker B (jobKind Y) — whose eligible jobs sit on a later page — never advances far enough to reach them. Under sustained head-page activity by one job kind, the other job kind starves indefinitely.

Fix: key `scanPageCursors` on `(ReadyQueueSlice, requesterId)` or equivalent, so each requester maintains an independent scan position through the slice.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedPlannerWorkerService.java
```diff
@@ -689,13 +765,184 @@ record PlannedViewJob(ReconcileScope scope, ReconcileViewTask viewTask) {}
 
   record PlannedSnapshotJob(ReconcileScope scope, ReconcileSnapshotTask snapshotTask) {}
 
+  private List<SubmitLeasedPlanTableResultRequest.Chunk> loadStagedPlanTableChunks(
```
mrouvroy-floe, line 768: [minor] **Over-declared chunk_count causes infinite retryable loop with no terminal escape**

`loadStagedPlanTableChunks` (and `loadStagedPlanSnapshotChunks`) iterate `0..chunkCount-1` and call `loadStagedPlanTableChunk`, which throws `StorageAbortRetryableException` when a declared chunk index has not yet been staged. The under-declare direction is guarded via `requireNoStagedPlanTableChunkAt` (throws `IllegalArgumentException`), but the over-declare direction — client sends `chunkCount = N` yet only staged `N-1` chunks due to a client bug or partial staging failure — produces a retryable abort forever. The missing chunk never arrives, the success RPC never completes, and the plan job never terminalises.

Fix: bound the number of retries for a missing declared chunk, or treat a chunk that is persistently absent after a threshold as a terminal `FAILED_PRECONDITION` so the job fails fast instead of spinning indefinitely.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
```diff
@@ -111,9 +111,23 @@ public boolean persistSuccess(
                     () -> {
```
mrouvroy-floe, line 112: [minor] **accepted=false idempotency record committed after stats already persisted, stranding a stats write against a non-terminal job**

This path folds `jobs.applyLeaseOutcome(...SUCCEEDED...)` into the idempotent-create body *after* the side-effecting `finalizeChunkedSuccess(...)` has already persisted stats (the parallel path in `LeasedFileGroupExecutionService` behaves the same). `applyLeaseOutcome` returns `false` (it does not throw) when the lease epoch is stale. In that case the body still returns a non-null `CreateResult`, so the idempotency record commits with `accepted=false` even though the stats writes happened and the job remains non-terminal.

Impact is limited because `resultId` is generated per attempt, so a re-lease produces a fresh `resultId` and a new idempotent attempt — but a stats write is stranded against a non-terminal job under a lost-lease race, potentially producing duplicate or inconsistent aggregates.

Fix: consider throwing a retryable or precondition status when `applyLeaseOutcome` returns `false`, or performing the stats write only after confirming the lease outcome was accepted.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/projection/ReconcileJobProjectionStore.java
```diff
@@ -63,12 +71,6 @@ public void upsert(StoredReconcileJobProjection projection) {
           && currentProjection.appliedGeneration() > projection.appliedGeneration()) {
```
mrouvroy-floe, line 71: [minor] **upsert now allows same-generation last-writer-wins overwrite without documentation**

The guard now only rejects a write when `currentProjection.appliedGeneration() > projection.appliedGeneration()` (incoming is an older generation). The previous behaviour also protected an *equal*-generation projection with differing content. With two concurrent refreshers at the same requested generation, one can now clobber a fresher same-generation projection with staler content until the next generation bump.

This converges in practice — generations are monotonic and reads self-heal via `freshProjection` — and arguably improves convergence speed, but it is a real semantic change to the concurrent write path that is easy to misread as an accidental dropped guard.

Fix: if the equal-generation last-writer-wins behaviour is intentional, add a short inline comment stating so to prevent future readers from treating it as a bug.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReconcileLeaseGrpcStatus.java
```diff
@@ -0,0 +1,83 @@
...
+        continue;
+      }
+      try {
+        ErrorInfo errorInfo = detail.unpack(ErrorInfo.class);
+        if (LEASE_PRECONDITION_FAILED_REASON.equals(errorInfo.getReason())
+            && ERROR_DOMAIN.equals(errorInfo.getDomain())) {
+          return true;
+        }
+      } catch (com.google.protobuf.InvalidProtocolBufferException ignored) {
+        return false;
```
mrouvroy-floe: [nit] **hasLeasePreconditionFailureDetail returns on first unparseable detail instead of continuing**

In the loop over `status.getDetailsList()`, `catch (InvalidProtocolBufferException) { return false; }` exits immediately rather than continuing to the next detail. If a status ever carries multiple `ErrorInfo`-typed details and an earlier one fails to unpack, a genuine lease-precondition marker in a later detail would be missed and the error treated as generic/retryable.

Low probability given details are packed uniformly by the server, but `continue;` is the safe and correct form here.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotPlanningReconcileExecutor.java
mcusack-floe: Fixed. The snapshot-planning lease-precondition catch path now calls `context.beforeHandledCompletion().run()` before returning the cancelled result.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReconcileLeaseGrpcStatus.java
mcusack-floe: hasLeasePreconditionFailureDetail now continues past unparseable detail entries instead of returning false on the first bad detail.

mcusack-floe:
Responses to review summary https://github.com/eng-floe/floecat/pull/380#pullrequestreview-4678195587: Connector/view planner submit methods: fixed. Connector and view submit paths now convert lease-precondition failures, and the remote executors catch them, call `beforeHandledCompletion`, and return the lease-invalid cancellation result instead of falling through to normal poller completion. In-memory cancellation-blocking predicate: fixed. The in-memory ready queue now respects the cancellation-blocking predicate before matching/leasing candidates, matching production behavior. Targeted unit coverage passed with the reconciler/service test set.

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/NativeReconcileReadyQueueStore.java
mcusack-floe: Fixed. Ready queue scan cursors are keyed by the normalized lease request as well as the slice, so different requester shapes no longer reset or starve each other through a shared per-slice cursor.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedPlannerWorkerService.java
mcusack-floe: Fixed. Missing declared plan-table/plan-snapshot chunks now fail with terminal `FAILED_PRECONDITION` and a declared-chunk message instead of retrying indefinitely. Present-but-incomplete chunks remain retryable.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
mcusack-floe: Fixed. File-group and snapshot-finalize success paths now throw a tagged lease-precondition failure when `applyLeaseOutcome` returns false, so idempotency success is not committed after side effects on a rejected lease.

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/projection/ReconcileJobProjectionStore.java
mcusack-floe: Fixed. Added an inline comment documenting the intentional same-generation last-writer-wins projection behavior and the self-healing read path.

PR #383, opened 2026-07-15
Make full-rescan stats publish atomically and skip existing incremental file stats

mcusack-floe:
## Summary

This changes file-group stats persistence so incremental reconciles no longer fail on already-present file stats. Incremental file-group submissions now use putTargetStatsIfAbsent, skip existing records, and only invalidate/cache-count records that were newly created.

For full rescans, file-group chunks no longer overwrite the active stats generation as they complete. Instead, each chunk stages file stats into an unpublished full-rescan-<parentJobId> generation. Snapshot finalization publishes that generation atomically with the merged aggregate stats, so readers continue seeing the previous published stats until finalize succeeds.
Also updates the test override stats store and coverage for staged generation publication, full-rescan finalization, duplicate stats handling, and incremental skip behavior.

## Type of Change

- [X] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
+++ b/service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
@@ -21,6 +21,7 @@ import static ai.floedb.floecat.service.error.impl.GeneratedErrorMessages.Messag
 
 import ai.floedb.floecat.catalog.rpc.IndexArtifactRecord;
 import ai.floedb.floecat.catalog.rpc.PutIndexArtifactItem;
+import ai.floedb.floecat.catalog.rpc.StatsTarget;
 import ai.floedb.floecat.catalog.rpc.Table;
 import ai.floedb.floecat.catalog.rpc.TargetStatsRecord;
 import ai.floedb.floecat.common.rpc.MutationMeta;
@@ -277,13 +278,17 @@ public class LeasedFileGroupExecutionService extends BaseServiceImpl {
                             () -> {
                               long creatorStartNanos = System.nanoTime();
                               long snapshotId = plannedTask.snapshotId();
-                              persistTargetStats(
-                                  principalContext,
-                                  tableId,
-                                  snapshotId,
-                                  requiredResultId,
-                                  fileStats,
-                                  metrics);
+                              if (lease.fullRescan) {
+                                validateFileStats(tableId, snapshotId, fileStats);
+                              } else {
+                                persistTargetStats(
+                                    principalContext,
+                                    tableId,
+                                    snapshotId,
+                                    requiredResultId,
+                                    fileStats,
+                                    metrics);
+                              }
                               persistIndexArtifacts(
                                   principalContext,
                                   tableId,
@@ -383,9 +388,18 @@ public class LeasedFileGroupExecutionService extends BaseServiceImpl {
                                   chunk ->
                                       partialAggregateRecords(chunk.getStatsRecordsList()).stream())
                               .toList();
+                      List<TargetStatsRecord> fileStats =
+                          stagedChunks.stream()
+                              .flatMap(
+                                  chunk ->
+                                      fileScopedStatsRecords(chunk.getStatsRecordsList()).stream())
+                              .toList();
                       List<TargetStatsRecord> mergedPartialAggregates =
                           mergedPartialAggregates(
                               tableId, latestTask.snapshotId(), latestTask, partialAggregates);
+                      if (lease.fullRescan) {
+                        persistDraftFileGroupStats(lease, tableId, latestTask, fileStats);
+                      }
                       ReconcileFileGroupTask persistedTask =
                           latestTask
                               .withFileResults(validatedFileResults)
@@ -504,12 +518,56 @@ public class LeasedFileGroupExecutionService extends BaseServiceImpl {
       return;
     }
     long batchStartNanos = System.nanoTime();
-    statsStore.putTargetStatsBatch(tableId, snapshotId, nonNullStats);
-    statsOrchestrator.invalidateStatsCache(tableId, snapshotId, nonNullStats);
+    List<TargetStatsRecord> created = new java.util.ArrayList<>(nonNullStats.size());
+    for (TargetStatsRecord record : nonNullStats) {
+      if (statsStore.putTargetStatsIfAbsent(record)) {
+        created.add(record);
+      }
+    }
+    if (!created.isEmpty()) {
+      statsOrchestrator.invalidateStatsCache(tableId, snapshotId, created);
+    }
+    metrics.statsItemCount += created.size();
     long batchNanos = System.nanoTime() - batchStartNanos;
     metrics.statsNanos += batchNanos;
     metrics.statsStorePutNanos += batchNanos;
-    metrics.statsItemCount += nonNullStats.size();
+  }
+
+  private void persistDraftFileGroupStats(
+      ReconcileJobStore.LeasedJob lease,
+      ResourceId tableId,
+      ReconcileFileGroupTask plannedTask,
+      List<TargetStatsRecord> fileStats) {
+    long snapshotId = plannedTask.snapshotId();
+    List<TargetStatsRecord> nonNullStats = validateFileStats(tableId, snapshotId, fileStats);
+    List<StatsTarget> targetsToReplace =
+        plannedTask.filePaths().stream().map(path -> StatsTargetIdentity.fileTarget(path)).toList();
+    statsStore.replaceTargetStatsInGeneration(
+        tableId, snapshotId, statsGenerationId(lease), targetsToReplace, nonNullStats);
+  }
+
+  private static List<TargetStatsRecord> validateFileStats(
+      ResourceId tableId, long snapshotId, List<TargetStatsRecord> fileStats) {
+    List<TargetStatsRecord> nonNullStats = nonNullStatsRecords(fileStats);
+    for (TargetStatsRecord record : nonNullStats) {
+      if (!record.hasFile() || !record.hasTarget()) {
+        throw new IllegalArgumentException("file-group stats must be file-target records");
+      }
+      if (!tableId.equals(record.getTableId()) || record.getSnapshotId() != snapshotId) {
+        throw new IllegalArgumentException("file-group stats do not match file-group task");
+      }
+    }
+    return nonNullStats;
+  }
+
+  static String statsGenerationId(ReconcileJobStore.LeasedJob lease) {
+    String parentJobId = lease == null || lease.parentJobId == null ? "" : lease.parentJobId.trim();
+    String jobId = lease == null || lease.jobId == null ? "" : lease.jobId.trim();
+    String source = parentJobId.isBlank() ? jobId : parentJobId;
+    if (source.isBlank()) {
+      throw new IllegalArgumentException("reconcile job id is required for stats generation");
+    }
+    return "full-rescan-" + source;
   }
 
   private void persistIndexArtifacts(
--- a/service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
+++ b/service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
@@ -71,8 +71,10 @@ import io.grpc.StatusRuntimeException;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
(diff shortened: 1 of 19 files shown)


service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/ReconcilePlannerScheduler.java
```diff
@@ -57,7 +57,10 @@ public class ReconcilePlannerScheduler {
           java.util.Set.of(
               ReconcileCapturePolicy.Output.TABLE_STATS,
               ReconcileCapturePolicy.Output.FILE_STATS,
-              ReconcileCapturePolicy.Output.COLUMN_STATS));
+              ReconcileCapturePolicy.Output.COLUMN_STATS,
+              ReconcileCapturePolicy.Output.PARQUET_PAGE_INDEX),
+          ReconcileCapturePolicy.DefaultColumnScope.ALL,
```
mrouvroy-floe: [major] **Undocumented capture-policy change bundled into a stats-atomicity fix**

The `DEFAULT_CAPTURE_POLICY` change adds `ReconcileCapturePolicy.Output.PARQUET_PAGE_INDEX` to the outputs and switches the default column scope from `FIRST_N` to `DefaultColumnScope.ALL`. Neither change is mentioned in the PR description, which is scoped to fixing stats persistence.

Every auto-scheduled reconcile will now capture parquet page indexes and column stats for *all* columns rather than the first 32 — a real increase in reconcile work, storage, and cost for every table the scheduler touches.

Split this into its own PR with an explicit justification, or document the intent and expected cost impact here if the change is intentional.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
```diff
@@ -504,12 +518,56 @@ private void persistTargetStats(
       return;
     }
     long batchStartNanos = System.nanoTime();
-    statsStore.putTargetStatsBatch(tableId, snapshotId, nonNullStats);
-    statsOrchestrator.invalidateStatsCache(tableId, snapshotId, nonNullStats);
+    List<TargetStatsRecord> created = new java.util.ArrayList<>(nonNullStats.size());
+    for (TargetStatsRecord record : nonNullStats) {
```
mrouvroy-floe: [major] **Incremental stats persist regresses from a batched write to sequential per-record round-trips**

The incremental path previously called `statsStore.putTargetStatsBatch(...)`, which wrote blobs and reserved pointers via `compareAndSetBatch` in batches of up to `MAX_POINTER_BATCH_SIZE` (100). The replacement loop calls `putTargetStatsIfAbsent` once per record.

Each `putTargetStatsIfAbsent` call synchronously performs `ensureActiveGeneration` (pointer read + manifest resolution), `createIfAbsent` (pointer get + blob head + blob put + single-key CAS), and `updateTargetLatestSnapshotIfNewer` (its own CAS-retry loop). For a chunk carrying hundreds of file stats, this means hundreds of sequential DynamoDB/S3 round-trips where there used to be a single batched CAS — materially inflating incremental-reconcile latency on the hot path this PR is targeting.

Consider reintroducing a batch-write abstraction for the incremental path, or confirm that the skip-if-present semantics are worth the latency cost and document it explicitly.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
```diff
@@ -504,12 +518,56 @@ private void persistTargetStats(
...
+      if (statsStore.putTargetStatsIfAbsent(record)) {
```
mrouvroy-floe: [major] **Incremental path drops table/snapshot validation, allowing stats pollution across keyspaces**

The previous `putTargetStatsBatch(tableId, snapshotId, nonNullStats)` call passed through `StatsRepository.requireRecordForSnapshot(...)`, which enforced that every record belongs to the supplied `tableId` and `snapshotId`. The new `putTargetStatsIfAbsent(record)` derives the storage key entirely from the record itself with no such check.

A malformed or misdirected worker response can therefore write stats under a different table or snapshot, polluting that keyspace, while the cache invalidation still targets only the planned `tableId`/`snapshotId` — leaving the affected entry stale and the corrupted entry untouched.

Validate each file-stat record against the planned `tableId` and `snapshotId` before entering the loop, or add a table/snapshot assertion inside `putTargetStatsIfAbsent`.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
```diff
@@ -504,12 +518,56 @@ private void persistTargetStats(
...
-    metrics.statsItemCount += nonNullStats.size();
+  }
+
+  private void persistDraftFileGroupStats(
+      ReconcileJobStore.LeasedJob lease,
+      ResourceId tableId,
+      ReconcileFileGroupTask plannedTask,
+      List<TargetStatsRecord> fileStats) {
+    long snapshotId = plannedTask.snapshotId();
+    List<TargetStatsRecord> nonNullStats = validateFileStats(tableId, snapshotId, fileStats);
```
mrouvroy-floe, line 540: [major] **Full-rescan draft generation stages out-of-group file stats that survive into the published manifest**

`persistDraftFileGroupStats` stages every validated file stat into the draft generation, but validation only checks table/snapshot identity — it does not verify that `record.getFile().getFilePath()` is one of the files in `plannedTask.filePaths()`.

The previous full-rescan path filtered the replacement set through `coverage.expectedFiles()`, so extra stats submitted by a buggy worker for files outside its assigned group were silently discarded. With the draft generation path, those records remain in the generation and become active when `publishStatsGeneration` flips the manifest pointer, potentially exposing corrupt or stale stats to readers.

Build a `Set` from `plannedTask.filePaths()` and reject any file-stat record whose canonical file path is not a member before staging it into the draft generation.

---

service/src/main/java/ai/floedb/floecat/service/repo/impl/StatsRepository.java
```diff
@@ -171,6 +171,74 @@ public void putTargetStatsBatch(
     }
   }
 
+  @Override
+  public void replaceTargetStatsInGeneration(
```
mrouvroy-floe, line 185: [minor] **Abandoned full-rescan draft generations are never reclaimed by GC**

`replaceTargetStatsInGeneration` writes record pointers and blobs under the `full-rescan-<parentJobId>` generation but never writes the generation manifest; that only happens in `publishStatsGeneration` at finalize. `deleteUnreferencedGenerations` keys reclamation on manifest existence — a missing manifest causes an unconditional `continue`, deliberately protecting in-flight replaces.

A full rescan that stages some chunks and then never finalizes (job crash, cancellation, or a retry that receives a new plan job and therefore a new `parentJobId`) leaves its staged record pointers and blobs permanently orphaned. GC will never reclaim them, and only a full `deleteAllStatsForSnapshot` prefix wipe removes them — an unbounded leak over time.

Consider a cleanup mechanism: a time-bounded TTL on unfinalized draft generation prefixes, a tombstone written on job failure/cancellation, or an audit pass that removes draft generations older than the max expected job duration.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
```diff
@@ -504,12 +518,56 @@ private void persistTargetStats(
...
+        throw new IllegalArgumentException("file-group stats do not match file-group task");
+      }
+    }
+    return nonNullStats;
+  }
+
+  static String statsGenerationId(ReconcileJobStore.LeasedJob lease) {
+    String parentJobId = lease == null || lease.parentJobId == null ? "" : lease.parentJobId.trim();
+    String jobId = lease == null || lease.jobId == null ? "" : lease.jobId.trim();
+    String source = parentJobId.isBlank() ? jobId : parentJobId;
```
mrouvroy-floe: [nit] **`statsGenerationId` fallback to `jobId` risks a silent generation mismatch**

When `parentJobId` is blank, `statsGenerationId` silently falls back to the lease's own `jobId`. Today this is safe because file-group children and the finalize job always carry a non-blank `parentJobId`, so both resolve to the same `full-rescan-<parentJobId>`.

If that invariant broke for one side, file-group chunks would stage into `full-rescan-<childJobId>` while the finalizer would publish `full-rescan-<finalizerJobId>` — a different, empty generation — silently dropping every staged file stat with no error.

Prefer failing loudly (throw on blank `parentJobId`) over a silent fallback, so a mismatch is caught immediately rather than surfacing as mysteriously missing stats.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/ReconcilePlannerScheduler.java
mcusack-floe: Addressed in the latest commit by removing the unrelated default capture-policy change from this PR. `DEFAULT_CAPTURE_POLICY` is back to the existing default outputs (`TABLE_STATS`, `FILE_STATS`, `COLUMN_STATS`) and no longer adds `PARQUET_PAGE_INDEX` or changes the default column scope, so this stats atomicity/GC patch does not alter auto-scheduled capture breadth.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
mcusack-floe: Addressed in the latest commit by restoring a batched API with the skip-if-present semantics this path needs. `StatsStore.putTargetStatsBatchIfAbsent(tableId, snapshotId, records)` now lets the incremental path reserve/create the missing records as a batch, and `LeasedFileGroupExecutionService` invalidates cache only for the records actually created. That removes the sequential per-record store loop while preserving idempotent duplicate handling.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
mcusack-floe: Addressed in the same batch-if-absent change. The new API keeps `tableId` and `snapshotId` in the method signature, and the repository validates every canonicalized record against that table/snapshot before writing. The latest commit also adds a regression test for rejecting records outside the requested snapshot/keyspace.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
mcusack-floe: Addressed in the latest commit. Full-rescan draft stats now go through `validateFileStats(...)` for table/snapshot/file-target validation and then check each staged file stat against the planned file-group paths before writing into the draft generation. Out-of-group file stats now fail before staging, so they cannot survive into the published manifest.

service/src/main/java/ai/floedb/floecat/service/repo/impl/StatsRepository.java
mcusack-floe: Addressed in the latest commit with job-aware durable cleanup rather than a time-based guess. Terminal failed/cancelled full-rescan `PLAN_SNAPSHOT` jobs now persist `statsCleanupState=PENDING`, only become eligible after descendants are terminal/finalized, and are revisited by cancellation/projection maintenance until cleanup completes.

The repository cleanup path is also fenced now: publication moves `WRITING -> PUBLISHING -> PUBLISHED`, while cleanup can only claim `WRITING -> DELETING`. `deleteUnpublishedStatsGeneration(...)` returns `DELETED`, `NOT_DELETABLE_PUBLISHED`, or `RETRYABLE_IN_PROGRESS`; `DurableReconcileJobStore` marks cleanup complete only for the first two and leaves `PUBLISHING`/in-progress cases pending for a later maintenance retry.

Regression coverage now includes transient cleanup failure followed by retry, no cleanup while descendants are live, published-generation refusal, the publish/delete interleaving after the final manifest check, and `RETRYABLE_IN_PROGRESS` followed by a later successful delete.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
mcusack-floe: Addressed in the latest commit by removing the silent fallback. `statsGenerationId(...)` now requires a nonblank parent job id and throws if that invariant is missing, so a file-group/finalizer mismatch fails loudly instead of writing to a generation derived from the child job id. There is also a regression test for the missing-parent case.

mrouvroy-floe (approved):

PR #384, opened 2026-07-15
Refresh AWS clients on closed connection pools

mcusack-floe:
## Summary

Fix some newly discovered connection pool shutdowns in AWS:
- Rebuild AWS credentials providers when refreshing Glue and S3 clients so refreshed clients do not retain stale/closed provider state
- Route Delta manifest materialization and Delta connector S3 access through fresh provider factories on client refresh
- Add coverage proving refreshed Glue/S3 client construction creates new credentials provider instances

## Type of Change

- [X] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist


core/connectors/common/src/main/java/ai/floedb/floecat/connector/common/auth/AwsGlueClientFactory.java
```diff
@@ -33,12 +34,12 @@ private AwsGlueClientFactory() {}
...
       Map<String, String> options, Map<String, String> authProps) {
     String region = resolveRegion(options, "us-east-1");
-    AwsCredentialsProvider credentials = resolveCredentials(options, authProps);
+    Supplier<AwsCredentialsProvider> credentials = credentialsProviderFactory(options, authProps);
     return new RefreshingAwsClient<>(
         () ->
             GlueClient.builder()
                 .region(Region.of(region))
-                .credentialsProvider(credentials)
+                .credentialsProvider(credentials.get())
```
mrouvroy-floe: [major] **Close refreshed credential providers**

This now gives each refreshed `GlueClient` a newly-built `AwsCredentialsProvider`, but that provider is not owned or closed anywhere. In AWS SDK v2, providers such as `DefaultCredentialsProvider` and `ProfileCredentialsProvider` implement `SdkAutoCloseable`; the generated `GlueClient.close()`/`S3Client.close()` closes only the client handler, and `RefreshingAwsClient` only receives the AWS service client.

That means every refresh after a connection-pool shutdown can leave the previous provider chain alive. This matters for default/profile chains that can wrap STS, web identity, container, or instance profile providers, and the same pattern appears in the S3 builders in `DeltaConnectorFactory` and `DeltaManifestMaterializer`.

Make the refresh factory own both resources, for example by wrapping the service client together with the provider and closing any `SdkAutoCloseable` provider when the old client is replaced/closed, or by adding an explicit cleanup hook to `RefreshingAwsClient`. Please also add coverage that the old provider is closed on refresh, not only that a new instance is built.

---

mcusack-floe: Addressed in `bfa6d1e50` (`Address PR change; centralize retry helper`).

This ended up being wider than the original Glue credential-provider leak because the leak was a symptom of a broader ownership problem: the old retry paths only owned the AWS service client, while the actual refreshable resource is often a bundle of service client + credential provider + dependent STS client. Fixing that in one call site would still have left different AWS clients with different definitions of refresh, retry, close, and closed-pool detection.

The commit moves the retry/refresh primitive into `core/engine-utils` as `ai.floedb.floecat.aws.RefreshingAwsClient` and makes it own:

- the service client plus associated closeable resources, including credential providers
- centralized `Connection pool shut down` detection
- exact failed-resource identity checks before replacement
- at-most-one retry for checked, unchecked, and async calls
- close-state handling and shutdown races
- refresh serialization for STS-dependent resources

Then the AWS call sites were moved onto that helper rather than keeping local retry loops:

- connector Glue/STS and Delta S3 clients
- storage S3 and DynamoDB managers/callers, including async DynamoDB completions
- service durable DynamoDB callers via manager operation APIs
- Secrets Manager default/per-account clients, including STS-dependent credential-provider invalidation

I also added coverage for provider/resource closure on refresh, async exceptional-completion retry, use-after-close, stale failed-client races, and the Secrets Manager STS-dependent cases, including concurrent same-account and cross-account credential-pool failures. The important bit is that refreshed clients now carry and close their associated credential providers instead of leaving provider chains behind, and the same helper now enforces that behavior across the AWS clients instead of relying on each caller to remember the retry/cleanup pattern.

core/engine-utils/src/main/java/ai/floedb/floecat/aws/RefreshingAwsClient.java
```diff
@@ -0,0 +1,342 @@
...
+    } catch (RuntimeException e) {
+      if (!ClosedAwsClientDetector.isConnectionPoolShutdown(e)) {
+        throw e;
+      }
+      warn(e);
+      refreshAfterFailure(resource, e);
+      return applyAsync(operation, currentResource());
+    }
+
+    CompletableFuture<R> result = new CompletableFuture<>();
```
mrouvroy-floe: [major] **`callAsync` drops upstream cancellation**

`callAsync` wraps the AWS SDK `CompletionStage` in a fresh `CompletableFuture`, but cancelling the returned future never cancels the underlying SDK future created at `operation.apply(...)`.

This regresses the DynamoDB store cancellation path. `DynamoDbKvStore.fromStage(...)` calls `future.cancel(true)` when a `Uni` subscription terminates, and an existing test (`get_subscription_cancel_cancels_underlying_future`) covers this contract for the fixed-client path. After this PR, the manager-backed producer passes `clientManager.callAsync(...)`, so `fromStage` only cancels the wrapper future while the real DynamoDB request continues running.

Abandoned reads/writes consume HTTP connections after request timeout or client disconnect, recreating the pooled-client pressure the PR aimed to fix.

Fix: propagate cancellation by chaining it — e.g. capture the inner `CompletableFuture` and cancel it when the outer one is cancelled, or return the SDK stage directly without wrapping.

---

storage/aws/src/main/java/ai/floedb/floecat/storage/aws/secrets/ProdSecretsManager.java
```diff
@@ -74,8 +74,21 @@ public ProdSecretsManager(AwsClients awsClients) {
...
+        new AtomicReference<>(secretsClient);
+    AtomicReference<StsClient> initialStsClient = new AtomicReference<>(stsClient);
+    this.secretsClient =
+        RefreshingAwsClient.withResourceFactory(
+            "Secrets Manager",
+            () ->
+                RefreshingAwsClient.clientResource(
+                    takeInitialOrBuild(initialSecretsClient, this::newSecretsClient)));
+    this.stsClient =
+        RefreshingAwsClient.withResourceFactory(
```
mrouvroy-floe, line 108: [minor] **Shared STS/Secrets clients still leak their credentials provider on refresh**

The per-account path was fixed correctly: `newPerAccountSecretsClientResource` wraps the `StsAssumeRoleCredentialsProvider` via `RefreshingAwsClient.closeableResource(creds)` so it is closed when the client is refreshed or closed.

However, the shared `secretsClient` and `stsClient` factories (see `newSecretsClient`/`newStsClient` at lines 292–303) build their client through `awsClients.secretsManagerClient()` / `awsClients.stsClient()`. Those no-arg overloads create a fresh `DefaultCredentialsProvider` internally via `resolveCredentials()` and never expose it. The `ClientResource` therefore only owns the service client, so on refresh the previous provider chain is not closed — the exact leak the earlier review flagged, still present for these two shared clients.

Fix: extract the `DefaultCredentialsProvider` created inside these factory methods and wrap it alongside the service client so it is closed on refresh, mirroring the pattern used in `newPerAccountSecretsClientResource`.

---

core/engine-utils/src/main/java/ai/floedb/floecat/aws/RefreshingAwsClient.java
mcusack-floe: Addressed. `RefreshingAwsClient.callAsync` now returns a cancellable wrapper that tracks the active SDK `CompletableFuture` and forwards `cancel(...)` to it. The active stage is set for both the initial call and the retry call, so cancellation propagates whether the request is still on the first attempt or has moved to the refreshed client. The completion handler also stops copying results after the wrapper has been cancelled.

Coverage added:
- `RefreshingAwsClientTest.callAsyncCancellationCancelsActiveStage`
- `RefreshingAwsClientTest.callAsyncCancellationCancelsRetryStage`
- `DynamoDbKvStoreTest.get_subscription_cancel_cancels_manager_wrapped_future`

Verified with:
`mvn -pl core/engine-utils,storage/aws -Dtest=RefreshingAwsClientTest,ProdSecretsManagerTest,DynamoDbKvStoreTest -Dsurefire.failIfNoSpecifiedTests=false test`

storage/aws/src/main/java/ai/floedb/floecat/storage/aws/secrets/ProdSecretsManager.java
mcusack-floe: Addressed. The shared Secrets Manager and STS refresh paths now use owned `ClientResource` factories instead of raw clients:
- `AwsClients.newSecretsManagerClientResource()` and `AwsClients.newStsClientResource()` build the default credentials provider, pass it to the service client builder, and wrap the provider with the client so refresh/close closes both resources.
- `ProdSecretsManager(AwsClients)` now wires its production shared clients directly to those resource factories, lazily, so the initial shared client is owned by `RefreshingAwsClient` from creation instead of being produced via `awsClients.secretsManagerClient()` / `awsClients.stsClient()` and losing the provider handle.
- The package-private raw-client constructor path still wraps supplied test clients as client-only resources, while refreshes through `AwsClients` use the owned resources.

Coverage added/updated:
- refreshed shared Secrets provider closes when replaced
- injected production constructor path closes the initial shared Secrets provider when the first client is replaced
- refreshed provider remains open while active and closes on manager shutdown

Verified with:
`mvn -pl core/engine-utils,storage/aws -Dtest=RefreshingAwsClientTest,ProdSecretsManagerTest,DynamoDbKvStoreTest -Dsurefire.failIfNoSpecifiedTests=false test`

storage/aws/src/main/java/ai/floedb/floecat/storage/aws/s3/S3BlobStore.java
```diff
@@ -133,22 +131,17 @@ public Optional<BlobHeader> head(String key) {
 
       throw mapAndWrap("HEAD", k, e);
     } catch (SdkClientException e) {
-      clientFailureHandler.accept(client, e);
       throw new StorageAbortRetryableException(msg("HEAD", k, e.getMessage()));
     } catch (RuntimeException e) {
```
mrouvroy-floe, line 136: [major] **Post-retry closed-pool failures no longer mapped to retryable exception**

`S3ClientManager.call(...)` now owns the first closed-pool refresh/retry, but `S3BlobStore` no longer maps a closed-pool failure that occurs on the retried client.

If the refreshed client also throws a `RuntimeException` containing `Connection pool shut down`, the catch blocks at lines 135, 157, 200, 254, 272, and 319 simply rethrow it as an unexpected runtime failure. Before this change, `mapClosedPoolOrRethrow(...)` converted that exhausted closed-pool case to `StorageAbortRetryableException`, which is the storage-layer retry contract.

Restore the post-retry mapping in `S3BlobStore` for `RuntimeException`s that still satisfy the closed-pool predicate, so a second transient pool shutdown surfaces as a retryable error rather than a hard failure. The centralized refresh in `S3ClientManager` can remain — the fix is purely in the catch blocks.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/DynamoReconcileLeaseBackend.java
```diff
@@ -67,7 +67,12 @@ public void bind(
       Supplier<DynamoDbClient> dynamoDbSupplier,
       String table,
       BiConsumer<DynamoDbClient, Throwable> clientFailureHandler) {
```
mrouvroy-floe: [minor] **Retained `bind` overload silently ignores `clientFailureHandler`**

The `bind(Supplier<DynamoDbClient>, String, BiConsumer<DynamoDbClient, Throwable>)` overload still accepts a `clientFailureHandler` argument, but the body calls `dynamoCaller.bind(dynamoDbSupplier)` and discards the handler entirely. The same pattern exists in `DynamoReconcileJobIndexBackend` and `DynamoReconcileReadyQueueBackend`.

Before this PR, passing a real handler to this overload enabled connection-pool-shutdown refresh. Production has migrated to the new `bind(DynamoDbClientManager, String)` overload (which handles refresh via `RefreshingAwsClient`), but the old signature is preserved with a parameter that is now a no-op. A future caller wiring production through this overload and expecting refresh will get none, with no compile-time or runtime signal.

Either remove the three-argument overload entirely, or add a deprecation annotation and a Javadoc note explaining that `clientFailureHandler` is ignored and callers should migrate to `bind(DynamoDbClientManager, String)`.

---

storage/aws/src/main/java/ai/floedb/floecat/storage/aws/s3/S3BlobStore.java
mcusack-floe: Addressed in `4c3211e69` (`Address PR comments`).

`S3BlobStore` now imports `ClosedAwsClientDetector` and routes each `catch (RuntimeException e)` for `HEAD`, `GET`, `PUT`, `LIST`, `DELETE`, and `DELETE_PREFIX` through `mapClosedPoolOrRethrow(...)`. If the post-refresh/retry failure still contains `Connection pool shut down`, it is converted back to `StorageAbortRetryableException` with the original failure as the cause; otherwise the runtime exception is rethrown unchanged.

I also added `S3BlobStoreTest` coverage for all six public operations to verify that a runtime closed-pool failure from the `S3Caller` path is surfaced as `StorageAbortRetryableException`.

Verified with:
`mvn -pl storage/aws,service -Dtest=S3BlobStoreTest,DynamoReconcileLeaseBackendTest,DynamoReconcileReadyQueueBackendTest,DurableReconcileJobStoreTest,DurableReconcileJobStoreLeaseOutcomeTest -Dsurefire.failIfNoSpecifiedTests=false test`

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/DynamoReconcileLeaseBackend.java
mcusack-floe: Addressed in `4c3211e69` (`Address PR comments`).

I removed the misleading `bind(Supplier<DynamoDbClient>, String, BiConsumer<DynamoDbClient, Throwable>)` overload from all three Dynamo reconcile backends:

- `DynamoReconcileLeaseBackend`
- `DynamoReconcileJobIndexBackend`
- `DynamoReconcileReadyQueueBackend`

For fixed-client test wiring, the remaining overload is now explicit: `bind(Supplier<DynamoDbClient>, String)`. Production should continue to use `bind(DynamoDbClientManager, String)`, which is the path that owns refresh through `RefreshingAwsClient`. This removes the no-op handler parameter instead of documenting a stale compatibility path.

Updated the affected tests to call the two-argument fixed-client overload.

Verified with:
`mvn -pl storage/aws,service -Dtest=S3BlobStoreTest,DynamoReconcileLeaseBackendTest,DynamoReconcileReadyQueueBackendTest,DurableReconcileJobStoreTest,DurableReconcileJobStoreLeaseOutcomeTest -Dsurefire.failIfNoSpecifiedTests=false test`

mrouvroy-floe (approved):
**[nit] `@Produces` `secretsManagerClient()`/`stsClient()` build a provider-tracking `ClientResource` then discard it** — `storage/aws/src/main/java/ai/floedb/floecat/storage/aws/AwsClients.java:1969`

Both CDI producers now do `return newSecretsManagerClientResource().client();` (and the STS equivalent). `newSecretsManagerClientResource()` deliberately wraps the client with `closeableResource(credentials)` so the provider can be closed later, but `.client()` throws that wrapper away, so the provider is never tracked for cleanup here.

This is harmless in practice — `ProdSecretsManager` now builds its own refreshing clients via `newSecretsManagerClientResource()`/`newStsClientResource()`, and a search finds no remaining `@Inject SecretsManagerClient`/`StsClient` consumers, so these two `@Singleton` producers appear to be dead (never instantiated). The provider would only leak for the app lifetime even if they were.

Consider removing the now-unused `@Produces` producers (and the `secretsManagerClient()`/`stsClient()` no-arg overloads) if nothing depends on them, to avoid the misleading impression that the discarded `ClientResource` participates in provider lifecycle management.

---

PR #385, opened 2026-07-15
fix: Harden reconcile job GC and add observability

mcusack-floe:
## Summary

This PR hardens durable reconcile job garbage collection and makes it more aggressive and observable.

Changes include:
- Makes reconcile job GC incremental with continuation tokens for canonical jobs, dedupe pointers, root summaries, ready queue entries, and account traversal.
- Cleans leaked generic `PointerStore` root-summary/projection pointers that native Dynamo job indexes cannot see.
- Deletes child job projection pointers during canonical job GC.
- Uses owned deletes for dedupe cleanup so repointed dedupe pointers are not removed incorrectly.
- Quarantines unreadable canonical jobs, ready entries, dedupe targets, and root summaries instead of treating unreadable payloads as orphans.
- Makes ready queue GC conservative around fresh/racing queue updates.
- Fixes account traversal fairness by resuming within a cached account page instead of restarting the page each tick.
- Increases reconcile job GC cadence and budget.
- Reduces finished job retention default from 7 days to 24 hours.
- Adds reconcile job GC summary logs, quarantine/error metrics, scheduler progress gauges, and telemetry docs/contract updates.

**Testing**
- `mvn -pl telemetry-hub/core,service -Dtest=GcMetricsTest,ServiceTelemetryContributorTest,ReconcileJobGcTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn -pl telemetry-hub/tool-docgen -am process-classes`
- `mvn -pl service -Dtest=ReconcileJobGcTest,ReconcileControlImplTest -Dsurefire.failIfNoSpecifiedTests=false test`

## Type of Change

- [X] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist


service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGc.java
```diff
@@ -229,27 +291,24 @@ public GlobalResult runReadySlice(String pageTokenIn) {
...
-          stale = !validReadyKeys.contains(ready.readyPointerKey());
+        JsonNode record = readRecordByReference(canonical.blobUri());
+        if (record == null) {
+          quarantined++;
+          continue;
         }
 
-        if (stale && readyQueueBackend.deleteReadyEntry(ready.readyPointerKey())) {
+        String state = text(record, "state");
+        if (!"JS_QUEUED".equals(state)
```
mrouvroy-floe: [major] **Stale ready pointers for JS_QUEUED jobs are never reclaimed**

**Problem:** `runReadySlice` skips deletion for any ready entry whose canonical record is still `JS_QUEUED`, without checking whether the scanned `readyPointerKey` matches the record's current ready keys. When a queued job is requeued to a later `readyPointerKey`, the old ready row remains visible as `JS_QUEUED` forever from this GC's perspective — even after `ready-stale-grace-ms` has elapsed.

**Impact:** Stale rows accumulate in the global ready scan and depend on the lease scanner to clean them opportunistically. The previous logic reconstructed the job's valid ready keys and deleted any non-current entry even for `JS_QUEUED` jobs; the new logic removes that backstop entirely. In normal operation `buildReadyQueueMutation` atomically prunes superseded pointers, so the residual risk is a stale pointer leaking only when that atomic delete was partial or lost — but the safety net is now gone.

**Fix:** Preserve the grace-window protection for genuinely fresh mismatches, but after the grace window restore the current-key check for queued records: build the valid key set from `readyPointerKey` plus `currentReadyPointerKeys(record)`, and delete only when the scanned key is not in that set.

---

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGc.java
```diff
@@ -193,15 +221,44 @@ && deleteJobIndexPointerIfOwned(dedupePointerKey, canonical.pointerKey())) {
       }
     }
 
+    String rootSummaryPrefix = Keys.reconcileRootJobSummaryByAccountPointerPrefix(accountId);
```
mrouvroy-floe, line 293: [major] **Connector root-summary pointers are never scanned and can be orphaned permanently**

**Problem:** The orphan root-summary GC pass only traverses `reconcileRootJobSummaryByAccountPointerPrefix(accountId)` and removes the connector summary as a side-effect after reading the account summary. Because `ReconcileJobRootSummaryStore.upsert` and `delete` write account and connector pointers in separate, non-atomic operations, a crash or partial failure after the account pointer is deleted can strand a `/root-summaries/by-connector/...` entry with no account-side counterpart.

**Impact:** This GC will never visit the orphaned connector pointer directly, so connector-filtered job lists can keep showing a deleted root job indefinitely.

**Fix:** Add a separate traversal over the account's `by-connector` root-summary prefix with its own continuation token, and run the same readable/quarantine/orphan-delete logic starting from each connector pointer.

---

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGc.java
```diff
@@ -111,12 +139,7 @@ public AccountResult runAccountSlice(String accountId, String jobTokenIn, String
         JsonNode record = readRecordByReference(canonical.blobUri());
         String jobId = decodeJobId(jobPrefix, canonical.pointerKey());
         if (record == null) {
-          if (jobIndexBackend.purgeEntriesByCanonicalReference(canonical.pointerKey())) {
-            ptrDeleted++;
-            if (jobId != null) {
-              blobDeleted += deleteJobBlobs(accountId, jobId);
-            }
-          }
+          canonicalQuarantined++;
```
mrouvroy-floe: [minor] **Unreadable canonical entries are quarantined forever with no reclamation path**

**Problem:** Previously an unreadable canonical record triggered `purgeEntriesByCanonicalReference(...)` plus blob deletion. Now it is only counted as `canonicalQuarantined` and skipped on every future tick, so the index entry and its blobs are never reclaimed.

**Impact:** Canonical entries always store job state inline (`encodeInlineJobState` → `inline:reconcile-job:`), and `readRecordByReference` only decodes that prefix. A `null` return therefore indicates a permanently corrupt/undecodable payload, not a transient read failure — quarantine never self-heals and these entries accumulate indefinitely. `purgeEntriesByCanonicalReference` is now dead production code (referenced only from tests/impls).

**Fix:** Document the intentional trade-off if quarantine-forever is desired, and either add a TTL-based automated purge path or restore targeted deletion for entries that remain unreadable past a configurable threshold.

---

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGc.java
```diff
@@ -394,7 +509,8 @@ private boolean deleteCanonicalFootprint(
         new ReconcileJobIndexStore.JobIndexWriteBatch(
             deletes,
             new ReconcileJobIndexStore.ReadyQueueMutation(
-                List.of(), new java.util.ArrayList<>(readyDeletes))));
+                List.of(), new java.util.ArrayList<>(readyDeletes))),
+        pointerDeletes);
```
mrouvroy-floe, line 728: [nit] **Default `compareAndSetBatch` no-op silently disables canonical GC for unimplemented backends**

**Problem:** `deleteCanonicalFootprint` relies on the two-arg `compareAndSetBatch(batch, pointerDeletes)` to atomically remove projection/root-summary pointers together with index entries. The interface default returns `false` (a no-op) when `extraPointerOps` is non-empty.

**Impact:** Both current backends (`Memory`/`Dynamo`) override it correctly. However, if a future backend forgets to override it, canonical GC silently stops reclaiming any root job — the batch is never applied, `expired`/`ptrDeleted` stay 0, and there is no error or log to diagnose the regression.

**Fix:** Replace the `return false` default with `throw new UnsupportedOperationException(...)` so an unimplemented backend fails loudly rather than silently disabling GC.

---

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGc.java
mcusack-floe: Resolved in 9472bc5c3. Account GC now has an independent connector root-summary scan with its own continuation token, and all account/connector summary scans limit fetches to the remaining batch budget so continuation tokens cannot skip unprocessed rows. Added coverage for connector-only orphan cleanup and partial-page continuation. Verified with `mvn -pl service -Dtest=ReconcileJobGcTest -Dsurefire.failIfNoSpecifiedTests=false test`.

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGc.java
mcusack-floe: Resolved in 9472bc5c3. `runReadySlice` now keeps the stale-grace protection for fresh mismatches but, after the grace window, deletes stale `JS_QUEUED` ready rows whose scanned key is not in `ReadyQueueKeys.readyPointerKeys(...)`. I also added regressions for superseded queued ready rows and next-attempt movement. Verified with `mvn -pl service -Dtest=ReconcileJobGcTest -Dsurefire.failIfNoSpecifiedTests=false test`.

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGc.java
mcusack-floe: Resolved in 9472bc5c3. Unreadable canonical entries now get an account-scoped quarantine marker that binds the canonical key hash, version, blob/reference, and first-seen time. A later pass purges only if the same unreadable canonical is still present past `floecat.gc.reconcile-jobs.canonical-quarantine-retention-ms`; readable replacements clear the marker, and conditional purge prevents deleting a raced replacement. Verified with quarantine retention, marker clearing, and replacement-race tests in `ReconcileJobGcTest`.

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGc.java
mcusack-floe: Resolved in 9472bc5c3. The default two-arg `ReconcileJobIndexBackend.compareAndSetBatch(...)` now fails loudly with `UnsupportedOperationException` when extra pointer operations are supplied and the backend has not overridden the mixed-operation path. Current Memory and Dynamo backends override it. Added `indexBackendDefaultRejectsExtraPointerOps` coverage and verified with `ReconcileJobGcTest`.

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGc.java
```diff
@@ -160,26 +214,32 @@ && deleteJobIndexPointerIfOwned(dedupePointerKey, canonical.pointerKey())) {
...
           continue;
         }
+        JsonNode record = readRecordByReference(canonical.blobUri());
+        if (record == null) {
+          dedupeQuarantined++;
+          continue;
+        }
 
         if (TERMINAL_STATES.contains(text(record, "state"))) {
           if (deleteJobIndexPointerIfPresent(dedupe.pointerKey())) {
```
mrouvroy-floe: [major] **Dedupe-scan delete races against re-enqueue: use owned delete**

`deleteJobIndexPointerIfPresent(dedupe.pointerKey())` reloads and deletes whatever version of the dedupe row is currently present, without verifying it still points at the scanned `dedupe.blobUri()`.

A racing enqueue can repoint the same dedupe key to a new live canonical job between the scan/load and this delete; GC will then remove the freshly-created dedupe pointer, orphaning the new job. This is exactly the class of race the canonical terminal path already avoids with `deleteJobIndexPointerIfOwned(...)`.

Use an owned delete in both dedupe-scan delete branches: require the current dedupe row's reference to still equal the scanned `dedupe.blobUri()` before proceeding with the delete.

---

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGc.java
```diff
@@ -309,6 +510,66 @@ private static long longValue(JsonNode node, String field, long defaultValue) {
...
+    if (summary.connectorId() != null && !summary.connectorId().isBlank()) {
+      String connectorSummaryKey =
+          Keys.reconcileRootJobSummaryByConnectorPointer(
+              accountId,
+              summary.connectorId(),
+              rootSummarySortableJobToken(summary.createdAtMs(), summary.jobId()));
+      appendPointerDeleteIfPresent(deletes, connectorSummaryKey);
+    }
+    java.util.ArrayList<PointerStore.CasOp> ops = new java.util.ArrayList<>();
+    ops.add(new PointerStore.CasCheckAbsent(canonicalKey));
```
mrouvroy-floe: [major] **Wrong Dynamo absence check in deleteRootSummaryIfOrphan allows stale delete**

`deleteRootSummaryIfOrphan()` guards the transaction with `new PointerStore.CasCheckAbsent(canonicalKey)`, but on the Dynamo backend this translates via `genericPointerKey(...)` to the generic pointer row at `accounts/<acct>/reconcile/jobs/by-id/<job>`, not the native canonical job row at `reconcile-job/<acct>` / `job/<job>`.

If a canonical job is recreated after the pre-check load but before the transaction commits, the generic row will be absent while the native row is populated; the transaction will still delete the root summary and projection, corrupting the new job. The in-memory backend does not expose this because it stores canonical rows in the generic `PointerStore` keyspace.

Move the race guard into the native Dynamo layer so the absence check targets the same row that a new canonical job would occupy.

---

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGcScheduler.java
```diff
@@ -122,35 +173,81 @@ void tick() {
...
         String jobToken = jobTokenByAccount.getOrDefault(accountId, "");
         String dedupeToken = dedupeTokenByAccount.getOrDefault(accountId, "");
+        String rootSummaryToken = rootSummaryTokenByAccount.getOrDefault(accountId, "");
+        String connectorRootSummaryToken =
+            connectorRootSummaryTokenByAccount.getOrDefault(accountId, "");
 
         long sliceStart = System.nanoTime();
-        var result = gc.runAccountSlice(accountId, jobToken, dedupeToken);
+        var result =
+            gc.runAccountSlice(
```
mrouvroy-floe: [major] **A failing account permanently wedges GC for all subsequent accounts**

`accountPageIndex` is only advanced *after* `gc.runAccountSlice(...)` returns successfully (line 277). The tick-level `catch (Throwable t)` at line 288 aborts the entire tick without advancing the cursor.

If `runAccountSlice` throws for the account at `accountPageIndex` — e.g. a DynamoDB throttle, a transient RPC error, a `TransactionCanceledException`, or the `UnsupportedOperationException` path in `ReconcileJobIndexBackend.compareAndSetBatch` — the tick ends with the cursor still pointing at that same account. Every subsequent tick resumes at the identical index, hits the same failure, and aborts again, so no account after the poison one ever gets collected.

Advance `accountPageIndex` (or skip/quarantine the offending account) inside the existing per-account error handling so a single bad account cannot permanently block all accounts behind it.

---

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGc.java
mcusack-floe: Resolved in 2fdfb6144. The dedupe scan now uses `deleteJobIndexPointerIfOwned(dedupe.pointerKey(), dedupe.blobUri())` in both delete branches, so GC only deletes the dedupe row if it still points at the scanned canonical reference. Added `accountSliceDoesNotDeleteDedupePointerRepointedDuringDedupeScan` to cover the scan/repoint/delete race. Verified with `mvn -pl service -Dtest=ReconcileJobGcTest -Dsurefire.failIfNoSpecifiedTests=false test`.

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGcScheduler.java
mcusack-floe: Resolved in 2fdfb6144. Account slice failures are now caught inside the account loop, recorded as `account-failed`, logged with the account id, and the scheduler advances the account cursor in all cases. Ready-slice and account-page fetch failures still fail the tick, but one bad account no longer blocks every account behind it. Verified with `mvn -pl service -Dtest=ReconcileJobGcTest -Dsurefire.failIfNoSpecifiedTests=false test`.

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGc.java
mcusack-floe: Resolved in 2fdfb6144. Root-summary orphan cleanup no longer encodes the canonical absence guard as a generic `PointerStore.CasCheckAbsent`; it now submits a native `ReconcileJobIndexStore.JobIndexCheckAbsent(canonicalKey)` in the job-index batch while keeping summary/projection deletes as pointer CAS ops. Memory maps that to pointer-store CAS, and Dynamo maps it through the native job-index key parser, so the transaction checks the same canonical row used by recreate/upsert. Verified with `ReconcileJobGcTest`.

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/DynamoReconcileJobIndexBackend.java
```diff
@@ -276,6 +276,78 @@ public boolean purgeEntriesByCanonicalReference(String canonicalPointerKey) {
...
+    Map<String, Map<String, AttributeValue>> matches =
+        matchingCanonicalReferenceItems(canonicalPointerKey);
+    Map<String, AttributeValue> canonical = matches.get(canonicalPointerKey);
+    if (canonical == null
+        || longAttr(canonical, ATTR_VERSION) != expectedVersion
+        || !java.util.Objects.equals(
+            stringAttr(canonical, JobIndexBackendSupport.ATTR_BLOB_URI), expectedBlobUri)) {
+      return false;
+    }
+    if (matches.isEmpty() || matches.size() > 100) {
```
mrouvroy-floe: [major] **Large canonical purges silently fail and stall GC permanently**

`purgeEntriesByCanonicalReference` returns `false` whenever more than 100 rows reference the quarantined canonical, leaving the canonical, its marker, and all secondary rows in place indefinitely.

This is the worst case for GC correctness: the oversized match set is rescanned on every subsequent tick, `false` is returned each time, and the corrupt canonical is never collected — a permanent leak of native index rows.

Make the delete incremental: delete a bounded page of secondary rows per transaction, track progress across ticks, and only remove the canonical/marker once all secondary rows are gone.

---

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGc.java
```diff
@@ -78,47 +114,65 @@ public AccountResult runAccountSlice(String accountId, String jobTokenIn, String
...
             ptrDeleted++;
             if (jobId != null) {
               blobDeleted += deleteJobBlobs(accountId, jobId);
             }
+          } else {
+            canonicalQuarantined++;
           }
           continue;
         }
+        clearCanonicalQuarantineMarker(accountId, canonical);
```
mrouvroy-floe: [minor] **`clearCanonicalQuarantineMarker` issues a `PointerStore.get` for every readable canonical on every pass**

`clearCanonicalQuarantineMarker` is called unconditionally for every readable canonical job on every GC pass. It issues a `pointerStore.get(markerKey)` that almost always misses, because quarantine markers only exist for the rare case of a transiently-unreadable canonical that has since recovered.

With this PR the GC cadence drops from 10 m to 1 m and the batch limit rises from 1 000 to 5 000, multiplying the effect into thousands of near-always-empty point reads against `floecat_pointers` per account slice per minute. These negative lookups are unlikely to be served by existing caches, adding real read traffic in the hot path.

Gate the call: only invoke `clearCanonicalQuarantineMarker` when the canonical was previously observed as unreadable (e.g. track that state), or check for the marker's existence only after detecting a readable-again transition rather than on every pass.

---

service/src/main/resources/application.properties
```diff
@@ -92,14 +92,15 @@ floecat.gc.transaction.accounts-page-size=50
...
-floecat.gc.reconcile-jobs.retention-ms=604800000
+floecat.gc.reconcile-jobs.tick-every=1m
+floecat.gc.reconcile-jobs.page-size=500
+floecat.gc.reconcile-jobs.batch-limit=5000
+floecat.gc.reconcile-jobs.global-ready-batch-limit=5000
+floecat.gc.reconcile-jobs.ready-stale-grace-ms=60000
+floecat.gc.reconcile-jobs.slice-millis=10000
+floecat.gc.reconcile-jobs.max-tick-millis=30000
+floecat.gc.reconcile-jobs.accounts-page-size=200
+floecat.gc.reconcile-jobs.retention-ms=86400000
```
mrouvroy-floe, line 103: [minor] **Finished-job retention silently reduced from 7 d to 24 h**

The default finished reconcile-job retention drops from 7 days to 24 hours here (and in `ReconcilerSettingsStore.DEFAULT_FINISHED_JOB_RETENTION_MS`). Operator dashboards, support workflows, and API consumers that expected up to a week of finished-job history will lose that history after a single day once this ships.

The PR marks "No breaking changes", but this is a user-visible behavioral change that can surface as "jobs disappeared" for anyone relying on the old window.

Confirm with stakeholders that 24 h is acceptable for observability and debugging needs, and call the retention reduction out explicitly in deployment notes and the changelog so it is not a silent surprise.

---

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGcScheduler.java
```diff
@@ -54,11 +55,20 @@ public class ReconcileJobGcScheduler {
...
+  private final AtomicInteger accountsProcessedLastTick = new AtomicInteger(0);
+  private final AtomicInteger quarantinedLastTick = new AtomicInteger(0);
+  private final AtomicInteger deletedLastTick = new AtomicInteger(0);
   private ScheduledTaskMetrics taskMetrics;
 
   private final Map<String, String> jobTokenByAccount = new ConcurrentHashMap<>();
   private final Map<String, String> dedupeTokenByAccount = new ConcurrentHashMap<>();
+  private final Map<String, String> rootSummaryTokenByAccount = new ConcurrentHashMap<>();
+  private final Map<String, String> connectorRootSummaryTokenByAccount = new ConcurrentHashMap<>();
   private volatile String readyToken = "";
```
mrouvroy-floe, line 68: [nit] **Continuation-token maps never evict deleted accounts**

`jobTokenByAccount`, `dedupeTokenByAccount`, `rootSummaryTokenByAccount`, and `connectorRootSummaryTokenByAccount` only remove an account entry when a slice returns a blank token. An account deleted while it still holds a non-blank token drops out of `accountRepo.list` but its entry lingers in all four maps indefinitely.

Over long-running processes with account churn this is slow unbounded growth; it also inflates the new `RECONCILE_JOB_ACTIVE_ACCOUNT_TOKENS` gauge with stale entries. Two of these maps predate the PR, but since two more are added here, consider periodically pruning entries for account IDs no longer seen in a full sweep.

---

mcusack-floe: Decision: valid, but we are deferring this to a follow-up. Two of the token maps already had this lifecycle behavior before the PR, and the right fix is a small full-sweep/prune mechanism that treats all four maps consistently. I am not changing this PR for the token-map eviction item.

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGcScheduler.java
mcusack-floe: Addendum: this is now covered by a focused scheduler regression test as well. `ReconcileJobGcSchedulerTest.tickAdvancesPastFailingAccountSlice` makes `acct-a` throw from `runAccountSlice` and asserts the same tick still invokes GC for `acct-b`. The scheduler fix catches per-account failures inside the account loop and advances the cursor in `finally`, preserving later-account progress. Verified with `mvn -pl service -Dtest=ReconcileJobGcTest,ReconcileJobGcSchedulerTest -Dsurefire.failIfNoSpecifiedTests=false test`.

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGc.java
mcusack-floe: Resolved in the latest pushed changes. `clearCanonicalQuarantineMarker` is no longer called for every readable canonical. Quarantine markers now carry the canonical pointer key, and account GC scans the quarantine-marker prefix directly; it only performs the canonical read for actual marker rows and clears the marker when that canonical has become readable again. Existing 3-line marker payloads are migrated before purge so the cleanup remains compatible with already-written markers. Verified with `mvn -pl service -Dtest=ReconcileJobGcTest,ReconcileJobGcSchedulerTest -Dsurefire.failIfNoSpecifiedTests=false test`.

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/DynamoReconcileJobIndexBackend.java
mcusack-floe: Resolved in the latest pushed changes. Dynamo canonical purge is now incremental instead of permanently returning false for >100 referenced rows. The conditional purge path fetches a bounded match set, keeps a native canonical version/blob condition in the transaction, deletes a bounded set of secondary rows, and returns false only to keep the quarantine marker/canonical for the next pass. Once the remaining reference set fits in one transaction, the final pass removes the canonical and marker normally. Verified with `mvn -pl service -Dtest=ReconcileJobGcTest,ReconcileJobGcSchedulerTest -Dsurefire.failIfNoSpecifiedTests=false test`.

service/src/main/resources/application.properties
mcusack-floe: Decision: the 24h finished reconcile-job retention default is intentional for this stage. I updated the docs so this is no longer silent: `docs/service.md` now calls out the 24h default and `floecat.gc.reconcile-jobs.retention-ms=86400000`, and `docs/cli-reference.md` documents the per-connector override via `connector settings update --finished-job-retention-sec`. No code change planned for this item.

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGcScheduler.java
```diff
@@ -54,11 +55,20 @@ public class ReconcileJobGcScheduler {
   private final AtomicInteger enabledGauge = new AtomicInteger(0);
   private final AtomicLong lastTickStartMs = new AtomicLong(0);
   private final AtomicLong lastTickEndMs = new AtomicLong(0);
+  private final AtomicInteger accountsProcessedLastTick = new AtomicInteger(0);
+  private final AtomicInteger quarantinedLastTick = new AtomicInteger(0);
+  private final AtomicInteger deletedLastTick = new AtomicInteger(0);
   private ScheduledTaskMetrics taskMetrics;
 
   private final Map<String, String> jobTokenByAccount = new ConcurrentHashMap<>();
```
mrouvroy-floe, line 63: [minor] **Continuation-token maps never evict deleted accounts**

`jobTokenByAccount`, `dedupeTokenByAccount`, `rootSummaryTokenByAccount`, and `connectorRootSummaryTokenByAccount` only drop an account entry when a slice returns a blank token (see `updateAccountTokens`). An account deleted while it still holds a non-blank continuation token disappears from `accountRepo.list(...)`, so its slice is never run again and its map entries are never cleared.

Over a long-running process with account churn this is slow unbounded memory growth, and it also permanently inflates the new `RECONCILE_JOB_ACTIVE_ACCOUNT_TOKENS` gauge, which is meant to reflect in-flight GC work.

Consider pruning entries for account IDs no longer observed in the current traversal cycle — e.g. retain only keys present in the freshly listed pages once a full cursor cycle completes, or drop an account's tokens when its slice is skipped because the account is gone.

---

service/src/main/java/ai/floedb/floecat/service/gc/ReconcileJobGc.java
```diff
@@ -158,31 +210,56 @@ && deleteJobIndexPointerIfOwned(dedupePointerKey, canonical.pointerKey())) {
       }
     }
 
+    while (scanned < batchLimit && System.currentTimeMillis() < deadline && pointerStore != null) {
```
mrouvroy-floe, line 227: [nit] **Quarantine-marker cleanup pass never pages past the first page**

Unlike every other scan in `runAccountSlice` (canonical, dedupe, root-summary, connector-summary), this loop always passes an empty page token and unconditionally `break`s after the first page (line 229). The computed `next` token is discarded and no per-account continuation token is stored across ticks.

If an account accumulates more quarantine markers than `min(pageSize, remaining)` and the front of the page is stable (canonicals still unreadable, so `clearCanonicalQuarantineMarkerIfReadable` is a no-op), markers beyond the first page are never revisited. Those whose canonical later recovers will leak in the pointer store indefinitely, since the main canonical loop only removes markers on successful post-retention purge.

Give this pass its own continuation token (as the root-summary scans do) so it drains the full marker set incrementally rather than re-reading only the first page every tick.

---

mcusack-floe: Resolved in the pushed changes. The canonical-quarantine marker cleanup pass now has its own continuation token (`nextCanonicalQuarantineToken`) carried through `ReconcileJobGc.AccountResult` and `ReconcileJobGcScheduler`, so it no longer restarts at the first marker page every tick. Added `ReconcileJobGcTest.accountSliceContinuesCanonicalQuarantineMarkerCleanupAfterPartialPage` to cover the partial-page case where a readable marker sits beyond the first page. Verified with `mvn -pl service -Dtest=ReconcileJobGcTest,ReconcileJobGcSchedulerTest,SnapshotFinalizeReconcileExecutorTest -Dsurefire.failIfNoSpecifiedTests=false test`.

mrouvroy-floe (approved):

PR #386, opened 2026-07-17
Planner stats: pinned-generation reads, verbatim sketch propagation, and broken-root self-healing

mrouvroy-floe:
## Summary

Fixes two planner-stats regressions surfaced by the TableRoot rework (#378/#381) and closes the
operational gap that let tables with broken roots fail every query forever. Three themes:

### 1. Planner reads are pinned and reproducible

Planner stats lookups now resolve in the generation frozen by the query's pin instead of racing
the live pointer, with a pinned-first ladder: cache (pinned keyspace) → pinned generation →
same-snapshot gap-fill → stale (≤ pinned snapshot) → capture. Completeness is planner-aware
(`satisfiesNeed`): a record that exists but cannot serve every requested stat role/sketch is a
partial, not a hit — so resolution never stops at a record the planner can only consume
downgraded, and never silently drifts to a newer snapshot. Root causes of the user-visible
NOT_FOUND are fixed underneath: system column ids fall back to ordinals (with build-time
collision validation), and legacy requests with an unspecified resource kind normalize to TABLE.

### 2. Producer sketches propagate verbatim

The rollup dropped `ScalarStats.sketches` and non-theta NDV envelope payloads even when a single
source contributed — downgrading every quantile/CDF consumer. A sole source's sketch payloads and
its entire NDV envelope now propagate untouched through both rollup stages; multi-source payloads
are dropped, never merged. Generations only enrich: publishing a new generation carries a
superseded generation's sketches forward (payload-only, incoming wins per role/type, no target
resurrection), and NDV envelopes are gated on their mode oneof so an estimate-less carrier
envelope can never read as a bogus zero estimate (CONTRACT note on `Ndv` in stats.proto).

### 3. Broken roots converge instead of failing forever

A table whose root pointer named a swept blob, or whose root violated its invariants, failed
every query identically with nothing scheduled to repair it. Query-path reads now report observed
breakage into the existing `RootResyncQueue` re-drive (new `RootRepairRequests`: rate-limited per
table, fire-and-forget, never masks the query's own error); the periodic GC pass re-derives the
root from committed state. Expected client states (fresh table, no currency) and replaced-blob
version mismatches deliberately do not report. Snapshot-not-found errors now name the table
instead of rendering an unfilled `{id}` placeholder.

Also: `PlannerStatsResolver` extracted from the orchestrator (behavior-preserving), per-target
lookup-outcome diagnostics (`planner_lookup_outcomes` counter + DEBUG summary), and the pin's
frozen stats generation is GC-rooted directly.

## Type of Change

- [x] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/statistics/PlannerStatsResolver.java
+++ b/service/src/main/java/ai/floedb/floecat/service/statistics/PlannerStatsResolver.java
@@ -28,7 +28,6 @@ import ai.floedb.floecat.stats.spi.StatsStore;
 import ai.floedb.floecat.stats.spi.StatsSyncOutcome;
 import ai.floedb.floecat.storage.errors.StorageAbortRetryableException;
 import ai.floedb.floecat.telemetry.MetricId;
-import ai.floedb.floecat.telemetry.Observability;
 import ai.floedb.floecat.telemetry.Tag;
 import ai.floedb.floecat.telemetry.Telemetry.TagKey;
 import com.github.benmanes.caffeine.cache.Cache;
@@ -69,9 +68,6 @@ import org.jboss.logging.Logger;
 final class PlannerStatsResolver {
 
   private static final Logger LOG = Logger.getLogger(PlannerStatsResolver.class);
-  // Same tags as StatsOrchestrator so the extraction does not move any dashboard series.
-  private static final String COMPONENT = "service";
-  private static final String OPERATION = "stats_orchestrator";
 
   /**
    * Fully-qualified cache key for a single column-stats record.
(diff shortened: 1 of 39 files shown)


service/src/main/java/ai/floedb/floecat/service/statistics/PlannerStatsResolver.java
```diff
@@ -0,0 +1,685 @@
...
+    // 1. Cache check in the primary keyspace. A cached record that fails its completeness
+    // predicate reads as a miss — the store may hold an enriched generation this query can use —
+    // but is NOT evicted: it still satisfies lesser needs, and records are immutable per
+    // generation so re-reads converge on the same bytes.
+    java.util.Map<String, StatsResolutionResult> out =
+        new java.util.LinkedHashMap<>(requests.size());
+    java.util.List<StatsCaptureRequest> cacheMisses = new java.util.ArrayList<>();
+    for (StatsCaptureRequest req : requests) {
+      String key = storageId(req);
+      TargetStatsRecord cached = statsCache.getIfPresent(cacheKeyFor(req, primaryToken));
```
mcusack-floe, line 251: Medium: [PlannerStatsResolver.java (line 244)](/Users/markcusack/development/floe/metacat/service/src/main/java/ai/floedb/floecat/service/statistics/PlannerStatsResolver.java:244) drops incomplete cached records before the fallback ladder. If a scalar-only pinned record is cached from an earlier scalar request, a later sketch request treats it as a cache miss and re-reads the pinned generation. If that re-read fails or the generation is unavailable, the resolver falls through to newest/stale/capture, but partialPinned is only populated from the store read, not from the cached record. That violates the class invariant that partial pinned records are still valid hits and “never fall through to stale or capture.” I’d keep incomplete cached records in a cachedPartial map and use them as the same last-resort partial candidate while still attempting the richer store/newest read.

mcusack-floe (changes requested):

service/src/main/java/ai/floedb/floecat/service/statistics/PlannerStatsResolver.java
mrouvroy-floe: Good catch — fixed. Incomplete cached records are now kept in a `cachedPartial` map at the cache-check step (the record is still a miss for the richer need, but retained), and before the newest gap-fill I seed `partialPinned` from it for any target the pinned store re-read didn't cover:

```java
if (!partialPinned.containsKey(key) && cachedPartial.containsKey(key)) {
    partialPinned.put(key, cachedPartial.get(key));
}
```

Store-derived partials still take precedence (they're fresher reads of the same pinned generation); the cached partial is the fallback exactly when the pinned re-read fails or the generation is unavailable. So the scalar-only-cached + later-sketch-request case now serves the cached record as a `PARTIAL` hit instead of falling through to stale/capture — the invariant holds.

Added a regression test, `resolvePlannerBatch_cachedPartialServesWhenPinnedRereadFailsAndNewestEmpty`: primes the cache with a scalar-only record, makes the pinned re-read throw and the newest generation come back empty with `stale_ok=false`, and asserts the cached partial is served (`verify(store, never()).getStaleTargetStatsBatch(...)` confirms no stale/capture).

mcusack-floe (approved):

PR #387, opened 2026-07-17
fix: remove stale paused cancellation cleanup markers

mcusack-floe:
## Summary

This updates reconcile cancellation maintenance to validate paused cleanup markers before skipping them. If a paused marker points to a missing, mismatched, or no-longer-cancelling root job, it is now deleted as obsolete instead of being scanned forever.

Also changes cancellation maintenance’s default tick interval from 1s to 5s and moves no-op/paused-only maintenance summaries to DEBUG, while keeping real cleanup/deletes/failures at INFO.

Test coverage added for obsolete paused marker cleanup via ReconcileMaintenanceServicesTest.

## Type of Change

- [X] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/DurableReconcileJobStore.java
+++ b/service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/DurableReconcileJobStore.java
@@ -488,7 +488,10 @@ public class DurableReconcileJobStore implements ReconcileJobStore {
       cancellationMaintenanceService = new ReconcileCancellationMaintenanceService();
     }
     cancellationMaintenanceService.bind(
-        pointerStore, this::cleanupCancellationRoot, readyScanLimit);
+        pointerStore,
+        this::cleanupCancellationRoot,
+        this::isObsoleteCancellationCleanupRoot,
+        readyScanLimit);
     return cancellationMaintenanceService;
   }
 
@@ -2219,6 +2222,20 @@ public class DurableReconcileJobStore implements ReconcileJobStore {
       throw ReconcileCancellationMaintenanceService.obsoleteMarker();
     }
     return propagateDirectChildCancellation(loaded.record, request, childPageSize);
+  }
+
+  private boolean isObsoleteCancellationCleanupRoot(
+      ReconcileCancellationMaintenanceService.CancellationCleanupRequest request) {
+    if (request == null) {
+      return true;
+    }
+    String accountId = blankToEmpty(request.accountId());
+    String rootJobId = blankToEmpty(request.rootJobId());
+    StoredEnvelope loaded = loadByAnyAccount(rootJobId).orElse(null);
+    return loaded == null
+        || loaded.record == null
+        || !accountId.equals(blankToEmpty(loaded.record.accountId))
+        || !isCancellationState(loaded.record.state);
   }
 
   private ReconcileCancellationMaintenanceService.CancellationCleanupResult
--- a/service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/DurableReconcileJobStore.java
+++ b/service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/DurableReconcileJobStore.java
@@ -2246,7 +2246,8 @@ public class DurableReconcileJobStore implements ReconcileJobStore {
     return root != null
         && "JS_CANCELLED".equals(blankToEmpty(root.state))
         && root.finishedAtMs > 0L
-        && root.childrenFinalized;
+        && root.childrenFinalized
+        && !needsAbandonedFullRescanStatsCleanup(root);
   }
 
   private ReconcileCancellationMaintenanceService.CancellationCleanupResult
--- a/service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/DurableReconcileJobStore.java
+++ b/service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/DurableReconcileJobStore.java
@@ -2212,30 +2212,41 @@ public class DurableReconcileJobStore implements ReconcileJobStore {
       return new ReconcileCancellationMaintenanceService.CancellationCleanupResult(
           true, "", false, false, false);
     }
+    StoredReconcileJob root = activeCancellationCleanupRoot(request);
+    if (root == null) {
+      throw ReconcileCancellationMaintenanceService.obsoleteMarker();
+    }
+    return propagateDirectChildCancellation(root, request, childPageSize);
+  }
+
+  private boolean isObsoleteCancellationCleanupRoot(
+      ReconcileCancellationMaintenanceService.CancellationCleanupRequest request) {
+    return activeCancellationCleanupRoot(request) == null;
+  }
+
+  private StoredReconcileJob activeCancellationCleanupRoot(
+      ReconcileCancellationMaintenanceService.CancellationCleanupRequest request) {
+    if (request == null) {
+      return null;
+    }
     String accountId = blankToEmpty(request.accountId());
     String rootJobId = blankToEmpty(request.rootJobId());
     StoredEnvelope loaded = loadByAnyAccount(rootJobId).orElse(null);
     if (loaded == null
         || loaded.record == null
         || !accountId.equals(blankToEmpty(loaded.record.accountId))
-        || !isCancellationState(loaded.record.state)) {
-      throw ReconcileCancellationMaintenanceService.obsoleteMarker();
+        || !isCancellationState(loaded.record.state)
+        || cancellationCleanupRootComplete(loaded.record)) {
+      return null;
     }
-    return propagateDirectChildCancellation(loaded.record, request, childPageSize);
+    return loaded.record;
   }
 
-  private boolean isObsoleteCancellationCleanupRoot(
-      ReconcileCancellationMaintenanceService.CancellationCleanupRequest request) {
-    if (request == null) {
-      return true;
-    }
-    String accountId = blankToEmpty(request.accountId());
-    String rootJobId = blankToEmpty(request.rootJobId());
-    StoredEnvelope loaded = loadByAnyAccount(rootJobId).orElse(null);
-    return loaded == null
-        || loaded.record == null
-        || !accountId.equals(blankToEmpty(loaded.record.accountId))
-        || !isCancellationState(loaded.record.state);
+  private boolean cancellationCleanupRootComplete(StoredReconcileJob root) {
+    return root != null
+        && "JS_CANCELLED".equals(blankToEmpty(root.state))
+        && root.finishedAtMs > 0L
+        && root.childrenFinalized;
   }
 
   private ReconcileCancellationMaintenanceService.CancellationCleanupResult
--- a/service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/queue/ReconcileCancellationMaintenanceService.java
+++ b/service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/queue/ReconcileCancellationMaintenanceService.java
@@ -132,12 +132,15 @@ public class ReconcileCancellationMaintenanceService {
           continue;
         }
         if (marker.request().paused()) {
-          paused++;
           if (isObsoleteCancellationRoot(marker.request())) {
             if (pointerStore.compareAndDelete(pointer.getKey(), pointer.getVersion())) {
               obsoleteDeleted++;
               deleted++;
+            } else {
+              paused++;
             }
+          } else {
+            paused++;
           }
           continue;
         }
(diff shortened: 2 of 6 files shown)


service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/queue/ReconcileCancellationMaintenanceService.java
```diff
@@ -99,6 +132,13 @@ private CancellationStats cleanupCancellationMarkers(long deadlineMs) {
           continue;
         }
         if (marker.request().paused()) {
+          paused++;
```
mrouvroy-floe: [minor] **`paused` stat double-counts markers that are deleted as obsolete**

A paused marker that turns out to be obsolete increments both `paused` and `obsoleteDeleted`/`deleted` for the same pointer.

In the summary log an operator reading `paused=5 obsolete_deleted=3` would reasonably assume 5 markers are still paused/pending, when in fact 3 of those were just deleted and only 2 remain. Since a stated goal of this PR is cleaner observability (paused-only ticks routed to DEBUG), a `paused` count that conflates "still paused" with "was paused but deleted" undercuts that.

Consider incrementing `paused` only when the marker is kept (i.e. not obsolete), so the counters partition the scanned markers cleanly:

```java
if (marker.request().paused()) {
  if (isObsoleteCancellationRoot(marker.request())
      && pointerStore.compareAndDelete(pointer.getKey(), pointer.getVersion())) {
    obsoleteDeleted++;
    deleted++;
  } else {
    paused++;
  }
  continue;
}
```

---

mrouvroy-floe (approved):

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/DurableReconcileJobStore.java
```diff
@@ -2221,6 +2224,20 @@ private ReconcileCancellationMaintenanceService.CancellationCleanupResult cleanu
...
+    if (request == null) {
+      return true;
+    }
+    String accountId = blankToEmpty(request.accountId());
+    String rootJobId = blankToEmpty(request.rootJobId());
+    StoredEnvelope loaded = loadByAnyAccount(rootJobId).orElse(null);
+    return loaded == null
+        || loaded.record == null
+        || !accountId.equals(blankToEmpty(loaded.record.accountId))
+        || !isCancellationState(loaded.record.state);
```
mrouvroy-floe: [major] **Completed JS_CANCELLED roots retain paused markers indefinitely**

`isObsoleteCancellationCleanupRoot` treats any root in `isCancellationState` (including `JS_CANCELLED`) as non-obsolete, so a paused marker whose root is already `JS_CANCELLED` with `finishedAtMs > 0` and `childrenFinalized == true` is never deleted.

This is inconsistent with the rest of the system: `requestParentCancellationCleanupIfNeeded` explicitly skips requesting further cleanup for a finished `JS_CANCELLED` root, and active cleanup returns complete for that state. The paused-marker branch in `ReconcileCancellationMaintenanceService` will simply increment `paused` and continue on every tick, leaving a stale pointer-store entry that is scanned forever.

Fix: refine `isObsoleteCancellationCleanupRoot` to treat a `JS_CANCELLED` root with `finishedAtMs > 0` and `childrenFinalized == true` as obsolete (i.e. return `true`), consistent with how the rest of the cancellation pipeline classifies that state as fully complete.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/DurableReconcileJobStore.java
```diff
@@ -2221,6 +2224,20 @@ private ReconcileCancellationMaintenanceService.CancellationCleanupResult cleanu
     return propagateDirectChildCancellation(loaded.record, request, childPageSize);
   }
 
+  private boolean isObsoleteCancellationCleanupRoot(
```
mrouvroy-floe, line 2222: [minor] **Duplicated obsolete-root detection logic risks silent divergence**

`isObsoleteCancellationCleanupRoot` (lines 2227-2239) repeats the same four conditions already checked by `cleanupCancellationRoot` before it throws `obsoleteMarker()` (lines 2217-2222): missing envelope, null record, account mismatch, and `!isCancellationState`.

Having the same predicate duplicated in two places invites drift — a future change to what "obsolete" means (e.g. adding a new terminal state) could be applied to one path but not the other, silently diverging the paused vs. non-paused cleanup behavior.

Fix: have `cleanupCancellationRoot` delegate to the new predicate directly — `if (isObsoleteCancellationCleanupRoot(request)) { throw ReconcileCancellationMaintenanceService.obsoleteMarker(); }` — so both paths share a single source of truth.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/queue/ReconcileCancellationMaintenanceService.java
```diff
@@ -99,6 +132,16 @@ private CancellationStats cleanupCancellationMarkers(long deadlineMs) {
           continue;
         }
         if (marker.request().paused()) {
+          if (isObsoleteCancellationRoot(marker.request())) {
```
mrouvroy-floe, line 135: [minor] **Paused markers now incur a store read on every maintenance tick**

Each paused marker now triggers `isObsoleteCancellationRoot` → `loadByAnyAccount` on every maintenance pass. Previously a paused marker was a free `continue` with no I/O.

A marker whose root is still `JS_CANCELLING` (waiting on delegated child cleanups) is valid but will generate a steady per-tick read until it is resumed via `requestCancellationCleanup` (see `DurableReconcileJobStore.java:2554`) or its root is GC'd. The 1s→5s tick change offsets this ~5x, so at expected marker volumes the impact is likely acceptable.

No immediate change required if paused markers are expected to be short-lived, but the read-amplification should be a conscious tradeoff. If long-lived paused markers are plausible, consider caching the obsolescence result or adding a staleness threshold before re-checking.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/queue/ReconcileCancellationMaintenanceService.java
mcusack-floe: Fixed. Obsolete paused markers that are successfully deleted now increment only obsolete_deleted/deleted, not paused. If the CAS delete loses, the marker is still counted as paused for this maintenance pass because this worker did not remove it. Added/updated coverage in ReconcileMaintenanceServicesTest to assert paused=0, obsoleteDeleted=1, deleted=1 for the deleted obsolete-paused path.

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/DurableReconcileJobStore.java
mcusack-floe: Fixed. cleanupCancellationRoot and the paused-marker obsolete check now share the same activeCancellationCleanupRoot helper, so missing roots, account mismatches, non-cancellation states, and completed JS_CANCELLED roots are classified from one predicate.

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/DurableReconcileJobStore.java
mcusack-floe: Fixed. Completed cancelled roots are now treated as obsolete for cancellation-cleanup markers when state is JS_CANCELLED, finishedAtMs > 0, and childrenFinalized is true. Added a regression test that creates a paused marker for a completed cancelled root and verifies cancellation maintenance deletes it.

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/queue/ReconcileCancellationMaintenanceService.java
mcusack-floe: Decision: no code change for now. The extra read is intentional so paused markers can be garbage-collected when their roots become obsolete/completed, and the maintenance tick reduction limits the steady-state cost. If paused markers prove long-lived or high-volume in practice, we can add a staleness threshold or cached obsolete-check result later.

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/DurableReconcileJobStore.java
```diff
@@ -2209,16 +2212,41 @@ private ReconcileCancellationMaintenanceService.CancellationCleanupResult cleanu
...
+        || cancellationCleanupRootComplete(loaded.record)) {
+      return null;
     }
-    return propagateDirectChildCancellation(loaded.record, request, childPageSize);
+    return loaded.record;
+  }
+
+  private boolean cancellationCleanupRootComplete(StoredReconcileJob root) {
+    return root != null
+        && "JS_CANCELLED".equals(blankToEmpty(root.state))
```
mrouvroy-floe, line 2247: [major] **`JS_CANCELLED` roots marked obsolete before retryable cleanup side-effects complete**

`cancellationCleanupRootComplete` treats any `JS_CANCELLED` root with `finishedAtMs > 0` and `childrenFinalized` as obsolete, causing paused markers to be deleted before `cleanupCancellationRoot` can retry pending side-effects.

This silently drops retryable work such as `cleanupAbandonedFullRescanStatsGenerationIfTerminal` when `deleteUnpublishedStatsGeneration` throws or returns `RETRYABLE_IN_PROGRESS`. The marker is removed on the next maintenance pass, so the retry never fires and the side-effect is silently skipped.

Fix: before treating a `JS_CANCELLED` root as obsolete, verify that all associated cleanup side-effects have reached a terminal (non-retryable) state, or retain the marker until `cleanupCancellationRoot` itself signals completion.

---

mcusack-floe: Fixed. cancellationCleanupRootComplete now treats a completed JS_CANCELLED root as obsolete only if it no longer needs abandoned full-rescan stats cleanup, via !needsAbandonedFullRescanStatsCleanup(root). That keeps the cancellation cleanup marker active while deleteUnpublishedStatsGeneration can still throw or return RETRYABLE_IN_PROGRESS, so cleanupCancellationRoot gets another pass to retry the side effect. Verified with the two retry/pending full-rescan stats cleanup tests plus the completed-root stale marker regression.

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/DurableReconcileJobStore.java
```diff
@@ -2209,16 +2212,42 @@ private ReconcileCancellationMaintenanceService.CancellationCleanupResult cleanu
...
     String accountId = blankToEmpty(request.accountId());
     String rootJobId = blankToEmpty(request.rootJobId());
     StoredEnvelope loaded = loadByAnyAccount(rootJobId).orElse(null);
     if (loaded == null
         || loaded.record == null
         || !accountId.equals(blankToEmpty(loaded.record.accountId))
-        || !isCancellationState(loaded.record.state)) {
-      throw ReconcileCancellationMaintenanceService.obsoleteMarker();
+        || !isCancellationState(loaded.record.state)
+        || cancellationCleanupRootComplete(loaded.record)) {
```
mrouvroy-floe, line 2239: [major] **Retryable stats cleanup marker dropped before side effects complete**

`activeCancellationCleanupRoot` now classifies a `JS_CANCELLED` root with `finishedAtMs > 0` and `childrenFinalized` as obsolete and deletes the cancellation cleanup marker. This happens before `cleanupCancellationRoot` has a chance to retry pending side effects.

The root already satisfies `cancellationCleanupRootComplete`, so when `cleanupAbandonedFullRescanStatsGenerationIfTerminal` returns `false` (e.g. because `deleteUnpublishedStatsGeneration` threw or returned `RETRYABLE_IN_PROGRESS`), the next maintenance pass treats the marker as obsolete rather than retrying the delete. This breaks full-rescan snapshot cancellation in the presence of transient failures.

Regression is surfaced by `DurableReconcileJobStoreTest.fullRescanSnapshotCancellationRetriesDraftStatsCleanupAfterTransientFailure` and `fullRescanSnapshotCancellationKeepsStatsCl...`.

Fix: ensure the obsolete-marker check only fires after all retryable side effects have been confirmed complete, or gate deletion on a stricter completion predicate that accounts for pending cleanup steps.

---

mcusack-floe: This is already fixed in the current PR head. `cancellationCleanupRootComplete` now also requires `!needsAbandonedFullRescanStatsCleanup(root)`, so completed JS_CANCELLED roots are not treated as obsolete while retryable full-rescan stats cleanup is still pending. The two cited regression tests pass with the current head commit.

mrouvroy-floe (approved):

PR #390, opened 2026-07-18
gc: stop CasBlobGc deleting live blobs when a pointer re-targets an old sha mid-pass

jchauzi-floe:
Implements both recommendations from eng-floe/core#1904 (CasBlobGc deleted live table blobs on staging, leaving permanently dangling pointers).

## The race

Table protos are content-addressed (`tables/<id>/table/<sha256>.pb`) and reconcile rewrites oscillate between recurring content states, so the same sha recurs. Three behaviors interacted:

1. `BaseResourceRepository.writeBlob`/`writeBlobStrictBytes` **skipped the S3 PUT** when the blob already existed with a matching etag — re-referencing an old blob never refreshed its `LastModified`.
2. `CasBlobGc.runForAccount` marks pointer roots **once per pass**, then sweeps unreferenced blobs whose age exceeds min-age anchored at pass start (`nowMs` frozen).
3. A pointer CAS that re-targets an old existing blob mid-pass (PUT skipped, `LastModified` stale) is invisible to both the mark and the age fence → live blob deleted → dangling pointer, unrecoverable once the sweep also collects the previously-live blob.

## Fix 1 — always PUT (`service/repo`)

`writeBlob` and `writeBlobStrictBytes` now PUT unconditionally; the post-write etag verification is unchanged. Combined with the existing pass-start-anchored fence this is provably safe: any blob re-referenced during a pass gets `LastModified > nowMs`, so `nowMs - lastModified < minAgeMs` always skips it that pass.

Chosen over a "touch on skip" (self-CopyObject with metadata replace) because the `BlobStore` SPI has no copy/touch primitive — adding one means new SPI surface plus a per-backend implementation, while the PUT path already exists on every backend and these are small metadata protos, so the cost is one small PUT per oscillating rewrite.

## Fix 2 — re-check the owning pointer before each delete (`service/gc`)

For pointer-rooted blob families the key encodes its owner. New `Keys.ownerPointerKeyForBlob` derives the owning pointer key from the blob key shape; `CasBlobGc.deleteUnreferenced` re-reads that pointer immediately before each delete and skips the blob it currently references. One extra pointer read per delete *candidate* only, and it shrinks the exposure window to milliseconds independently of fix 1 (belt and suspenders — either fix alone closes the observed race).

Derivable owners: account, catalog, namespace, table, view, connector, table-root (`root/current`), snapshot (by-id), constraints (by-snapshot), target-stats generation manifests (`targets-active`) and per-generation records.

Not derivable (documented on `ownerPointerKeyForBlob`, protected by fix 1 plus the existing chain-walk rooting/poisoning):
- root manifest pages (`.../root/manifest/<sha>.pb`) — referenced from `TableRoot` blob content, not by any pointer;
- per-target stats (`.../target-stats/<target>/<sha>.pb`) and file stats (`.../file-stats/<path>/<sha>.pb`) — their owning pointers live under `.../snapshots/<snapshot_id>/stats/...` and the snapshot id is not part of the blob key.

## Fix 3 — version-targeted delete + writer-side heal (`storage`, `service`)

Review found a residual TOCTOU inside a single delete candidate: both fences above are stale reads, so a writer landing a re-PUT + pointer CAS inside the GC's per-key HEAD→delete gap could still lose the blob. Closed by making the check and the act name the same immutable object:

- `BlobHeader` now carries the S3 object `versionId` observed at HEAD, and `CasBlobGc` deletes exactly that version via the new `BlobStore.delete(uri, versionId)`. A concurrent re-PUT mints a NEW version the targeted delete cannot touch, so the pointer stays resolvable in every interleaving. `InMemoryBlobStore` emulates an Enabled versioned bucket with store-wide unique, never-reused version ids (so delete-and-recreate cannot ABA a stale versioned delete).
- **Fail closed, never fall back.** GC deletes are gated on `BlobStore.supportsVersionedDeletes()`: when the store cannot guarantee immutable version identities, the pass collects NOTHING, warns once, and reports `deletesUnsupported` — surfaced by a new scheduler gauge (`floecat.service.gc.cas.delete_unsupported_accounts`) and by the account's backlog age, which keeps climbing. There is no unconditional-delete fallback anywhere in the GC path (a blank versionId on a capable store is also skipped, and `delete(uri, versionId)` rejects blank ids outright).
- On S3 the capability check is `GetBucketVersioning == Enabled`, cached with a 5-minute TTL (failures are cached as unsupported and retried after the TTL). **Suspended is unsafe too**, not just unversioned: a suspended bucket overwrites the `"null"` version in place, so even a nominally targeted delete could destroy a concurrent re-write. In an Enabled bucket a pre-versioning object's literal `"null"` versionId IS safe to target — the gate is on the bucket status, not the id's shape.
- **Deployment note:** the floecat runtime role needs `s3:GetBucketVersioning` on the blob bucket, and the bucket's versioning must be Enabled — otherwise GC fails closed forever (safe, but collects nothing; watch the new gauge). `docker/localstack-init.sh` now enables versioning on the dev/CI bucket so the safe delete path is actually exercised.
- SDK 2.44.4 does model conditional deletes (`ifMatch`/`ifMatchLastModifiedTime`/`ifMatchSize` on `DeleteObjectRequest`), but S3 supports those headers only on directory buckets — and `ifMatch` would not help anyway (the etag is the content sha, identical across the re-PUT). versionId targeting is the mechanism.
- **Behavior flag:** version-targeted deletes are HARD deletes (no delete marker), unlike the previous marker-stacking delete. Intended — real space reclaim — but it removes the accidental restore-from-marker net used during the incident remediation.
- **Writer-side backstop:** after a committed create/createIfAbsent/update pointer batch, `GenericResourceRepository` HEADs the canonical blob and re-PUTs it if missing (the writer still holds the bytes), logging a warn (the GC-race detection signal). Any residual hole becomes a transient blip, and a pre-existing dangling heals on the next write. One cheap HEAD per committed write. (It is a backstop only — it cannot replace the versioned delete, since a stale GC delete can land after the heal's HEAD.)

## Tests

- `CasBlobGcTest.aPointerRetargetedToAnOldExistingBlobMidPassIsNotSwept` — reproduces the staging interleave: pointer on B at mark time, CAS back to old blob A injected mid-pass (from the sweep's first blob LIST), min-age fence already passed for A. Fails without fix 2, passes with it; also asserts the next pass keeps A and collects B.
- `CasBlobGcTest.aPointerRetargetedBetweenTheGcHeadAndItsDeleteIsNotSwept` — the reviewer's narrower ordering: re-PUT + CAS injected between the GC's HEAD and its delete; only the version-targeted delete can refuse (verified to fail against an unconditional delete).
- `GenericResourceRepositoryUpdateTest.update_rewritesTheCanonicalBlobWhenGcSweptItBeforeTheCommit` — the writer-side heal re-PUTs a canonical blob swept between writeBlob and the pointer commit.
- `CasBlobGcTest.sweepFailsClosedWhenTheStoreCannotDeleteByVersion` / `aBlobWhoseHeaderLacksAVersionIdIsSkippedNotDeleted` — the fail-closed paths: capability off → nothing collected and the skip is reported; unnameable version → per-blob skip.
- `InMemoryBlobStoreVersioningTest` — monotonic versionId per put; a superseded version is not deletable; a stale pre-delete id cannot delete a recreated blob (ABA); blank version rejected; models Enabled.
- `S3BlobStoreTest` — head exposes/omits versionId; version-targeted delete names exactly the observed version; blank version rejected (no request sent); versioning gate Enabled/Suspended/none; lookup failure fails closed; status cached; closed-pool mapping for the new op.
- `BaseResourceRepositoryWriteBlobTest` — `putBlob`/`putBlobStrictBytes` must re-PUT when an identical blob already exists (pins fix 1 independently).
- `KeysTest.ownerPointerKeyForBlob*` — owner derivation for every pointer-rooted family (including percent-encoding round trip and unslashed LIST keys) and null for every non-derivable/malformed shape.

`mvn -pl core/storage-spi,storage/memory,storage/aws,service -am test`: all green.

### DIFF
--- a/core/proto/src/main/proto/floecat/common/common.proto
+++ b/core/proto/src/main/proto/floecat/common/common.proto
@@ -115,6 +115,9 @@ message BlobHeader {
   google.protobuf.Timestamp created_at = 5;
   google.protobuf.Timestamp last_modified_at = 6;
   repeated Tag tags = 7;
+  // Store-assigned version of the blob observed by this header (S3 object versionId), used for
+  // version-targeted deletes; empty when the store has no versioning.
+  string version_id = 8;
 }
 
 message Tag {
--- a/core/storage-spi/src/main/java/ai/floedb/floecat/storage/spi/BlobStore.java
+++ b/core/storage-spi/src/main/java/ai/floedb/floecat/storage/spi/BlobStore.java
@@ -32,17 +32,30 @@ public interface BlobStore {
   boolean delete(String uri);
 
   /**
+   * Whether {@link #delete(String, String)} acts on immutable version identities. {@code false}
+   * means a version-targeted delete cannot be trusted to leave a concurrent re-write intact — e.g.
+   * an S3 bucket whose versioning status is not {@code Enabled}: unversioned and suspended buckets
+   * overwrite the {@code "null"} version in place — so callers deleting for correctness must skip
+   * deleting instead. The default is {@code false}: fail closed.
+   */
+  default boolean supportsVersionedDeletes() {
+    return false;
+  }
+
+  /**
    * Deletes only the blob version that {@code versionId} names (as observed via {@link
    * BlobHeader#getVersionId()}). A different version — in particular one written concurrently after
    * the caller's {@link #head} — must survive, so a check-then-delete caller acts on exactly the
    * object it checked. Returns true when the named version was deleted (or already absent); false
    * when the store can tell the blob has moved past it and nothing was deleted.
    *
-   * <p>A blank {@code versionId} (store without versioning) and the default implementation both
-   * fall back to the unconditional {@link #delete(String)}; version-capable stores must override.
+   * <p>Callers must gate on {@link #supportsVersionedDeletes()}; a blank {@code versionId} is a
+   * caller bug and is rejected. There is deliberately NO fallback to the unconditional {@link
+   * #delete(String)} — that would silently reintroduce the delete-after-re-reference race this
+   * method exists to close.
    */
   default boolean delete(String uri, String versionId) {
-    return delete(uri);
+    throw new UnsupportedOperationException("versioned deletes not supported by this store");
   }
 
   void deletePrefix(String prefix);
--- a/core/storage-spi/src/main/java/ai/floedb/floecat/storage/spi/BlobStore.java
+++ b/core/storage-spi/src/main/java/ai/floedb/floecat/storage/spi/BlobStore.java
@@ -31,6 +31,20 @@ public interface BlobStore {
 
   boolean delete(String uri);
 
+  /**
+   * Deletes only the blob version that {@code versionId} names (as observed via {@link
+   * BlobHeader#getVersionId()}). A different version — in particular one written concurrently after
+   * the caller's {@link #head} — must survive, so a check-then-delete caller acts on exactly the
+   * object it checked. Returns true when the named version was deleted (or already absent); false
+   * when the store can tell the blob has moved past it and nothing was deleted.
+   *
+   * <p>A blank {@code versionId} (store without versioning) and the default implementation both
+   * fall back to the unconditional {@link #delete(String)}; version-capable stores must override.
+   */
+  default boolean delete(String uri, String versionId) {
+    return delete(uri);
+  }
+
   void deletePrefix(String prefix);
 
   default Map<String, byte[]> getBatch(List<String> uris) {
--- a/docker/localstack-init.sh
+++ b/docker/localstack-init.sh
@@ -6,6 +6,10 @@ BUCKET="${FLOECAT_BLOB_S3_BUCKET:-floecat-dev}"
 TABLE="${FLOECAT_KV_TABLE:-floecat_pointers}"
 
 awslocal s3api create-bucket --bucket "${BUCKET}" --region "${REGION}" >/dev/null 2>&1 || true
+# CAS blob GC fails closed (collects nothing) unless the blob bucket's versioning is Enabled:
+# version-targeted deletes are what let a delete race a concurrent re-reference safely.
+awslocal s3api put-bucket-versioning --bucket "${BUCKET}" \
+  --versioning-configuration Status=Enabled >/dev/null 2>&1 || true
 awslocal s3api create-bucket --bucket "bucket" --region "${REGION}" >/dev/null 2>&1 || true
 awslocal s3api create-bucket --bucket "warehouse" --region "${REGION}" >/dev/null 2>&1 || true
 awslocal s3api create-bucket --bucket "yb-iceberg-tpcds" --region "${REGION}" >/dev/null 2>&1 || true
--- a/service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
+++ b/service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
@@ -70,9 +70,22 @@ public class CasBlobGc {
       int blobsDeleted,
       int referenced,
       int tablesScanned,
-      boolean poisoned) {}
+      boolean poisoned,
+      boolean deletesUnsupported) {}
 
   public Result runForAccount(String accountId) {
+    if (!blobStore.supportsVersionedDeletes()) {
+      // Fail closed: without immutable version identities every delete is the
+      // eng-floe/core#1904 race (an S3 bucket whose versioning is not Enabled overwrites version
+      // state in place), so nothing may be collected. One warn per pass; the scheduler gauges
+      // this so a misconfigured bucket is noticed rather than silently never collecting.
+      LOG.warnf(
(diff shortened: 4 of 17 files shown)


mrouvroy-floe (approved):

jchauzi-floe:
Acknowledging the P1 from review: the pre-delete fences in `deleteUnreferenced` (age fence + owner re-check) were both stale reads followed by an unconditional `DeleteObject`, so a writer landing a re-PUT + pointer CAS inside the per-candidate HEAD→delete gap could still lose the blob — the original race, narrowed but not closed.

Closed now by making the check and the act name the same immutable object:

- **Version-targeted delete.** `BlobHeader` now carries the S3 object `versionId` observed at HEAD, and `CasBlobGc` deletes exactly that version via the new `BlobStore.delete(uri, versionId)`. A concurrent re-PUT mints a NEW version the targeted delete cannot touch, so the pointer stays resolvable in every interleaving. A blank versionId (unversioned bucket) degrades explicitly to the unconditional delete — GC race-safety on S3 therefore requires bucket versioning (staging and prod buckets are versioned). `InMemoryBlobStore` emulates versions with a per-key monotonic counter so the semantics are deterministically tested.
- **Writer-side backstop.** After a committed create/createIfAbsent/update pointer batch, the repository HEADs the canonical blob and re-PUTs it if missing (the writer still holds the bytes), logging a warn — any residual hole becomes a transient blip, and a pre-existing dangling heals on the next write.
- **Regression test for this exact ordering**: `CasBlobGcTest.aPointerRetargetedBetweenTheGcHeadAndItsDeleteIsNotSwept` injects the re-PUT + CAS between the GC's HEAD and its delete; verified to fail against the previous unconditional delete.

SDK note: SDK 2.44.4 does model conditional deletes (`ifMatch` / `ifMatchLastModifiedTime` / `ifMatchSize` on `DeleteObjectRequest`), but S3 supports those headers only on directory buckets — and `ifMatch` would not help here anyway (the etag is the content sha, identical across the re-PUT). versionId targeting is the mechanism.

One behavior flag: version-targeted deletes are HARD deletes (no delete marker), unlike the previous marker-stacking delete — intended (real space reclaim), but it removes the accidental restore-from-marker net we used during the incident.

jchauzi-floe:
Acknowledging both follow-up findings — both fixed.

**[P1] Fail closed unless bucket versioning is Enabled.** The blank-versionId → unconditional-delete fallback is gone from every layer:

- New `BlobStore.supportsVersionedDeletes()` capability (default false — fail closed). `CasBlobGc.runForAccount` gates on it: when the store cannot guarantee immutable version identities the pass collects **nothing**, warns once per pass, and reports a new `deletesUnsupported` result flag, surfaced by a scheduler gauge (`floecat.service.gc.cas.delete_unsupported_accounts`) and by the account's clean-sweep backlog age, which keeps climbing — a misconfigured prod bucket gets noticed instead of silently never collecting.
- `S3BlobStore` implements the capability as `GetBucketVersioning == Enabled`, cached with a 5-minute TTL (lookup failures are cached as unsupported and retried after the TTL). The gate is on the bucket **status**, not the versionId's shape: in an Enabled bucket a pre-versioning object's literal `"null"` id is safe to target (new PUTs mint fresh ids and retain it as noncurrent); in a Suspended bucket the `"null"` version is overwritten in place, so Suspended reports unsupported.
- `delete(uri, versionId)` now rejects blank ids outright (S3 and in-memory), and the SPI default throws rather than silently ignoring the version. A capable store returning a versionless header is skipped per blob.
- `docker/localstack-init.sh` now enables versioning on the blob bucket, so dev/CI exercises the real delete path.
- Deployment note (also in the PR body): the runtime role needs `s3:GetBucketVersioning` on the blob bucket — without it GC fails closed forever (safe, but collects nothing).
- Agreed on the heal: it stays a backstop (a stale GC delete can land after its HEAD); the versioned delete is the closure.

**[P2] In-memory version ids unique across delete-and-recreate.** `InMemoryBlobStore` now mints ids from a store-wide `AtomicLong` that survives deletes, so a delete-and-recreate can never resurrect an observed id — the ABA is impossible. Regression: `InMemoryBlobStoreVersioningTest.aStaleVersionIdCannotDeleteARecreatedBlob`. The store also reports `supportsVersionedDeletes() == true` (models an Enabled bucket), so all CasBlobGc tests exercise the real delete path.

New tests: `CasBlobGcTest.sweepFailsClosedWhenTheStoreCannotDeleteByVersion`, `CasBlobGcTest.aBlobWhoseHeaderLacksAVersionIdIsSkippedNotDeleted`, `S3BlobStoreTest.versionedDeletesAreSupportedOnlyWhileBucketVersioningIsEnabled` / `versioningStatusLookupFailureFailsClosed` / `versioningStatusIsCachedAcrossCalls` / `aBlankVersionIdIsRejectedNotDegradedToUnconditional` (+ in-memory equivalents).

PR #391, opened 2026-07-18
catalog: stop stripping engine hints on content-identical table updates

jchauzi-floe:
Partial fix for eng-floe/core#1905 (reconciler rewrites table protos on every cycle for unchanged sources). Companion core-side PR restores the floedb hint-clear policy wiring; this PR fixes the server-side layers.

## The loop being broken

Every PLAN_TABLE apply sends `UpdateTable` with mask `[schema_json, upstream, properties]`. `EngineHintSchemaCleaner` ran before `TableServiceImpl`'s no-op check and cleared on **mask presence** — so a content-identical apply stripped all `engine.hint.*` properties, forced a rewrite, and the next query's self-heal re-persisted the hints in **two** separate updates. Net: three pointer CAS ops + blob churn per reconcile-cycle-interleaved-with-queries, for static tables.

## service/catalog: diff-driven short-circuit

`EngineHintSchemaCleaner.cleanHints` now returns before consulting any extension when the relation shape is unchanged between `before` and `after` — for tables: `schema_json`, `upstream`, `display_name`, `namespace_id`; for views: everything except `properties`. An update that leaves the shape untouched cannot invalidate hints, so content-identical applies fall through to the existing `desired.equals(current)` no-op check.

The existing decision-routing tests keep their intent: the `runClean` harness now models an update whose schema actually changed.

## metagraph: combined relation+column hint persist

New `EngineHintPersistence.persistRelationAndColumnHints` (default: delegate to the two existing calls, so external implementations are unaffected) with an `EngineHintPersistenceImpl` override that applies the relation hint and all column hints in a **single** `tableRepository.update()` — one blob write + pointer CAS instead of two back-to-back. The caller flip in core's `FloeEngineSpecificDecorator.completeRelation` lands separately once this is merged and re-pinned (the decorator lives in eng-floe/core).

## Regression tests

- `EngineHintSchemaCleanerTest.unchangedShape_reconcileMaskKeepsAllHintsAndSkipsExtension` — the exact reconcile shape from the incident; red before this change.
- `EngineHintSchemaCleanerTest.changedSchema_reconcileMaskStillClears` / `changedDisplayName_doesNotShortCircuit` — clearing still happens on real changes.
- `EngineHintPersistenceImplTest.persistRelationAndColumnHints_singleRepositoryUpdate` / `_noopWhenAllPresent`.

`mvn -pl service -am test`: floecat-service 1380 run / 0 failures (1 pre-existing skip); all upstream modules green (reconciler 265, catalog 213, metagraph 16, …).

No overlap with #390 (CasBlobGc dangling-pointer fix) — different files; this PR reduces the pointer-oscillation churn that fed that race, but is independent of it.

## Relation to the noncurrent-version growth issue

#390's always-PUT fix means every legitimate rewrite mints a noncurrent object version; #392 added the `NoncurrentVersionExpiration` lifecycle rule (localstack init + `docs/storage-aws.md` deployment prerequisite) that reclaims them. This PR is the source-level half of the same story: it eliminates the oscillating rewrites that minted most versions in the first place, so the lifecycle rule only has to reclaim what legitimate rewrites still produce.

Follow-ups (out of scope here): flip the core decorator to the combined persist after re-pin; deterministic proto serialization for content-addressed blobs (map-order variance mints distinct shas for identical content).

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/metagraph/hint/EngineHintPersistenceImpl.java
+++ b/service/src/main/java/ai/floedb/floecat/service/metagraph/hint/EngineHintPersistenceImpl.java
@@ -160,6 +160,74 @@ public final class EngineHintPersistenceImpl implements EngineHintPersistence {
     }
   }
 
+  @Override
+  public void persistRelationAndColumnHints(
+      ResourceId relationId,
+      String relationPayloadType,
+      byte[] relationPayload,
+      String engineKind,
+      String engineVersion,
+      List<EngineHintPersistence.ColumnHint> columnHints) {
+    if (relationId == null) {
+      return;
+    }
+    String relationKey =
+        relationPayload == null ? null : EngineHintMetadata.tableHintKey(relationPayloadType);
+    String relationEncoded =
+        relationPayload == null
+            ? null
+            : EngineHintMetadata.encodeValue(engineKind, engineVersion, relationPayload);
+    Map<String, EngineHintPersistence.ColumnHint> deduped =
+        columnHints == null ? Map.of() : deduplicateColumnHints(columnHints);
+    if (relationKey == null && deduped.isEmpty()) {
+      return;
+    }
+    ResourceKind kind = relationId.getKind();
+    if (kind == ResourceKind.RK_TABLE) {
+      persistTable(
+          relationId,
+          builder -> {
+            boolean changed = false;
+            if (relationKey != null
+                && !relationEncoded.equals(builder.getPropertiesMap().get(relationKey))) {
+              builder.putProperties(relationKey, relationEncoded);
+              changed = true;
+            }
+            changed |=
+                applyColumnHints(
+                    deduped,
+                    builder.getPropertiesMap(),
+                    builder::putProperties,
+                    engineKind,
+                    engineVersion);
+            return changed;
+          });
+      return;
+    }
+    if (kind == ResourceKind.RK_VIEW) {
+      persistView(
+          relationId,
+          builder -> {
+            boolean changed = false;
+            if (relationKey != null
+                && !relationEncoded.equals(builder.getPropertiesMap().get(relationKey))) {
+              builder.putProperties(relationKey, relationEncoded);
+              changed = true;
+            }
+            changed |=
+                applyColumnHints(
+                    deduped,
+                    builder.getPropertiesMap(),
+                    builder::putProperties,
+                    engineKind,
+                    engineVersion);
+            return changed;
+          });
+      return;
+    }
+    LOG.debugf("Skipping engine hints for non-table relation %s", relationId);
+  }
+
   private boolean applyColumnHints(
       Map<String, EngineHintPersistence.ColumnHint> deduped,
       Map<String, String> properties,
--- a/service/src/main/java/ai/floedb/floecat/service/metagraph/hint/EngineHintPersistenceImpl.java
+++ b/service/src/main/java/ai/floedb/floecat/service/metagraph/hint/EngineHintPersistenceImpl.java
@@ -186,46 +186,49 @@ public final class EngineHintPersistenceImpl implements EngineHintPersistence {
     if (kind == ResourceKind.RK_TABLE) {
       persistTable(
           relationId,
-          builder -> {
-            boolean changed = false;
-            if (relationKey != null
-                && !relationEncoded.equals(builder.getPropertiesMap().get(relationKey))) {
-              builder.putProperties(relationKey, relationEncoded);
-              changed = true;
-            }
-            changed |=
-                applyColumnHints(
-                    deduped,
-                    builder.getPropertiesMap(),
-                    builder::putProperties,
-                    engineKind,
-                    engineVersion);
-            return changed;
-          });
+          builder ->
+              applyRelationAndColumnHints(
+                  relationKey,
+                  relationEncoded,
+                  deduped,
+                  builder.getPropertiesMap(),
+                  builder::putProperties,
+                  engineKind,
+                  engineVersion));
       return;
     }
     if (kind == ResourceKind.RK_VIEW) {
       persistView(
           relationId,
-          builder -> {
-            boolean changed = false;
-            if (relationKey != null
-                && !relationEncoded.equals(builder.getPropertiesMap().get(relationKey))) {
-              builder.putProperties(relationKey, relationEncoded);
-              changed = true;
-            }
-            changed |=
-                applyColumnHints(
-                    deduped,
-                    builder.getPropertiesMap(),
-                    builder::putProperties,
-                    engineKind,
-                    engineVersion);
-            return changed;
-          });
+          builder ->
+              applyRelationAndColumnHints(
+                  relationKey,
+                  relationEncoded,
+                  deduped,
+                  builder.getPropertiesMap(),
+                  builder::putProperties,
+                  engineKind,
+                  engineVersion));
       return;
     }
     LOG.debugf("Skipping engine hints for non-table relation %s", relationId);
+  }
+
+  private boolean applyRelationAndColumnHints(
+      String relationKey,
+      String relationEncoded,
+      Map<String, EngineHintPersistence.ColumnHint> deduped,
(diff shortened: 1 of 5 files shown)


mrouvroy-floe (commented):
**[major] Shape masks do not trigger hint clearing, and the new shape comparison is effectively unreachable** — `service/src/main/java/ai/floedb/floecat/service/catalog/hint/EngineHintSchemaCleaner.java:58`

`shouldClearHints` only returns `true` for `schema_json` and `upstream`, so updates to fields that the new `relationShapeUnchanged` logic treats as shape-changing — such as `display_name`, `namespace_id`, or view-specific fields like `sql_definitions`, `output_columns`, `base_relations`, and `creation_search_path` — never reach the new diff check at all. Stale engine hints survive writes that should have invalidated them.

For views the problem is compounded by an asymmetry in `relationShapeUnchanged` itself: the table branch uses an explicit allowlist of four fields, while the view branch compares everything except `properties`. This means the two branches fail in opposite directions when the proto grows a new field. The view branch also appears effectively unreachable from `TableServiceImpl` / `ViewServiceImpl` because `schema_json` and `upstream` are not valid `UpdateView` mask paths, while `EngineHintPersistenceImpl` calls `cleanHints` through a separate path.

Fix: expand the predicate in `shouldClearHints` to cover all fields that `relationShapeUnchanged` considers shape-relevant, and align the table and view comparison strategies (prefer the view-style "exclude non-shape fields" approach over a hardcoded allowlist so new proto fields are handled safely by default).

---

jchauzi-floe:
Verified all three claims against the code; fixed in 2813bd5f.

**Claim 1 — confirmed, with one framing correction: the gap is pre-existing, not introduced here.** `shouldClearHints` only fired on `schema_json`/`upstream` paths on main too, so `display_name`/`namespace_id`/`catalog_id`-masked updates never invoked the cleaner before this PR either — a rename has always left stale hints behind. This PR made the gap more visible (the diff check names those fields as shape) but didn't create it. Fixed here anyway, per your direction.

**Claim 2 — confirmed.** The table allowlist and view everything-except-properties comparison failed in opposite directions on proto growth.

**Claim 3 — half confirmed.** The view branch was indeed unreachable from the service path: `VIEW_MUTABLE_PATHS` contains no `schema_json`/`upstream`, so the old gate could never fire for a valid UpdateView mask — `cleanViewHints` was dead code (also pre-existing). The `EngineHintPersistenceImpl` part is not accurate though: it never calls `cleanHints` — the only call sites are `TableServiceImpl` and `ViewServiceImpl`, both gated by `shouldClearHints`.

**The fix (2813bd5f):**
- Single `NON_SHAPE_FIELDS` denylist (`properties`, `description`) drives both layers. The gate now fires for any mask path whose root is not non-shape — so `display_name`, `namespace_id`, `catalog_id`, and all view shape fields (`sql_definitions`, `output_columns`, `base_relations`, `creation_search_path`) reach the diff, and so does any mask path added in the future.
- Both diff branches now use the same comparison: strip `properties`/`description`/`created_at`, compare the rest. **A new proto field on either message defaults to shape-changing** (safe direction): a real change to it clears hints; an unchanged update still compares equal and short-circuits, so the churn fix this PR exists for is preserved — content-identical updates with ANY mask remain no-ops (covered by `displayNameMask_contentIdentical_keepsHintsAndSkipsExtension` and `viewContentIdentical_keepsHintsAndSkipsExtension`).

**New tests:** `shouldClearHints_firesForEveryShapePath`, `displayNameValueChange_withDisplayNameMask_clearsHints`, `namespaceValueChange_withNamespaceMask_clearsHints`, `displayNameMask_contentIdentical_keepsHintsAndSkipsExtension`, `viewSqlDefinitionChange_clearsHints`, `viewContentIdentical_keepsHintsAndSkipsExtension`. The view path is now genuinely reachable (gate includes view shape fields), so it's tested rather than documented as dead.

Full `mvn -pl service -am test`: 1385 run, 0 failures (1 pre-existing skip).

One observation out of scope for this PR: with the core-side policy wired (eng-floe/core#1906), a `catalog_id`-only change reaches the policy but `FloeHintClearPolicy.isRelationChange` doesn't compare `catalog_id` — whether a catalog move should invalidate relation hints is a policy question for its owners.

service/src/main/java/ai/floedb/floecat/service/metagraph/hint/EngineHintPersistenceImpl.java
```diff
@@ -160,6 +160,74 @@ public void persistColumnHints(
...
+        relationPayload == null
+            ? null
+            : EngineHintMetadata.encodeValue(engineKind, engineVersion, relationPayload);
+    Map<String, EngineHintPersistence.ColumnHint> deduped =
+        columnHints == null ? Map.of() : deduplicateColumnHints(columnHints);
+    if (relationKey == null && deduped.isEmpty()) {
+      return;
+    }
+    ResourceKind kind = relationId.getKind();
+    if (kind == ResourceKind.RK_TABLE) {
```
mrouvroy-floe, line 186: [minor] **`RK_TABLE` and `RK_VIEW` branches in `persistRelationAndColumnHints` are byte-identical lambdas**

The two `builder -> { ... }` modifier lambdas for the table and view branches are identical apart from the builder type — same relation-hint check, same `applyColumnHints` call, same `changed` accumulation.

Duplicated logic like this drifts: a future fix to the relation-hint comparison or column-hint application has to be made in two places, and it's easy to update only one.

Consider factoring the shared body into a small generic helper (e.g. a `BiFunction`/method that takes `getPropertiesMap`, `putProperties`, and returns `changed`) so both `persistTable` and `persistView` feed the same modifier. The existing `applyColumnHints` already takes the map + setter as parameters, so the relation-hint half can follow the same shape.

---

service/src/test/java/ai/floedb/floecat/service/metagraph/hint/EngineHintPersistenceImplTest.java
```diff
@@ -192,6 +192,52 @@ void persistColumnHints_writesViewProperties() {
     verify(cacheInvalidator).accept(VIEW_ID);
   }
 
+  @Test
```
mrouvroy-floe, line 195: [minor] **New `persistRelationAndColumnHints` paths only exercise the table branch**

The two new tests cover `RK_TABLE` with both a relation payload and column hints present. The `RK_VIEW` branch, the relation-only path (`columnHints` empty), and the columns-only path (`relationPayload == null`) are all untested.

These are exactly the branches the interface doc calls out as supported ("`relationPayload` may be null (columns only) and `columnHints` may be empty (relation only)"), and the view branch is a full copy of the table logic — so a regression there would go unnoticed. This method also has no in-repo caller yet (the core decorator flip lands separately), which makes tests the only guard until then.

Adding a `viewRepository`-backed case plus a relation-only and columns-only case would close the gap cheaply.

---

mrouvroy-floe (approved):

service/src/main/java/ai/floedb/floecat/service/metagraph/hint/EngineHintPersistenceImpl.java
jchauzi-floe: Done in 31f9d605 — both branches now feed a shared `applyRelationAndColumnHints(relationKey, relationEncoded, deduped, properties, setter, engineKind, engineVersion)` helper in the style of `applyColumnHints`; the RK_TABLE/RK_VIEW lambdas are one-liners.

service/src/test/java/ai/floedb/floecat/service/metagraph/hint/EngineHintPersistenceImplTest.java
jchauzi-floe: Done in 31f9d605 — added `persistRelationAndColumnHints_singleViewUpdate` (viewRepository-backed, asserts one update with both hint kinds and no table write), `persistRelationAndColumnHints_relationOnly` (empty columnHints), and `persistRelationAndColumnHints_columnsOnly` (null relationPayload), covering every mode the interface doc promises.

PR #392, opened 2026-07-18
gc: harden CAS sweep against the wider reachability-gap class 

mrouvroy-floe:
## Summary 

#390 closes the observed corruption (eng-floe/core#1904) — the content-addressed pointer flip-flop that let `CasBlobGc` delete a live table blob — via always-PUT (min-age fence works again), the per-candidate owner recheck, version-targeted deletes, and a fail-closed gate on non-versioned buckets. This follow-up hardens the sweep against the **broader class**: any gap between the referenced set and the truth (scan-visibility races, swallowed signals), where a wrong delete is permanent catalog corruption with no self-heal.

### What's here

- **`gc: harden the owner recheck — defer no-owner families, re-prove, isolate account failures`**
  - **Rescue, don't poison.** When a candidate's owner pointer still references it, the referenced set missed a live reference — keep the blob and count the rescue (`blobs-rescued` gauge, WARN), but do **not** poison or abort. A stale set can only cause false *keeps*: every owner-derivable candidate self-rechecks, every no-owner candidate is deferred and re-proven, so continuing is safe and lets the pass finish. (Poisoning on rescue let one table's persistent flip-flop starve the whole account's no-owner collection — the bug this replaces.)
  - **Deferred flush with per-table re-mark.** Families with no derivable owner (root manifest pages, per-target/file stats) are deferred (carrying the age-checked version) and deleted only after re-proving liveness per owning table against the **settled** store (root-chain re-walk + stats/constraints re-scan). Gated on `walkFailures` only; a per-table re-mark failure keeps that table's candidates without starving others.
  - **Fail-safe** on an underivable owner in a non-deferring pass (skip, never blind-delete).
  - **Scheduler isolates account failures.** `runForAccount` is wrapped so one account's storage fault — a transient SDK error, or 403 AccessDenied when the role lacks the now-required `s3:DeleteObjectVersion` — is logged, counted poisoned, and the tick advances instead of aborting.

- **`telemetry: register unsupported CAS delete accounts`** — surfaces #390's fail-closed gate (versioning disabled → GC skips) so a misconfigured bucket is observable, not silently never-collecting.

- **`storage/aws: require noncurrent version expiration`** — documents and locally provisions the deployment contract #390's always-PUT + versioning creates: versioning `Enabled` **and** a `NoncurrentVersionExpiration` lifecycle rule (else noncurrent versions of hot keys grow unbounded — the lifecycle rule, not GC code, is the right tool). Adds the missing `s3:DeleteObjectVersion` IAM prerequisite.

- **`docs: clarify Keys.ownerPointerKeyForBlob`** — the generation-scoped target-stats shape carries the snapshot id and *is* owner-derivable (inline recheck), unlike the snapshot-id-less shape (deferred). Doc-only.

## Type of Change

- [x] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
+++ b/service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
@@ -31,6 +31,7 @@ import jakarta.inject.Inject;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
@@ -51,6 +52,36 @@ import org.jboss.logging.Logger;
  * sweep on nodes serving queries ({@code floecat.gc.cas.enabled=false}) or share pin roots across
  * nodes first; the in-process context store already makes queries node-sticky, so this constraint
  * travels with the existing query-routing one.
+ *
+ * <p>Defense in depth: independent of how the referenced set was computed, the delete phase
+ * re-reads each candidate's OWNING pointer ({@link Keys#ownerPointerKeyForBlob}) immediately before
+ * deleting, and refuses — loudly — to delete a blob a live pointer still references. Such a rescue
+ * keeps the blob, counts {@code blobs-rescued}, and WARNs, then CONTINUES the pass — it does NOT
+ * poison the sweep. A stale referenced set can only cause false KEEPS here, never false deletes,
+ * because every owner-derivable candidate self-rechecks and every no-owner candidate is deferred
+ * and re-proven per table; poisoning the sweep on a rescue would instead let one table's persistent
+ * flip-flop starve the whole account's no-owner collection. A miss here is catalog data loss with
+ * no self-heal (reads fail with "dangling pointer, missing blob" and a reconcile re-references the
+ * dead URI). Deletes are version-targeted (the exact version the pass age-checked), and the sweep
+ * fails closed unless the store reports immutable version ids (S3 bucket versioning Enabled), so a
+ * concurrent re-PUT always survives as a new version the targeted delete cannot touch. Families
+ * with no owner pointer derivable from the key (manifest pages, per-target and file stats records)
+ * cannot be rescued individually, and lexicographic listing puts some of them before any blob whose
+ * rescue could reveal the stale set — so their deletion is DEFERRED, and the flush independently
+ * re-proves liveness per owning table against the settled store (root-chain re-walk plus
+ * constraints/stats pointer re-scan) before deleting. Superseded-but-pinned blobs are outside the
+ * recheck's reach (their owner pointer has moved on) and remain guarded by pin roots and
+ * walk-failure poisoning alone.
+ *
+ * <p><b>Versioned-bucket operations note.</b> This sweep only ever deletes the CURRENT version it
+ * age-checked; it never visits noncurrent versions. With the always-PUT write path, an oscillating
+ * live URI accrues a version per rewrite, and a garbage blob with N versions needs N sweep ticks to
+ * disappear (each targeted delete promotes the previous version to current). Neither is reclaimed
+ * fully by this GC: versioned deployments REQUIRE a bucket lifecycle rule expiring noncurrent
+ * versions (which also finishes off delete markers left by the unversioned {@code delete(key)}
+ * fallback, which reclaims nothing physically on a versioned bucket). This is the deliberate trade:
+ * versioning is what makes the delete race closable at all, and it doubles as the
+ * corruption-recovery mechanism (staging was healed by restoring deleted versions).
  */
 @ApplicationScoped
 public class CasBlobGc {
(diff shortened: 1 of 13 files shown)


service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
```diff
@@ -300,11 +349,94 @@ public Result runForAccount(String accountId) {
...
+        String tableId = entry.getKey();
+        if (tableId.isEmpty()) {
+          continue; // unparseable key shape: fail safe, leave it for a future pass
+        }
+        // Same per-unit pin cadence as the inline delete passes (which re-root once per listing
+        // page): a query that pins a superseded root while the flush is running must still
+        // protect that root's whole chain before this table's candidates are deleted. A PIN-chain
+        // walk failure is global — the referenced set itself is unprovable — so it aborts the
+        // whole flush, unlike a per-table re-mark failure below.
+        rootLivePinChains(referenced, walkedPinRoots, walkFailures);
```
mcusack-floe, line 400: **P1 — A late query pin can still lose deferred blobs.** This snapshots the live pins only once per table, then remarkTable and the loop below delete every deferred candidate. If a query publishes a pin to a superseded root after this call, that root is absent from both the referenced and fresh sets, so a manifest page reachable only through it can be deleted at line 415. Version-targeted deletion does not close this race because publishing a query pin is read-only and does not re-PUT the page. Please either recheck and expand pins immediately before each deferred delete, or coordinate pin publication atomically with GC so a newly published pin cannot fall into this window.

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
```diff
@@ -51,6 +52,33 @@
  * sweep on nodes serving queries ({@code floecat.gc.cas.enabled=false}) or share pin roots across
  * nodes first; the in-process context store already makes queries node-sticky, so this constraint
  * travels with the existing query-routing one.
+ *
+ * <p>Defense in depth: independent of how the referenced set was computed, the delete phase
+ * re-reads each candidate's OWNING pointer ({@link Keys#ownerPointerKeyForBlob}) immediately before
+ * deleting, and refuses — loudly — to delete a blob a live pointer still references. A rescue
+ * proves the referenced set is incomplete, so it also poisons the rest of the delete phase: every
```
mcusack-floe: **P3 — This class-level contract now contradicts the implementation.** The new behavior deliberately counts a rescue without poisoning or aborting the remaining delete phase, but this text still says a rescue poisons the rest of the phase. Please update the documentation so future maintenance does not restore the behavior this PR is removing.

kwesterfeld-floe (approved):

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
kwesterfeld-floe: **Javadoc contradicts the rescue-doesn't-poison behavior this PR implements.** This sentence says a rescue "poisons the rest of the delete phase," but the actual `ownedBy(...)` branch below (and its own comment) keeps the blob and explicitly does **not** poison or abort the pass — and `rescuesContentAddressedBlobTheScanSawUnderItsSupersededTarget` asserts `assertFalse(result.poisoned())`. The PR summary agrees ("Rescue, don't poison").

This class Javadoc is the primary safety-model explanation, so having it assert the opposite of the code will mislead the next maintainer. Looks like a leftover from the #390 design. Please reconcile it with the "rescue keeps, never poisons" model — the reason continuing is safe is that every owner-derivable candidate self-rechecks and every no-owner candidate is deferred and re-proven per-table.

storage/memory/src/main/java/ai/floedb/floecat/storage/memory/InMemoryBlobStore.java
```diff
@@ -130,6 +130,11 @@ public boolean delete(String uri, String versionId) {
       throw new IllegalArgumentException("versioned delete requires a versionId: " + uri);
     }
     final String k = normalize(uri);
+    if (!map.containsKey(k)) {
```
kwesterfeld-floe, line 141: **Asymmetric with the version-mismatch branch below.** This returns `true` when the key is wholly absent ("named version already absent counts as deleted"), but the `computeIfPresent` branch a few lines down returns `false` when the key is *present with a different current version* — where the named version is *also* absent. Two cases that are identical under the stated principle return opposite results.

It's defensible (this store models only the current version, so "present, different version" must signal "the current copy survived" for the residual-race probe), but the comment's stated principle doesn't match that reasoning. Suggest either broadening the comment to explain the single-version modeling, or making the semantics uniform. Low behavioral risk: in the GC flow `delete` is only ever called after a successful `head`, so this branch is effectively exercised only by the new unit test.

kwesterfeld-floe (commented):
Two items to address before merge (details inline). Nice work — the deferred-flush re-mark and the rescue-don't-poison reasoning hold up, and the test coverage against the real corruption interleave is excellent.

mcusack-floe (approved):

PR #393, opened 2026-07-20
GetUserObjects: add fast-path identity-only resolution

mrouvroy-floe:
## Summary

This PR adds an identity-only (conditional-request) fast path to the user-objects bundle: clients advertise known_table_blob_versions, and the server withholds the payload and returns early when it can prove the client already holds the exact served version. 

This is needed to validate objects in the planner cache still do exist while avoiding re-running the whole machinery of resolve + build + decoration.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist

- [x]  No credentials, tokens, or secrets are included

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
+++ b/service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
@@ -586,11 +586,16 @@ public class UserObjectBundleService {
    */
   private record SystemExecution(String backendKind, FlightEndpointRef flightEndpoint, String storagePath) {
     String tokenMaterial() {
-      return backendKind
-          + '\0'
-          + (flightEndpoint != null ? flightEndpoint.toString() : "")
-          + '\0'
-          + storagePath;
+      // Build from the endpoint's explicit, contractual fields (host/port/tls) rather than
+      // FlightEndpointRef.toString(): protobuf documents Message.toString() as non-contractual and
+      // subject to change, and this token is persisted by clients and matched across queries. The
+      // reserved `ticket` field is deliberately excluded — workers must not inspect it, and it is
+      // not routing identity. The token then moves exactly when the routing it covers moves.
+      String endpoint =
+          flightEndpoint != null
+              ? flightEndpoint.getHost() + ':' + flightEndpoint.getPort() + ':' + flightEndpoint.getTls()
+              : "";
+      return backendKind + '\0' + endpoint + '\0' + storagePath;
     }
   }
 
--- a/service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
+++ b/service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
@@ -551,24 +551,67 @@ public class UserObjectBundleService {
     // cacheIdentity alone would collide across all system tables — a client that cached one would
     // be served another identity-only under the shared token and reuse the wrong schema. Mixing the
     // id in makes the token unique per relation while still moving with engine content (the version
-    // changes on catalog upgrade). Note it does NOT cover a system table's service-resolved
-    // Flight/storage endpoint (configuredEndpointForKey), which is static Quarkus deployment config
-    // — a change to it requires a redeploy, which resets client caches, so a caller cannot route to
-    // a stale endpoint across the change. If endpoint config ever becomes hot-reloadable, fold it
-    // in.
+    // changes on catalog upgrade). It also folds in a system table's resolved EXECUTION metadata
+    // (backend kind + the resolved Flight/storage endpoint) — an identity-only reply omits that
+    // metadata, and a config-resolved endpoint (configuredEndpointForKey) can change without moving
+    // cacheIdentity. A floecat redeploy does NOT reset an external caching client, so without this
+    // the client would match the token, get no endpoint, and route to the stale one. The exact
+    // value served is resolved once in resolveSystemExecution and reused here, so the token can
+    // never drift from what buildRelation stamps.
     ResourceId relId = relation.relationId();
-    String keyMaterial =
-        relId.getAccountId()
-            + '\0'
-            + relId.getId()
-            + '\0'
-            + relId.getKindValue()
-            + '\0'
-            + cacheIdentity;
+    StringBuilder keyMaterial =
+        new StringBuilder()
+            .append(relId.getAccountId())
+            .append('\0')
+            .append(relId.getId())
+            .append('\0')
+            .append(relId.getKindValue())
+            .append('\0')
+            .append(cacheIdentity);
+    if (relation.node() instanceof SystemTableNode systemTableNode) {
+      keyMaterial.append('\0').append(resolveSystemExecution(systemTableNode).tokenMaterial());
+    }
     return Optional.of(
         RelationPinIdentity.newBuilder()
-            .setTableBlobVersion(Hashing.sha256Hex(keyMaterial))
+            .setTableBlobVersion(Hashing.sha256Hex(keyMaterial.toString()))
             .build());
+  }
+
+  /**
+   * A system table's resolved execution metadata: the backend kind plus the concrete endpoint the
+   * bundle serves (a Flight endpoint, whether built-in, node-declared, or config-resolved, or a
+   * storage-path fallback). Resolved in ONE place so buildRelation (which stamps these fields) and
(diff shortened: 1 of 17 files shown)


service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -490,13 +507,188 @@ private long decorationTotalNanos() {
...
+  private Optional<RelationPinIdentity> scopedPinIdentity(
+      String correlationId,
+      ResolvedRelation relation,
+      QueryContext queryContext,
+      EngineContext ctx) {
+    return pinIdentityFor(correlationId, relation, queryContext)
+        .map(
+            id ->
+                id.toBuilder()
+                    .setTableBlobVersion(possessionToken(id.getTableBlobVersion(), ctx))
```
mcusack-floe: **[P1] The possession token does not cover snapshot-backed schema changes.** This value ultimately comes from `TablePin.table_blob_version`, but the columns served by `logicalSchemaForRelation` come from the pinned snapshot's `schema_json`. `CreateSnapshot` and `UpdateSnapshot` can change that schema without changing the root's definition ref. A client holding the old definition token can therefore get an identity-only response for a new schema and reuse stale columns or types. Please derive the token from a persisted identity for the exact served schema, include the snapshot blob version (with the expected cache-fragmentation tradeoff), or enforce the claimed invariant by advancing the definition ref whenever snapshot schema changes.

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -711,6 +884,36 @@ private RelationInfo buildRelation(
...
+    // pinIdentityFor.
+    // Identity-only reuse of such a table therefore assumes its Flight endpoint config is
+    // deploy-time
+    // fixed; if it ever becomes hot-reloadable, fold the resolved endpoint into the token there.
+    if (servesFullSchema(relation.candidate())
+        && relationDecorationSucceeded
+        && countColumnsWithStatus(columnResults, ColumnStatus.COLUMN_STATUS_FAILED) == 0) {
+      scopedIdentity
+          .filter(id -> !id.getTableBlobVersion().isBlank())
+          .ifPresent(builder::setPinIdentity);
```
mcusack-floe: **[P1] This drops the entire pin identity for projected or incompletely decorated responses, not just the payload cache token.** Those responses consequently lose `pin_fingerprint`, snapshot identity, AS-OF provenance, and `constraints_ref_version`, even though the table was pinned. That prevents callers from distinguishing current and historical pins and conflicts with the `RelationInfo.pin_identity` contract. Please preserve the identity and leave only the possession token blank, or introduce a separate optional cache-token field.

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -490,13 +507,188 @@ private long decorationTotalNanos() {
...
+        relId.getAccountId()
+            + '\0'
+            + relId.getId()
+            + '\0'
+            + relId.getKindValue()
+            + '\0'
+            + cacheIdentity;
+    return Optional.of(
+        RelationPinIdentity.newBuilder()
+            .setTableBlobVersion(Hashing.sha256Hex(keyMaterial))
```
mcusack-floe: **[P2] The system-relation token does not cover the execution metadata omitted by a slim response.** `backend_kind`, the advertised Floecat endpoint, and endpoints resolved by `configuredEndpointForKey` can change while this hash remains identical. A service redeploy does not reset an external planner's cache, so the next request can match this token, omit the new endpoint, and make the planner reuse stale routing information. Please fold the resolved backend/endpoint payload into the token or keep system relations on the full-response path.

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -711,6 +884,36 @@ private RelationInfo buildRelation(
...
+    // and this stamp, so a cache miss under a populated hint set does not hash the relation twice.
+    //
+    // Caveat for system tables: a slim reply omits the endpoint metadata, and for a config-resolved
+    // endpoint (configuredEndpointForKey) that value is NOT covered by the token — see
+    // pinIdentityFor.
+    // Identity-only reuse of such a table therefore assumes its Flight endpoint config is
+    // deploy-time
+    // fixed; if it ever becomes hot-reloadable, fold the resolved endpoint into the token there.
+    if (servesFullSchema(relation.candidate())
+        && relationDecorationSucceeded
```
mcusack-floe: **[P2] `relationDecorationSucceeded` does not represent all payload decoration succeeding.** It is set to false when `decorateRelation` throws, but the `decorateView` catch only logs. A view whose column decoration succeeds after a transient view-decoration failure will still receive this token, and later identity-only responses will keep the incomplete cached view from self-healing. Please aggregate success across every payload-decoration phase (at minimum mark the `decorateView` failure here; also include `completeRelation` if it can mutate the served builder).

mcusack-floe (commented):
Reviewed the identity-only resolution path. I found four cache-correctness issues that should be addressed before relying on the fast path.

mcusack-floe (changes requested):

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
mrouvroy-floe: Confirmed and fixed — this was the real one. Went with your first option (derive from a persisted identity for the exact served schema) rather than the coarser snapshot-blob-version, so warmth survives data-only ingests.

SnapshotManifestEntry now carries a schema_fingerprint (sha256 of the snapshot's schema_json, sentinel for blank), stamped at all three manifest-entry writers where the Snapshot is already in hand (zero extra reads); it's copied onto TablePin at pin build and folded into the possession token in place of the bare definition version. A data-only ingest keeps the same fingerprint → token stable → identity-only reply; a snapshot-backed schema change moves it → full refetch.

Hardened three ways after your comment surfaced the mechanism: the fingerprint is over a canonicalized schema_json (object keys sorted, whitespace normalized, array/column order preserved) so a producer that re-serializes an unchanged schema doesn't churn it; pins from pre-fingerprint manifest entries fall back to snapshot_blob_version (correct, cold-on-ingest, self-heals); and preserveAuxRefs carries it through partial in-place rewrites like the other aux refs. 7f87f629c, 84fed369b, 6e96bcb48.

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
mrouvroy-floe: Fixed — took "preserve the identity, blank only the token." A separate cache-token field turned out unnecessary: pin_fingerprint already is the data-identity and deliberately excludes table_blob_version, so that field is already the dedicated possession-token slot. The stamp now always sets pin_identity (fingerprint, snapshot id, AS-OF provenance, constraints_ref_version all preserved) and blanks only table_blob_version when the reply is projected or decoration-incomplete. b816f6f77.

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
mrouvroy-floe: Fixed — folded the resolved execution metadata into the token. resolveSystemExecution (the same helper buildRelation uses to stamp the endpoint, so they can't drift) contributes backend_kind + the resolved Flight endpoint to keyMaterial, so a config change moves the token even though a redeploy doesn't reset an external planner's cache. Built from the endpoint's explicit host/port/tls fields rather than FlightEndpointRef.toString() (protobuf disclaims toString() stability, and this is persisted/matched across queries). b816f6f77, 16cf1fe0f.

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
mrouvroy-floe: Fixed — the decorateView catch now flips the success flag instead of only logging, and the stamp gate aggregates success across every payload-decoration phase (relation + view), so a transiently-incomplete view no longer receives a token that would keep the cached copy from self-healing. b816f6f77.

mcusack-floe (approved):

PR #394, opened 2026-07-20
fix: Propagate actionable storage authority failures through reconciliation

mcusack-floe:
## Summary

- Return a cataloged FAILED_PRECONDITION when credential vending lacks a storage authority.
- Decode structured Floecat errors so safe root-cause messages reach PLAN_TABLE logs and persisted failures.
- Make invalid requests and missing-authority failures terminal while preserving retryable preconditions.
- Log unexpected service failures with correlation IDs, bounded stack traces, and credential redaction.
- Add coverage for rendering, retry classification, submission, logging sanitization, and service errors.

## Type of Change

- [X] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist

### DIFF
--- a/reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReconcileFailureClassifier.java
+++ b/reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReconcileFailureClassifier.java
@@ -16,14 +16,19 @@
 
 package ai.floedb.floecat.reconciler.impl;
 
+import ai.floedb.floecat.common.rpc.Error;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
+import io.grpc.protobuf.StatusProto;
 import java.util.LinkedHashSet;
 import org.apache.iceberg.exceptions.ForbiddenException;
 import org.apache.iceberg.exceptions.NotAuthorizedException;
 import software.amazon.awssdk.awscore.exception.AwsServiceException;
 
 final class ReconcileFailureClassifier {
+  private static final String STORAGE_AUTHORITY_MISSING = "storage.authority.missing";
+  private static final String QUERY_SNAPSHOT_NOT_FINALIZED = "query.snapshot.not.finalized";
+
   private ReconcileFailureClassifier() {}
 
   static Exception normalize(Exception error) {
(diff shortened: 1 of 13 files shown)


service/src/main/java/ai/floedb/floecat/service/error/impl/GrpcErrors.java
```diff
@@ -353,6 +353,9 @@ public static StatusRuntimeException build(
 
   private static final int DEFAULT_MAX_DETAIL_CHARS = 512;
   private static final Pattern WHITESPACE = Pattern.compile("\\s+");
+  private static final Pattern SENSITIVE_LOG_VALUE =
```
mrouvroy-floe, line 356: [minor] **`sanitizeLogDetail` leaks the secret half of space-separated credential values**

The value-capture group `([^,;\s}\]]+)` stops at the first whitespace character, so only the token immediately after the separator is redacted.

For a header-style value such as `Authorization: Bearer eyJhbGci...` the regex matches `authorization`, the `: ` separator, and then `Bearer` — producing `authorization:[REDACTED] eyJhbGci...`, leaving the actual JWT/session token in the log. The same happens for `Authorization=Basic dXNlcjpwYXNz` (the base64 credentials survive).

The existing tests only cover `key=value` forms where the secret has no spaces, so this gap is untested. Consider extending the value class to consume the rest of the run (e.g. allow trailing tokens after a scheme word, or capture up to the next delimiter/newline rather than the next space), and add a `Bearer <token>` case to `BaseServiceImplTest`.

---

mrouvroy-floe (commented):
**[minor] Duplicate retry classification in the poller was not updated with the new gRPC-aware rules** — `reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteReconcileExecutorPoller.java:937`

This PR teaches `QueuedReconcileWorkerSupport.retryDispositionOf`/`retryClassOf` to inspect a `StatusRuntimeException` (INVALID_ARGUMENT → TERMINAL, `storage.authority.missing` → TERMINAL, `query.snapshot.not.finalized` → DEPENDENCY_NOT_READY). But `RemoteReconcileExecutorPoller` keeps its own copies of `retryDispositionOf`/`retryClassOf` (lines 937 and 951) that only handle `ReconcileFailureException` and otherwise default to `RETRYABLE`/`TRANSIENT_ERROR`.

The worker path is covered because credential-vending failures are caught inside `reconcileSingleTableTask`/`reconcileDiscoveryTableTask` and wrapped into a `Result` (so the new classification applies). However, any raw gRPC `StatusRuntimeException` that escapes `RemoteDefaultReconcileExecutor.execute()` as a *thrown* exception — e.g. from `getPlanTableInput`, `submitPlanTableFailure`, or `workerAuthorizationHeader` — hits the poller's `catch` block (around line 644) and is classified by these older methods, so an INVALID_ARGUMENT there is still treated as RETRYABLE.

If this transport-level path should remain RETRYABLE the divergence may be intentional, but it's worth either sharing a single classification helper or leaving a comment explaining why the two behave differently, to avoid the two copies drifting further.

---

service/src/main/java/ai/floedb/floecat/service/error/impl/GrpcErrors.java
mcusack-floe: Fixed in `fbb83d50b`. Sensitive values now consume through the next field delimiter or newline, so complete `Bearer <token>` and `Basic <credentials>` values are redacted. Added a `Bearer jwt-token` regression case to `BaseServiceImplTest`.

mcusack-floe:
Addressed the review note about duplicate retry classification in `fbb83d50b`. I moved the gRPC-aware retry rules into `ReconcileFailureClassifier`, and both `QueuedReconcileWorkerSupport` and `RemoteReconcileExecutorPoller` now delegate to it. Added coverage for wrapped `INVALID_ARGUMENT` being terminal and unknown `FAILED_PRECONDITION` remaining retryable/transient.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReconcileFailureClassifier.java
```diff
@@ -16,14 +16,19 @@
...
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
+import io.grpc.protobuf.StatusProto;
 import java.util.LinkedHashSet;
 import org.apache.iceberg.exceptions.ForbiddenException;
 import org.apache.iceberg.exceptions.NotAuthorizedException;
 import software.amazon.awssdk.awscore.exception.AwsServiceException;
 
 final class ReconcileFailureClassifier {
+  private static final String STORAGE_AUTHORITY_MISSING = "storage.authority.missing";
```
mrouvroy-floe, line 29: [minor] **Hardcoded message-key strings silently decouple from the service-side contract**

`STORAGE_AUTHORITY_MISSING` (`storage.authority.missing`) and `QUERY_SNAPSHOT_NOT_FINALIZED` (`query.snapshot.not.finalized`) are re-declared here as bare literals, while the authoritative source is the service module's `GeneratedErrorMessages.MessageKey` / `errors_en.properties`. There is no compile-time or test-time link between the two.

This is the load-bearing signal for terminal classification: if the service renames `storage.authority.missing`, `retryDisposition` falls through to `RETRYABLE`, so a genuinely terminal missing-authority failure would be retried indefinitely instead of being surfaced. The reconciler tests construct their own `Error` with the same literal string, so they wouldn't catch a rename either — the regression would be silent in production.

Consider referencing a shared constant, or adding a cross-module test that asserts the reconciler literal equals `GeneratedErrorMessages.MessageKey.STORAGE_AUTHORITY_MISSING.suffix()` so a rename fails the build.

---

mrouvroy-floe (commented):
**[minor] `sanitizeLogDetail` regex misses quoted-JSON and bare-token secrets** — `service/src/main/java/ai/floedb/floecat/service/error/impl/GrpcErrors.java:602`

`SENSITIVE_LOG_VALUE` only matches `key=value` / `key: value` pairs where the key is immediately followed by group 2's `\s*[:=]\s*`.

A JSON-shaped payload such as `"secretAccessKey":"wJalr..."` is not redacted, because the closing `"` sits between the key and the `:`, so group 2 never matches. Bare credential tokens with no adjacent label are likewise passed through. AWS SDK error text and nested config/STS dumps can arrive in either shape, so redaction here is best-effort rather than complete.

The `params.toString()` case (`{key=value}`) is covered, and over-redaction is the safe direction, so this isn't a blocker — but given the stated goal is credential redaction, it's worth either broadening the pattern to tolerate quotes around the key/value or documenting the limitation so it isn't mistaken for a guarantee.

---

PR #395, opened 2026-07-20
fix: refresh lease-scoped S3 credentials during reconciliation

mcusack-floe:
## Summary

- Reuse the existing refreshable credential-provider path for long-running Java connector captures.
- Propagate table scope so discovery jobs can securely renew credentials against their active lease.
- Validate discovery-table identity when vending renewed credentials.
- Log successful credential refreshes at INFO.
- Move routine lease/maintenance success logs to DEBUG.
- Add and update coverage for credential renewal and scoped vending behavior.

## Type of Change

- [X] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist


service/src/main/resources/application.properties
```diff
@@ -127,10 +127,10 @@ floecat.reconciler.job-store.lease-max-concurrency=${FLOECAT_RECONCILER_JOB_STOR
 floecat.reconciler.job-store.lease-acquire-timeout-ms=${FLOECAT_RECONCILER_JOB_STORE_LEASE_ACQUIRE_TIMEOUT_MS:250}
 floecat.reconciler.job-store.lease-scan-budget-ms=${FLOECAT_RECONCILER_JOB_STORE_LEASE_SCAN_BUDGET_MS:8000}
 floecat.reconciler.job-store.lease-maintenance.enabled=${FLOECAT_RECONCILER_JOB_STORE_LEASE_MAINTENANCE_ENABLED:true}
-floecat.reconciler.job-store.lease-maintenance.tick-every=${FLOECAT_RECONCILER_JOB_STORE_LEASE_MAINTENANCE_TICK_EVERY:1s}
+floecat.reconciler.job-store.lease-maintenance.tick-every=${FLOECAT_RECONCILER_JOB_STORE_LEASE_MAINTENANCE_TICK_EVERY:1m}
```
mrouvroy-floe: [major] **1m maintenance tick makes `reclaim-interval-ms` ineffective and slows orphaned-job recovery**

Expired-lease reclaim only runs when `ReconcileJobMaintenanceScheduler.tick()` fires. Inside `runLeaseMaintenanceOnce`, `reclaimExpiredLeasesIfDue` is gated by `reclaim-interval-ms` (default `5000`), but the gate can only *let work through as often as the scheduler ticks*. Dropping the tick from `1s` to `1m` means reclaim now runs at most once per minute instead of roughly every 5s.

The net effect is that the `reclaim-interval-ms=5000` tuning knob becomes a no-op, and worst-case recovery of an orphaned job (dead worker, `lease-ms=120000`) grows from ~125s to ~180s. The same coupling applies to `ready-index-maintenance.tick-every` and the ready-index repair path. Because this is a behavior/SLA change rather than a pure log-noise reduction, it deserves its own justification and ideally its own PR — the description frames it only as reduced cadence, not as intentionally rate-limiting reclaim.

If the intent is purely to cut log/scan volume, consider keeping the tick short and relying on `reclaim-interval-ms` for throttling, or explicitly raise `reclaim-interval-ms` so the effective cadence is documented in one place rather than emerging from the scheduler period.

---

core/connectors/common/src/main/java/ai/floedb/floecat/connector/common/auth/RefreshingAwsCredentialsProviderRegistry.java
```diff
@@ -165,7 +169,16 @@ private AwsCredentials resolveCredentials() {
           now = Instant.now();
           if (shouldRefresh(snapshot, refreshSkew, now)) {
             try {
-              current = Objects.requireNonNull(refresher.get(), "refresher returned null");
+              ResolvedStorageCredentials refreshed =
+                  Objects.requireNonNull(refresher.get(), "refresher returned null");
+              toAwsCredentials(refreshed);
```
mrouvroy-floe, line 185: [nit] **Discarded `toAwsCredentials(refreshed)` reads as dead code**

`toAwsCredentials(refreshed);` is called purely to validate that the refreshed credentials are convertible before publishing them to `current` (so a malformed refresh falls into the `catch` and keeps the still-valid snapshot). That intent is not obvious — the return value is dropped and a reader may flag it as a mistake or a future refactor may delete it.

A one-line comment (e.g. `// validate convertibility before publishing; keep prior creds on failure`) would make the defensive check self-documenting.

---

service/src/main/resources/application.properties
mcusack-floe: Addressed in f5e4dde8a. I restored both lease-maintenance and ready-index-maintenance scheduler defaults to 1s, including the annotation fallbacks and application.properties defaults, so reclaim-interval-ms remains effective and orphan recovery retains its prior cadence. The log-volume reduction remains in place independently: routine fast runs now log at DEBUG, while incomplete work, repairs, failures, or runs taking at least 30s log at INFO.

core/connectors/common/src/main/java/ai/floedb/floecat/connector/common/auth/RefreshingAwsCredentialsProviderRegistry.java
mcusack-floe: Addressed in f5e4dde8a. I added a comment explaining that the discarded conversion validates refreshed credentials before publishing them, allowing the catch path to retain still-valid prior credentials when a refresh returns malformed values.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ServerSideStorageConfigResolver.java
```diff
@@ -270,13 +264,17 @@ private ConnectorConfig resolve(
...
       String locationPrefix,
+      Optional<ResourceId> tableId,
       Optional<String> executionJobId,
       Optional<String> executionLeaseEpoch) {
     VendStorageCredentialsRequest.Builder request =
         VendStorageCredentialsRequest.newBuilder()
             .setAccountId(accountId)
             .setLocationPrefix(locationPrefix)
             .setUsage(StorageCredentialUsage.SCU_SERVER);
+    if (tableId != null && tableId.isPresent()) {
```
mrouvroy-floe: [nit] **Redundant null check on `Optional` `tableId`**

`if (tableId != null && tableId.isPresent())` mixes a null guard with an `Optional` presence check.

Every caller in this PR passes a non-null `Optional` (`Optional.of(...)`, `Optional.empty()`, or `Optional.ofNullable(...)`), and treating an `Optional` parameter as possibly-null is an anti-pattern that suggests the contract is unclear.

Either drop the null check (`if (tableId.isPresent())`) to match how `executionJobId`/`executionLeaseEpoch` are handled a few lines below, or if a null really is reachable, document why. Not blocking.

---

mrouvroy-floe (commented):
**[major] Lease-renewal DEBUG log still runs two unconditional store reads on the hot path** — `service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/NativeReconcileLeaseStore.java:1204`

`logLeaseRenewal` now logs only at DEBUG (the previous `EXEC_FILE_GROUP` INFO branch was removed), but it still calls `resolveJobKindForLeaseLog(accountId, jobId)` unconditionally on every successful renewal.

That helper performs two backing-store reads (`jobIndexStore.loadCanonicalSnapshot(...)` then `jobIndexStore.readRecord(...)`) plus a `String.format`, purely to build a message that is thrown away whenever DEBUG is disabled — the production default. Previously those reads at least fed a production-visible INFO line for `EXEC_FILE_GROUP` jobs; after this change they are pure overhead.

This matters because lease renewal is a high-frequency operation, and this PR increases renewal traffic further: the new credential-refresh path re-vends via `vendStorageCredentials`, which calls `renewLease` each time. Every renewal now pays for two store lookups whose only output is a discarded DEBUG string.

Guard the work behind the level check so the lookup and formatting only run when the log will actually be emitted:

```java
if (!LOG.isDebugEnabled()) {
  return;
}
String jobKind = resolveJobKindForLeaseLog(accountId, jobId);
...
LOG.debug(message);
```

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ServerSideStorageConfigResolver.java
mcusack-floe: Addressed in 5088badff. All callers provide a non-null Optional, so I removed the redundant null guard and kept the parameter contract consistent with the execution-binding Optionals.

mcusack-floe:
Addressed the lease-renewal hot-path review in 5088badff. `logLeaseRenewal` now returns immediately when DEBUG is disabled, before resolving the job kind or formatting the message, so successful renewals no longer perform the two diagnostic store reads in the production-default logging configuration.

mrouvroy-floe (commented):
**[minor] Described maintenance-cadence reduction (1s → 1m) is missing from the diff** — `service/src/main/resources/application.properties:130`

The PR description states: "Reduce lease and ready-index maintenance cadence from 1 second to 1 minute." No such change is present in this PR — the only files touched are `.java`, and the maintenance tick defaults are still `1s`.

`floecat.reconciler.job-store.lease-maintenance.tick-every` (line 130) and `floecat.reconciler.job-store.ready-index-maintenance.tick-every` (line 133) both still default to `1s`. The diff to `ReconcileLeaseMaintenanceService`/`ReconcileReadyIndexMaintenanceService` only changes when the per-run summary is logged at DEBUG vs INFO (threshold raised to 30s); it does not change how often the sweep runs.

Either an intended commit (bumping these two properties to `1m`) was dropped from the branch, or the description overstates the scope. Please confirm which — if the cadence reduction was intended, add the property changes; otherwise correct the description so reviewers/operators don't assume maintenance frequency (and its associated renewal/scan load) was reduced when it was not.

---

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityServiceImpl.java
```diff
@@ -564,6 +576,90 @@ private CredentialScope resolveLeaseScopedLocation(ReconcileJobStore.ReconcileJo
     return CredentialScope.forSingleLocation(locationPrefix);
   }
 
+  private CredentialScope resolveDiscoveryTableLocation(
```
mrouvroy-floe, line 575: [nit] **`resolveDiscoveryTableLocation` gates on blank `destinationTableId` but not `discoveryMode()`**

`resolveDiscoveryTableLocation` accepts any `PLAN_TABLE` job whose `tableTask.destinationTableId()` is blank, whereas the sibling `allowsConnectorBootstrapScope` additionally requires `job.tableTask.discoveryMode()`.

This is not exploitable — the method still requires the requested table's upstream connector/namespace/table to match the leased `tableTask` and throws `PERMISSION_DENIED` otherwise, so the worst case is vending credentials for the exact table the lease already authorizes. But the inconsistency makes the two scope-resolution paths harder to reason about.

Consider aligning the guard (add `&& job.tableTask.discoveryMode()`) so both discovery-scoped paths share the same precondition, or add a comment explaining why the blank-`destinationTableId` check alone is sufficient here.

---

mcusack-floe: Addressed. resolveDiscoveryTableLocation now explicitly requires tableTask.discoveryMode(), matching the connector-bootstrap path and failing closed for strict PLAN_TABLE tasks with a blank destination ID. I also added a regression test verifying that a strict unbound table task cannot use discovery-table scope. The code is committed locally and will appear in the next push.

mcusack-floe:
Addressed the maintenance-cadence review note by correcting the PR description. The inaccurate 1s-to-1m cadence bullet has been removed; lease and ready-index maintenance remain at 1s so reclaim-interval-ms and orphan recovery keep their intended behavior. The log-volume reduction remains limited to moving routine successful summaries to DEBUG.

mrouvroy-floe (approved):

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReconcilerService.java
```diff
@@ -280,101 +284,107 @@ public List<ReconcileSnapshotTask> planSnapshotTasks(
...
-          connector.enumerateSnapshots(
-              effectiveTask.sourceNamespace(),
-              effectiveTask.sourceTable(),
+    ReconcileContext ctx =
+        buildContext(
+            principal,
+            Optional.ofNullable(bearerToken),
+            Optional.ofNullable(executionJobId),
+            Optional.ofNullable(executionLeaseEpoch));
+    try (ActiveConnector active = activeConnectorForResult(ctx, connectorId)) {
```
mrouvroy-floe: [major] **Execution-bound connector-root vend in `planSnapshotTasks` may be scope-rejected and duplicates the table-scoped vend**

`planSnapshotTasks` now builds the context with the execution job/lease (lines 287-292), so the `activeConnectorForResult(ctx, connectorId)` call on line 293 vends credentials at the **connector source root** (`sourceStorageLocation(config)`, `tableId = Optional.empty()`) *with* an execution binding.

On the server, an execution-bound request routes through `resolveExecutionBoundLocation`. For a non-discovery job (`leasedTableId(job)` non-null), `resolveLeaseScopedLocation` returns the narrower *table* location, and `isWithinExecutionScope(connectorRoot, tableLocation)` computes `matchesLocationPrefix(connectorRoot, tableLocation)` — i.e. "does the connector root start with the table location" — which is `false` whenever the connector root is a strict parent of the table (the normal case for a warehouse-root filesystem connector, per the recent "Bootstrap file system table location from connector" work). That yields `PERMISSION_DENIED`, aborting snapshot planning.

Before this PR the same call built its context without an execution binding, so the server took the internal-resolve path and accepted the connector-root location. The discovery path is safe (it falls through to `resolvePlannerBootstrapLocation`, which matches the connector root), and a single-table connector where `storage_location == tableLocation` is also safe (equal-length match) — which is likely why the added unit tests don't surface it. But a non-discovery connector whose root is broader than the captured table would now fail.

Even when it does not fail, this connector-level vend is largely redundant: its result is only used as the base config for the immediately-following `tableScopedResolvedConfig` vend, yet it still issues a second `vendStorageCredentials` call (which calls `renewLease` server-side) and registers a refreshable provider that is discarded — extra renewal load of exactly the kind flagged in the prior review round.

Please confirm behavior against the localstack/docker-smoke integration tests (currently pending), and consider whether the connector-root vend here can be skipped (or made execution-unbound) when a table-scoped vend follows.

---

mcusack-floe: Addressed in 86ff3d55e. Snapshot planning now authenticates the connector/catalog without performing the execution-bound connector-root storage vend, then performs exactly one managed storage credential vend for the destination table ID and its resolved table storage location. This avoids the parent-root scope rejection for filesystem connectors and also handles catalog-based connectors that have a catalog URI but no connector-level S3 path. Added a regression test that verifies only one resolver call occurs and that it carries the table ID and table-scoped S3 location. ReconcilerServiceTest passes: 48 tests, 0 failures.

mrouvroy-floe (approved):
**[minor] Duplicated `DELTA`/`ICEBERG` credential-resolution blocks in `resolve`** — `reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ServerSideStorageConfigResolver.java:265`

The `DELTA` and `ICEBERG` branches of `resolve(...)` are now near-identical: each computes `locationPrefix`, calls `vendStorageCredentials(resolveRequest(...))`, and runs the same `mergeResolvedStorageConfig(...)` with the same `refreshExecutionBoundCredentials` supplier. This PR touched both blocks to thread `tableId` through, which is exactly the kind of change that is easy to apply to one branch and forget in the other.

The risk is future drift: because the refreshable-credential wiring is duplicated, a later fix to the refresh/merge logic can silently diverge between Delta and Iceberg connectors, producing subtly different vend behavior for the two kinds.

Consider extracting the shared body into a single helper (e.g. `resolveManagedForKind(kind, correlationId, authorizationToken, ..., tableId, connector, config)`) that both branches call, so the vend + merge + refresher construction lives in one place.

---

mcusack-floe:
Addressed in 00ea2df7a. I extracted the shared Delta/Iceberg storage credential vend, merge, and refresh-provider construction into resolveStorageBackedConnector, leaving the kind-specific gate in resolve so both connector types now use one implementation. I also corrected the stale snapshot-planning lifecycle test: an early existingSnapshotIds failure now asserts that no table credential vend occurs, matching the intentional removal of the connector-root vend. Validation: make fmt completed with no reformatted files; ReconcilerServiceInternalLogicTest, ReconcilerServiceTest, and ServerSideStorageConfigResolverTest passed (80 tests, 0 failures).

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ServerSideStorageConfigResolver.java
```diff
@@ -371,6 +363,13 @@ private ResolvedStorageCredentials refreshExecutionBoundCredentials(
                         + " session token, and expiresAt"));
   }
 
+  private static boolean isTerminalExecutionCredentialRefresh(StatusRuntimeException error) {
+    Status.Code code = error.getStatus().getCode();
```
mrouvroy-floe, line 367: [minor] **`FAILED_PRECONDITION` refresh rejection permanently poisons the provider**

`isTerminalExecutionCredentialRefresh` treats any `FAILED_PRECONDITION` (alongside `UNAUTHENTICATED`/`PERMISSION_DENIED`) as terminal. Once a refresh throws `TerminalCredentialRefreshException`, `RefreshingAwsCredentialsProviderRegistry.Entry` caches it in `terminalFailure` and every subsequent `resolveCredentials()` call re-throws it for the life of that provider entry — i.e. the whole capture is killed with no retry.

For genuine lease loss (`renewLease` returns false / job no longer active) this is exactly right. But `StorageAuthorityServiceImpl` also emits `FAILED_PRECONDITION` for "reconcile lease is not bound to a concrete storage location", which is a resolution-side condition rather than a definitive lease revocation. If that ever surfaces transiently mid-capture, an otherwise-valid long-running provider is poisoned permanently and the job fails terminally.

Consider narrowing the terminal set to the codes that unambiguously mean the lease is gone, or distinguishing lease-invalidity from location-resolution failures (e.g. a distinct status/detail) so only true lease loss becomes non-retryable. Blast radius is bounded to a single job attempt, hence minor.

---

mcusack-floe: Addressed in 031083d5d. Credential refresh no longer treats every FAILED_PRECONDITION as terminal. Genuine lease loss (lease renewal rejected, job missing, or job no longer active) is now emitted with the structured RECONCILE_LEASE_PRECONDITION_FAILED marker, and ServerSideStorageConfigResolver only caches a terminal refresh failure when that marker is present. The marker is also carried in the Floecat Error params so it survives the service error-mapping/localization path. Ordinary resolution-side FAILED_PRECONDITION responses, including an execution without a concrete storage location, remain non-terminal and do not poison the provider. UNAUTHENTICATED and PERMISSION_DENIED remain terminal. Added regression coverage on both sides: StorageAuthorityServiceImplTest verifies rejected leases are marked while the unbound-location error is not; ServerSideStorageConfigResolverTest verifies only the structured lease precondition is terminal. Focused results: 28 service tests and 47 reconciler tests passed.

mrouvroy-floe (commented):
**[minor] `planSnapshotTasks` is now dead production code but was extended in this PR** — `reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReconcilerService.java:267`

`RemoteDefaultReconcileExecutor.snapshotTasksForSuccessfulPlan` no longer calls `reconcilerService.planSnapshotTasks(...)`; it now builds tasks directly from `tableExecution.enumeratedSnapshotIds()`. A repo-wide search finds no remaining production caller of `planSnapshotTasks` — only tests (`ReconcilerServiceTest`, `ReconcilerServiceInternalLogicTest`) invoke it.

This matters because this same PR *extended* the orphaned method: it added the `executionJobId`/`executionLeaseEpoch` parameters and switched it to table-scoped credential vending (`tableScopedResolvedConfig`). That is effort spent — and future maintenance burden incurred — on a code path nothing exercises, and the surviving tests give a false impression that this planning logic is live.

There is also a subtle behavioral gap worth confirming: the old `planSnapshotTasks` applied `isSnapshotCaptureCompleteForScope` (capture-completeness) filtering before emitting tasks, whereas the new remote path emits a task for every enumerated snapshot id (`enumeratedSnapshotIds` is the metadata-pass bundle list, scope/mode-filtered but not completeness-filtered). This appears safe only because the downstream snapshot-capture step re-checks completeness, so already-captured snapshots become no-ops rather than duplicate work.

Either remove `planSnapshotTasks` and its now-dead tests, or, if it is being retained intentionally for a not-yet-wired path, add a comment saying so — and confirm the completeness-filtering difference above is acceptable.

---

mcusack-floe:
Addressed the latest review in 442b37b8b. I removed ReconcilerService.planSnapshotTasks, its dead-only helpers (filterPlannableSnapshotBundles and selectionSnapshotIds), the table-scoped connector-opening fallback that existed only for that path, and the tests that directly exercised the orphaned API. The production path is now the sole snapshot-job planning path. I also renamed the propagated value from enumeratedSnapshotIds to captureSnapshotIds to make its semantics explicit. The completeness-filtering concern was already handled in QueuedReconcileWorkerSupport: after the single connector enumeration, it filters IDs through isSnapshotCaptureCompleteForScope before returning TableExecutionResult; RemoteDefaultReconcileExecutor submits jobs only for those filtered IDs. I strengthened the live-path regression with snapshots 42 and 43: 43 is complete and excluded, while incomplete 42 is the only captureSnapshotId returned for job creation. Validation: make fmt completed; ReconcilerServiceTest, ReconcilerServiceInternalLogicTest, and RemoteDefaultReconcileExecutorTest passed (51 tests, 0 failures). Review: https://github.com/eng-floe/floecat/pull/395#pullrequestreview-4745786266

PR #398, opened 2026-07-21
Fix: Separate renewable catalog credentials from storage credentials

mcusack-floe:
## Summary

- Isolates Glue/Iceberg REST catalog authentication from S3 storage authentication.
- Adds catalog-scoped renewable AWS credential providers and SigV4 signing without overriding FileIO credentials.
- Preserves catalog credentials when table-scoped storage credentials are vended.
- Correctly manages borrowed versus owned credential-provider lifecycles.
- Updates Delta Glue authentication to use catalog-specific credentials.
- Moves Iceberg REST connectors to session-aware catalogs with explicit lifecycle handling.
- Retains configured storage credentials as a fallback while allowing table-vended credentials to override them.
- Adds regression coverage for credential separation, renewal, ownership, and deprecated SigV4 configuration.

## Type of Change

- [X] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist

### DIFF
--- a/connectors/catalogs/iceberg/src/main/java/ai/floedb/floecat/connector/iceberg/impl/IcebergConnectorFactory.java
+++ b/connectors/catalogs/iceberg/src/main/java/ai/floedb/floecat/connector/iceberg/impl/IcebergConnectorFactory.java
@@ -21,33 +21,31 @@ import ai.floedb.floecat.connector.common.auth.AwsProfileSupport;
 import ai.floedb.floecat.connector.common.auth.RefreshingAwsCredentialsProviderRegistry;
 import ai.floedb.floecat.connector.common.auth.RegistryBackedAwsCredentialsProvider;
 import ai.floedb.floecat.connector.spi.FloecatConnector;
-import java.time.Duration;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicBoolean;
+import org.apache.iceberg.CatalogProperties;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.catalog.Catalog;
+import org.apache.iceberg.catalog.SessionCatalog;
+import org.apache.iceberg.catalog.SupportsNamespaces;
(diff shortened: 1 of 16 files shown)


connectors/catalogs/iceberg/src/main/java/ai/floedb/floecat/connector/iceberg/impl/IcebergConnectorFactory.java
```diff
@@ -113,110 +111,80 @@ yield new IcebergFilesystemConnector(
...
-    for (Map.Entry<CatalogCacheKey, CatalogCacheEntry> entry : REST_CATALOG_CACHE.entrySet()) {
-      CatalogCacheEntry cached = entry.getValue();
-      if (!cached.isExpired(now)) {
-        continue;
-      }
-      if (REST_CATALOG_CACHE.remove(entry.getKey(), cached)) {
-        cached.retire();
-      }
-    }
+  private static RESTSessionCatalog createRestCatalog(Map<String, String> catalogProperties) {
```
mrouvroy-floe, line 164: [minor] **Removal of REST catalog cache adds a config round-trip per connector creation**

The previous implementation cached `RESTCatalog` instances (5-minute TTL, keyed by connection properties) and reused them across connector creations. This PR replaces that with `createRestCatalog`, which builds and `initialize()`s a fresh `RESTSessionCatalog` — including a new `HTTPClient` and the `GET /v1/config` handshake against the REST endpoint — every time `create()` is called.

Because connectors are built per `activeConnectorForResult` / capture (not pooled at a higher level that I could find), catalogs that are re-created frequently now pay an extra HTTP round-trip and lose HTTP connection reuse on each creation. Under load this can be a noticeable latency/connection-churn regression compared to the cached path.

This looks intentional — a shared cache would have served stale lease-scoped credentials, which is exactly what this PR is fixing, and per-instance lifecycle with `closeOnce` is cleaner. Please confirm the tradeoff is acceptable, and if catalog creation turns out to be hot, consider a credential-aware cache or reusing the catalog across the connector's own lifetime rather than dropping caching entirely. No code change required if the overhead is understood and acceptable.

---

mcusack-floe: Yes, this is intentional for now. Will revisit later, once the current authentication issues are resolved.

mrouvroy-floe (approved):

PR #399, opened 2026-07-22
fix: Prevent file-group work from starving reconcile control jobs

mcusack-floe:
## Summary

This PR improves reconcile scheduling under concurrent file-group and control-plane workloads.

- Reserves configurable worker capacity for planning and finalization so file-group execution cannot saturate the worker pool.
- Scans every eligible pinned-executor and job-kind bucket before reporting that no work is available.
- Rotates the first scanned bucket to preserve fairness between eligible queues.
- Reports lease-scan deadline exhaustion as a capacity error instead of an empty queue result.
- Adds configuration for reserved control slots and unit coverage for the new scheduling and leasing behavior.

## Type of Change

- [X] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist

### DIFF
--- a/reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteReconcileExecutorPoller.java
+++ b/reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteReconcileExecutorPoller.java
@@ -112,7 +112,7 @@ public class RemoteReconcileExecutorPoller {
         !hasControlExecutor || maxParallelism <= 1
             ? 0
             : Math.min(maxParallelism - 1, Math.max(0, configuredReservedControlSlots));
-    fileGroupSlots = new Semaphore(maxParallelism - reservedControlSlots, true);
+    fileGroupSlots = new Semaphore(maxParallelism - reservedControlSlots);
     workers = Executors.newFixedThreadPool(maxParallelism);
   }
 
@@ -275,6 +275,8 @@ public class RemoteReconcileExecutorPoller {
   }
 
   private static boolean isFileGroupOnly(ReconcileExecutor executor) {
+    // Capacity reservation relies on executor-scoped leases and a dedicated file-group executor.
+    // A mixed control/file-group executor must split those lease requests before it is supported.
     return executor != null
         && !executor.supportedJobKinds().isEmpty()
         && executor.supportedJobKinds().stream()
--- a/reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteReconcileExecutorPoller.java
+++ b/reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteReconcileExecutorPoller.java
@@ -201,6 +201,11 @@ public class RemoteReconcileExecutorPoller {
         lease = client.lease(request, workerLeaseSource());
       } catch (RuntimeException e) {
         releaseFileGroupSlot(fileGroupSlotReserved);
+        if (isLeaseScanBackpressure(e)) {
+          LOG.debugf(
+              "Skipping reconcile poll cycle due to lease scan backpressure: %s", e.getMessage());
+          return Optional.empty();
+        }
         if (shouldIgnoreStartupUnavailable(e)) {
           LOG.debugf(
               "Skipping local reconcile poll cycle until gRPC server is ready: %s", e.getMessage());
@@ -287,6 +292,11 @@ public class RemoteReconcileExecutorPoller {
       case LOCAL -> "local-poller";
       case REMOTE -> "remote-poller";
     };
+  }
+
+  static boolean isLeaseScanBackpressure(RuntimeException error) {
+    return error instanceof StatusRuntimeException statusError
+        && statusError.getStatus().getCode() == Status.Code.RESOURCE_EXHAUSTED;
   }
 
   private boolean shouldIgnoreStartupUnavailable(RuntimeException error) {
--- a/reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteReconcileExecutorPoller.java
+++ b/reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteReconcileExecutorPoller.java
@@ -16,6 +16,7 @@
 
 package ai.floedb.floecat.reconciler.impl;
 
+import ai.floedb.floecat.reconciler.jobs.ReconcileJobKind;
 import ai.floedb.floecat.reconciler.jobs.ReconcileJobStore;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
@@ -29,6 +30,7 @@ import java.util.Optional;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -46,6 +48,7 @@ public class RemoteReconcileExecutorPoller {
   private static final long MIN_CANCEL_CHECK_MS = 500L;
   private static final long HEARTBEAT_SHUTDOWN_WAIT_MS = 1_000L;
   private static final int DEFAULT_MAX_PARALLELISM = 1;
+  private static final int DEFAULT_RESERVED_CONTROL_SLOTS = 1;
 
   enum WorkerMode {
     LOCAL,
@@ -83,6 +86,8 @@ public class RemoteReconcileExecutorPoller {
   private final AtomicInteger leaseExecutorCursor = new AtomicInteger(0);
   private volatile WorkerMode workerMode = WorkerMode.LOCAL;
   private volatile int maxParallelism = DEFAULT_MAX_PARALLELISM;
+  private volatile int reservedControlSlots;
+  private volatile Semaphore fileGroupSlots;
   private volatile ExecutorService workers;
 
   @PostConstruct
@@ -96,6 +101,18 @@ public class RemoteReconcileExecutorPoller {
       maxParallelism = 0;
       return;
     }
+    boolean hasControlExecutor =
+        executorRegistry.orderedExecutors().stream()
+            .anyMatch(RemoteReconcileExecutorPoller::supportsControlWork);
+    int configuredReservedControlSlots =
+        config
+            .getOptionalValue("reconciler.reserved-control-slots", Integer.class)
+            .orElse(DEFAULT_RESERVED_CONTROL_SLOTS);
+    reservedControlSlots =
+        !hasControlExecutor || maxParallelism <= 1
+            ? 0
+            : Math.min(maxParallelism - 1, Math.max(0, configuredReservedControlSlots));
+    fileGroupSlots = new Semaphore(maxParallelism - reservedControlSlots, true);
     workers = Executors.newFixedThreadPool(maxParallelism);
   }
 
@@ -160,7 +177,7 @@ public class RemoteReconcileExecutorPoller {
   }
 
   private Optional<LeaseAssignment> leaseNextAssignment() {
-    var executors = executorRegistry.orderedExecutors();
+    List<ReconcileExecutor> executors = executorRegistry.orderedExecutors();
     if (executors.isEmpty()) {
       return Optional.empty();
     }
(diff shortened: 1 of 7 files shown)


service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/DurableReconcileJobStore.java
```diff
@@ -1191,7 +1191,16 @@ public Optional<LeasedJob> leaseNext(LeaseRequest request) {
             totalMs, scanStats.scanCount, scanStats.candidateCount, leased.isPresent());
       }
       observeLeaseNext(startedAtMs, scanStats, outcome);
+      if (scanStats.abortedByCaller) {
+        throw new CancellationException("reconcile lease scan cancelled by caller");
+      }
+      if (scanStats.abortedByDeadline) {
+        throw new LeaseScanCapacityExceededException(
```
mrouvroy-floe, line 1198: [minor] **Deadline abort surfaces as ERROR-level `worker task failed` log on the poller**

When a lease scan exhausts `leaseScanBudgetMs`, `leaseNext` now throws `LeaseScanCapacityExceededException` instead of returning empty. This maps to gRPC `RESOURCE_EXHAUSTED` in `ReconcileExecutorControlImpl.leaseNextOrThrowRateLimited`.

On the worker side, `RemoteReconcileExecutorPoller.shouldIgnoreStartupUnavailable` only special-cases `UNAVAILABLE`, so the `RESOURCE_EXHAUSTED` status propagates out of `leaseNextAssignment` and is caught by the worker task as a generic `RuntimeException`, logging `Remote reconcile worker task failed before outcome submission` at ERROR level. Under a sustained/large queue where scans routinely hit the budget, this is normal backpressure but reads as a task failure — every affected poll cycle emits a misleading ERROR that will trip log-based alerting.

Consider treating `RESOURCE_EXHAUSTED` like the startup-`UNAVAILABLE` case in the poller (skip the cycle quietly, or log at DEBUG/INFO with a backpressure-specific message) so the deadline path doesn't masquerade as a worker fault.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/NativeReconcileReadyQueueStore.java
```diff
@@ -328,64 +328,78 @@ private void deleteStaleReadyEntry(ReadyQueueEntry candidate) {
 
   private List<ReadyIndexSelection> readyScanSelections(LeaseRequest request) {
```
mrouvroy-floe, line 329: [minor] **Every `leaseNext` now scans all pinned + all job-kind buckets against a fixed scan budget**

`readyScanSelections` changed from one pinned bucket plus one unpinned bucket (2 slices) to every pinned-executor bucket plus every job-kind bucket. All selections share a single `LeaseScanStats` deadline, and each is scanned sequentially in `leaseReadyDue`.

The number of backend scan slices per lease call now scales with `pinnedExecutors + jobKinds` instead of being constant. On a busy deployment this multiplies backend I/O per call and makes it more likely a single call burns `leaseScanBudgetMs` before reaching the later buckets — which, combined with the change above, converts what used to be an empty result into a `RESOURCE_EXHAUSTED` rate-limit. That can reduce effective lease throughput, the opposite of the intended effect, when bucket cardinality is high.

This is the documented intent, so no change may be needed — but please confirm the pinned-executor and job-kind cardinality is bounded in production, and consider whether an early-exit once a lease is found (already present) plus a per-selection sub-budget would bound worst-case latency.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/DurableReconcileJobStore.java
mcusack-floe: Good catch. Fixed in c932760e2. The poller now treats RESOURCE_EXHAUSTED as expected lease-scan backpressure, logs it at DEBUG, and skips the current poll cycle instead of reporting a worker-task failure. The focused poller test suite passes with 22 tests.

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/NativeReconcileReadyQueueStore.java
mcusack-floe: Confirmed that the cardinality is bounded in the current production path because the poller issues executor-scoped lease requests. Each request contains one executor ID and at most two job kinds, so a lease call scans at most three slices. The shared scan budget already bounds total call latency, and scanning exits immediately when a lease is found. Given those bounds, I do not think a per-selection sub-budget is needed here.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteReconcileExecutorPoller.java
```diff
@@ -234,13 +268,37 @@ private boolean reserveWorkerSlot() {
     }
   }
 
+  private static boolean supportsControlWork(ReconcileExecutor executor) {
+    return executor != null
+        && executor.supportedJobKinds().stream()
+            .anyMatch(jobKind -> jobKind != ReconcileJobKind.EXEC_FILE_GROUP);
+  }
+
+  private static boolean isFileGroupOnly(ReconcileExecutor executor) {
```
mrouvroy-floe, line 277: [minor] **Reservation silently no-ops for a mixed-kind executor**

The throttle only engages when `isFileGroupOnly()` is true, i.e. an executor whose supported kinds are *exclusively* `EXEC_FILE_GROUP`. Any executor that supports `EXEC_FILE_GROUP` alongside a control kind returns `false` here, so its leases never acquire a `fileGroupSlots` permit.

That means such an executor's file-group work would not count against the reserved-capacity budget and could once again saturate the worker pool — the exact behavior this PR is trying to prevent. Today it's safe because `RemoteFileGroupReconcileExecutor` is the only `EXEC_FILE_GROUP` executor and it is file-group-only, so the gap is latent, not live.

Consider gating on "the leased job kind is `EXEC_FILE_GROUP`" rather than "the executor is file-group-only," or at minimum add a comment documenting the assumption so a future mixed-kind executor doesn't silently defeat the reservation.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteReconcileExecutorPoller.java
```diff
@@ -96,6 +101,18 @@ void init() {
...
+            .anyMatch(RemoteReconcileExecutorPoller::supportsControlWork);
+    int configuredReservedControlSlots =
+        config
+            .getOptionalValue("reconciler.reserved-control-slots", Integer.class)
+            .orElse(DEFAULT_RESERVED_CONTROL_SLOTS);
+    reservedControlSlots =
+        !hasControlExecutor || maxParallelism <= 1
+            ? 0
+            : Math.min(maxParallelism - 1, Math.max(0, configuredReservedControlSlots));
+    fileGroupSlots = new Semaphore(maxParallelism - reservedControlSlots, true);
```
mrouvroy-floe: [nit] **Fair `Semaphore` flag has no effect with barging `tryAcquire()`**

The semaphore is constructed as fair (`new Semaphore(..., true)`), but the only acquisition path is the no-timeout `slots.tryAcquire()` in `leaseNextAssignment`, which barges regardless of the fairness setting (per `Semaphore` javadoc).

The `true` therefore buys nothing at runtime and is slightly misleading to a reader who might assume acquisition order is guaranteed.

Drop the fairness argument (or switch to a fairness-honoring acquire) so the code reflects the actual behavior.

```suggestion
    fileGroupSlots = new Semaphore(maxParallelism - reservedControlSlots);
```

---

mcusack-floe: Fixed in 459252d1e. The semaphore now uses the default non-fair constructor, matching the untimed tryAcquire behavior. The focused poller suite passes with 22 tests, and make fmt completes cleanly.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteReconcileExecutorPoller.java
mcusack-floe: Agreed that this assumption should be explicit. Documented it in 459252d1e: capacity reservation relies on executor-scoped leases and the dedicated file-group executor. A future mixed control/file-group executor must split those lease requests before it can be supported safely.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteReconcileExecutorPoller.java
```diff
@@ -183,19 +214,22 @@ private Optional<LeaseAssignment> leaseNextAssignment() {
...
       Optional<ReconcileExecutor> executor = executorRegistry.executorFor(lease.get().lease());
       if (executor.isEmpty()) {
+        releaseFileGroupSlot(fileGroupSlotReserved);
         LOG.warnf(
             "Leased reconcile job jobId=%s kind=%s but no eligible executor matched locally",
             lease.get().lease().jobId, lease.get().lease().jobKind);
         continue;
       }
-      return Optional.of(new LeaseAssignment(executor.get(), lease.get()));
+      return Optional.of(new LeaseAssignment(executor.get(), lease.get(), fileGroupSlotReserved));
```
mrouvroy-floe, line 232: [minor] **Remap path reuses the requested executor's `fileGroupSlotReserved`**

When the leased job does not match the requested executor, the code remaps it to a different executor via `executorRegistry.executorFor(...)` but returns a `LeaseAssignment` that keeps the `fileGroupSlotReserved` flag computed from the *requested* executor.

This makes the reservation accounting inconsistent with the executor that actually runs the job. If the requested executor was control (`fileGroupSlotReserved=false`) but the remapped executor is file-group-only, a file-group job runs without holding a file-group permit — bypassing the very reservation this PR adds. The reverse (a control job holding a file-group permit) wastes file-group capacity. This is guarded in practice by the executor-scoped-lease assumption documented in `isFileGroupOnly`, so it is low-probability, but the remap path defeats that assumption by definition.

Consider recomputing the flag against the executor that will actually run: reserve/release based on `isFileGroupOnly(executor.get())` on this branch rather than propagating the requested executor's value.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteReconcileExecutorPoller.java
```diff
@@ -234,13 +268,39 @@ private boolean reserveWorkerSlot() {
     }
   }
 
+  private static boolean supportsControlWork(ReconcileExecutor executor) {
+    return executor != null
+        && executor.supportedJobKinds().stream()
```
mrouvroy-floe, line 273: [minor] **`supportedJobKinds()` dereferenced without null guard**

`supportsControlWork` (called for every executor at `init`) and `isFileGroupOnly` (called on every lease attempt) call `executor.supportedJobKinds().stream()` after only null-checking `executor` itself.

`ReconcileExecutorRegistry.leaseRequestFor` explicitly treats `supportedJobKinds() == null` as a real case, so the codebase already assumes an implementation may return null. If one ever does, `supportsControlWork` would NPE at startup and `isFileGroupOnly` on the hot lease path. All current implementations return non-null and the interface default is non-null, so this is defensive-consistency rather than an observed failure.

Mirror the existing guard from `leaseRequestFor` (treat a null result as an empty set) so the new helpers agree with the contract the rest of the code already codes to.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteReconcileExecutorPoller.java
mcusack-floe: The reservation is intentionally based on the executor-scoped lease request, because the permit must be acquired before taking a lease. The request constrains job kinds, so a control-only request cannot return file-group work and vice versa; remapping does not change the leased job kind. A mismatch would require the unsupported mixed-kind executor configuration documented below. Recomputing after the lease would also leave us needing to handle permit acquisition failure after the job is already leased, so I’d like to keep the current accounting.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteReconcileExecutorPoller.java
mcusack-floe: All executor implementations and the interface default return non-null capability sets. Guarding only these helpers would not make a null-returning executor safe: supportsJobKind() and registry matching would still dereference the result, while leaseRequestFor() would translate null to an unrestricted job-kind request. 

I’d prefer to retain the non-null contract rather than add partial normalization here.

PR #400, opened 2026-07-22
catalog: map iceberg and delta list columns to variant

jincongho-floe:
## Summary

List columns surface as the variant logical type instead of ARRAY, for both table formats — the scan-side list -> variant encoding in the engine is format-agnostic (iceberg and delta parquet files flow through the same reader chain), so the catalog mapping must be too.

IcebergTypeMappings maps LIST -> VARIANT and DeltaSchemaMapper / DeltaTypeMapper map arrays -> VARIANT. Both schema mappers emit one opaque leaf column per list; element struct fields stay inside the variant payload (reachable via variant subscripting) instead of expanding into arr[].field child columns. Stats behavior is unchanged: VARIANT, like ARRAY, is not stats-orderable.

## Type of Change

- [ x] feat (new functionality)
- [ ] fix (bug fix)

## Testing

- [ ] `make fmt`
- [ ] `make verify`

- [ ] No breaking changes

## Checklist

## Additional Notes

### DIFF
--- a/connectors/catalogs/iceberg/src/main/java/ai/floedb/floecat/connector/iceberg/impl/IcebergTypeMapper.java
+++ b/connectors/catalogs/iceberg/src/main/java/ai/floedb/floecat/connector/iceberg/impl/IcebergTypeMapper.java
@@ -35,7 +35,7 @@ import org.apache.iceberg.types.Type;
  * <p><b>Integer collapsing:</b> Both Iceberg {@code INTEGER} (32-bit) and {@code LONG} (64-bit)
  * collapse to canonical {@link ai.floedb.floecat.types.LogicalKind#INT} (64-bit).
  *
- * <p><b>Complex types:</b> {@code LIST} → {@link ai.floedb.floecat.types.LogicalKind#ARRAY}, {@code
+ * <p><b>Complex types:</b> {@code LIST} → {@link ai.floedb.floecat.types.LogicalKind#VARIANT} (lists surface as opaque variant columns), {@code
  * MAP} → {@link ai.floedb.floecat.types.LogicalKind#MAP}, {@code STRUCT} → {@link
  * ai.floedb.floecat.types.LogicalKind#STRUCT}. Element/value/field types are captured by child
  * {@code SchemaColumn} rows with their own paths.
--- a/core/connectors/common/src/main/java/ai/floedb/floecat/connector/common/resolver/IcebergTypeMappings.java
+++ b/core/connectors/common/src/main/java/ai/floedb/floecat/connector/common/resolver/IcebergTypeMappings.java
@@ -50,7 +50,10 @@ public final class IcebergTypeMappings {
       case STRING -> LogicalType.of(LogicalKind.STRING);
       case FIXED, BINARY -> LogicalType.of(LogicalKind.BINARY);
       case UUID -> LogicalType.of(LogicalKind.UUID);
-      case LIST -> LogicalType.of(LogicalKind.ARRAY);
+      // Lists surface as variant: the scan encodes list<T> rows as variant
+      // payloads, so the engine sees one opaque variant column regardless of
+      // element type or nesting depth (docs/sql/list_as_variant.md in core).
+      case LIST -> LogicalType.of(LogicalKind.VARIANT);
       case MAP -> LogicalType.of(LogicalKind.MAP);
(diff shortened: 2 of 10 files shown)


connectors/catalogs/iceberg/src/main/java/ai/floedb/floecat/connector/iceberg/impl/IcebergTypeMapper.java
```diff
@@ -35,7 +35,7 @@
  * <p><b>Integer collapsing:</b> Both Iceberg {@code INTEGER} (32-bit) and {@code LONG} (64-bit)
  * collapse to canonical {@link ai.floedb.floecat.types.LogicalKind#INT} (64-bit).
  *
- * <p><b>Complex types:</b> {@code LIST} → {@link ai.floedb.floecat.types.LogicalKind#ARRAY}, {@code
+ * <p><b>Complex types:</b> {@code LIST} → {@link ai.floedb.floecat.types.LogicalKind#VARIANT} (lists surface as opaque variant columns), {@code
  * MAP} → {@link ai.floedb.floecat.types.LogicalKind#MAP}, {@code STRUCT} → {@link
  * ai.floedb.floecat.types.LogicalKind#STRUCT}. Element/value/field types are captured by child
```
mrouvroy-floe, line 40: [minor] **Stale javadoc: list element fields no longer become child rows**

This sentence still claims "Element/value/field types are captured by child `SchemaColumn` rows with their own paths."

After this PR that is no longer true for lists — `IcebergSchemaMapper` no longer expands `list<struct>` into `arr[].field` child columns; only `struct` fields and `map` value structs are expanded. Leaving the text as-is contradicts the new `LIST -> VARIANT` behavior described two lines above.

Update the wording to note that list element fields stay inside the opaque variant payload and are not emitted as child columns.

---

core/connectors/common/src/main/java/ai/floedb/floecat/connector/common/resolver/IcebergTypeMappings.java
```diff
@@ -50,7 +50,10 @@ public static LogicalType toLogical(Type t) {
       case STRING -> LogicalType.of(LogicalKind.STRING);
       case FIXED, BINARY -> LogicalType.of(LogicalKind.BINARY);
       case UUID -> LogicalType.of(LogicalKind.UUID);
-      case LIST -> LogicalType.of(LogicalKind.ARRAY);
+      // Lists surface as variant: the scan encodes list<T> rows as variant
+      // payloads, so the engine sees one opaque variant column regardless of
+      // element type or nesting depth (docs/sql/list_as_variant.md in core).
```
mrouvroy-floe, line 55: [nit] **Comment references a doc not present in this repo**

Several new comments (here, and in `DeltaSchemaMapper` / `IcebergTypeMapper`) cite `docs/sql/list_as_variant.md in core` as the source of truth for the list-as-variant behavior.

No such file exists in this repository (there is no `docs/sql/` directory), so it presumably lives in a separate engine repo. As written, a reader in this repo can't follow the pointer.

Consider making the reference unambiguous (e.g. name the actual repo, or link the PR/commit that establishes the encoding) so the cross-repo contract is traceable.

---

mrouvroy-floe (commented):
**[minor] View schema mapping still maps `list -> ARRAY`** — `protocol-gateway/iceberg-rest/src/main/java/ai/floedb/floecat/gateway/iceberg/rest/services/view/ViewMetadataService.java:405`

This iceberg-REST view type mapping still emits `"list" -> "ARRAY"`, while every table-side schema/type mapper in this PR now maps lists to `VARIANT`.

The PR's stated rationale is that the list-as-variant encoding is format-agnostic because it happens in the engine's reader chain. If that reasoning extends to view columns, a view exposing a list would advertise `ARRAY` here while a table with the same column advertises `VARIANT`, which is an inconsistency clients could observe.

Please confirm whether views are intentionally out of scope. If they are, a one-line comment here noting the deliberate divergence would prevent this from looking like a missed spot; if not, this mapping should change to `VARIANT` too.

---

mrouvroy-floe (changes requested):
I’m not sure this belongs in OSS Floecat. 
It encodes a Floe-specific  detail and makes source arrays indistinguishable from native VARIANT values while discarding nested type information.
Let's discuss this offline a bit more ?

PR #401, opened 2026-07-22
fix: Prevent file-group work from starving reconcile control jobs (#399)

mcusack-floe:
## Summary

This PR improves reconcile scheduling under concurrent file-group and control-plane workloads.

Reserves configurable worker capacity for planning and finalization so file-group execution cannot saturate the worker pool.
Scans every eligible pinned-executor and job-kind bucket before reporting that no work is available.
Rotates the first scanned bucket to preserve fairness between eligible queues.
Reports lease-scan deadline exhaustion as a capacity error instead of an empty queue result.
Adds configuration for reserved control slots and unit coverage for the new scheduling and leasing behavior.

## Type of Change

- [X] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist

### DIFF
--- a/reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteReconcileExecutorPoller.java
+++ b/reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteReconcileExecutorPoller.java
@@ -16,6 +16,7 @@
 
 package ai.floedb.floecat.reconciler.impl;
 
+import ai.floedb.floecat.reconciler.jobs.ReconcileJobKind;
 import ai.floedb.floecat.reconciler.jobs.ReconcileJobStore;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
@@ -29,6 +30,7 @@ import java.util.Optional;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -46,6 +48,7 @@ public class RemoteReconcileExecutorPoller {
   private static final long MIN_CANCEL_CHECK_MS = 500L;
   private static final long HEARTBEAT_SHUTDOWN_WAIT_MS = 1_000L;
   private static final int DEFAULT_MAX_PARALLELISM = 1;
+  private static final int DEFAULT_RESERVED_CONTROL_SLOTS = 1;
 
   enum WorkerMode {
     LOCAL,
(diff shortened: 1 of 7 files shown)


service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/DurableReconcileJobStore.java
```diff
@@ -1191,7 +1191,16 @@ public Optional<LeasedJob> leaseNext(LeaseRequest request) {
             totalMs, scanStats.scanCount, scanStats.candidateCount, leased.isPresent());
       }
       observeLeaseNext(startedAtMs, scanStats, outcome);
+      if (scanStats.abortedByCaller) {
+        throw new CancellationException("reconcile lease scan cancelled by caller");
```
mrouvroy-floe, line 1195: [minor] **Caller-cancelled scan now throws `CancellationException`, triggering a wasted retry + ERROR log in the poller**

This change adds two throws where `leaseNext` previously returned an empty `Optional`. The deadline path (`RESOURCE_EXHAUSTED`) is handled gracefully by the poller, but the sibling `abortedByCaller` path now throws `CancellationException`, which `ReconcileExecutorControlImpl` maps to gRPC `CANCELLED`.

`GrpcRemoteReconcileExecutorClient.isTransportFailure` classifies `CANCELLED` as a transport failure, so the client burns its one retry, gets `CANCELLED` again, and rethrows. The poller's `leaseNextAssignment` catch treats `CANCELLED` as neither backpressure nor startup-unavailable, so it hits `throw e` and the worker task logs it at ERROR (`"Remote reconcile worker task failed before outcome submission"`). Previously an in-scan caller cancellation was a benign empty result.

Consider treating `CANCELLED` like backpressure in the poller (skip the cycle quietly), or excluding it from the client's transport-retry set, so genuine client/context cancellation doesn't produce retry churn and error-log noise.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteReconcileExecutorPoller.java
```diff
@@ -234,13 +268,39 @@ private boolean reserveWorkerSlot() {
     }
   }
 
+  private static boolean supportsControlWork(ReconcileExecutor executor) {
+    return executor != null
+        && executor.supportedJobKinds().stream()
+            .anyMatch(jobKind -> jobKind != ReconcileJobKind.EXEC_FILE_GROUP);
+  }
+
+  private static boolean isFileGroupOnly(ReconcileExecutor executor) {
```
mrouvroy-floe, line 277: [nit] **Reserved control slots silently become a no-op for a mixed control/file-group executor**

`init` computes a non-zero `reservedControlSlots` whenever `hasControlExecutor` is true (which a mixed executor satisfies), but `isFileGroupOnly` returns `false` for a mixed executor, so no lease ever acquires a `fileGroupSlots` permit and the reservation has no effect.

Today all executors are cleanly split (`RemoteFileGroupReconcileExecutor` is file-group-only, `RemoteDefaultReconcileExecutor` is control-only), so this is only latent. The behavior is documented in the comment here, but an operator who introduces a mixed executor and sets `reconciler.reserved-control-slots` would silently get no protection despite the config being accepted.

Worth a log line (or leaving as-is if a mixed executor is intentionally unsupported) so the silent no-op is discoverable.

---

PR #412, opened 2026-07-27
storage: give KV attributes a typed string/number model

mivashko-floe:
## Summary

  `Record.attrs` was `Map<String, String>`, so numeric bookkeeping values had to be
  stored as decimal strings, and the DynamoDB read path silently dropped every
  non-`S` attribute. Together those made an atomic counter impossible: incrementing
  meant read-parse-write under CAS, which loses bumps whenever two writers race.

  - Add `AttrValue`, a sealed string/number type, and retype `Record.attrs` to
    `Map<String, AttrValue>`.
  - Map `S`↔`StringValue` and `N`↔`NumberValue` in both directions in
    `DynamoDbKvStore`, so `N` attributes are no longer dropped on read. An
    unparseable `N` degrades to a string of its raw text rather than vanishing.
  - Add `KvStore.updateMetadataAttrsIfExists(key, sets, increments)`: one atomic
    partial update that sets attributes, adds to numeric attributes, and advances
    the CAS version together, and never creates a record. One `UpdateItem` on
    DynamoDB; one `computeIfPresent` in memory. 
  - Share the argument validation in `MetadataAttrUpdates` so both stores reject
    the same caller mistakes, synchronously, before any request is issued.
   
  Two deliberate limits, both documented at the SPI rather than papered over:

  - The new update is restricted to attrs-only records via
    `attribute_not_exists(#value)`. That guard is necessary but **not sufficient** —
    an empty value does not prove a record is attrs-only, since `PointerStoreEntity`
    encodes to no bytes and foreign writers share the table. The javadoc states the
    limit rather than claiming structural enforcement.
  - `ATTR_EXPIRES_AT` is still *written* as a string. Reads accept both forms, but
    writing it as `N` is unsafe under a rolling deploy: an old replica drops `N`
    attributes on read, and its next write then persists the pointer with no expiry
    attribute at all — permanent loss, not transient invisibility.
   
  Increments are strict about type in both stores, matching DynamoDB's `ADD`, which
  rejects an `S`-typed attribute regardless of whether its content looks numeric.
  The two stores differ only in exception type. The second commit fixes the
  `EntityContractTest` fake, which had been checking "does the text parse as a
  number" and so accepted `StringValue("42")` where production rejects it — no live
  failure, but a false pass waiting for the next entity test.

## Type of Change

- [ ] fix (bug fix)

## Testing

  - New `KvStoreContractTest` (both the in-memory and DynamoDB variants) pins the
    set/increment/version semantics so the two stores can't drift.
  - `EntityContractTest` (17 tests) for the fake-vs-production increment fix.

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

  Reads are backward-compatible in both directions: new binaries parse `S`-typed
  legacy values, and `ATTR_EXPIRES_AT` is still written as `S` specifically so old
  replicas keep seeing it.

  The one asymmetry worth understanding before merging a consumer: **an old replica
  drops `N` attributes on read.** Any table that a new binary starts writing numeric
  attributes into is no longer safe for an old binary to read-modify-write. This PR
  itself writes no numeric attributes to an existing table; it only makes them
  possible. The consumer in `core` handles the consequence by pointing at a fresh
  table.

## Checklist

## Additional Notes

  - Merge order: this PR must land before eng-floe/core#<core PR number>, which
    bumps the submodule to it.
  - `ControllerLeaseEntity` / `KvLeaseStore` (core side) deliberately do **not**
    adopt the new primitive — it has no expected-version check, so a lease renew
    would overwrite a lease another holder had stolen. Reviewers may want to sanity
    check that reasoning, since it is the main "why wasn't this used everywhere"
    question.

### DIFF
--- a/core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/AbstractEntity.java
+++ b/core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/AbstractEntity.java
@@ -65,9 +65,9 @@ public abstract class AbstractEntity<M extends MessageLite> implements KvAttribu
         return null;
       }
       var m = (M) defaultInstance.getParserForType().parseFrom(data);
-      var ts = r.attrs().get(ATTR_EXPIRES_AT);
-      if (ts != null) {
-        m = setExpiresAt(m, Long.parseLong(ts));
+      long ts = AttrValue.longOr(r.attrs(), ATTR_EXPIRES_AT, 0L);
+      if (ts > 0) {
+        m = setExpiresAt(m, ts);
       }
       return m;
     } catch (InvalidProtocolBufferException e) {
@@ -84,10 +84,10 @@ public abstract class AbstractEntity<M extends MessageLite> implements KvAttribu
     return 0L;
   }
 
-  protected static Map<String, String> pointerAttrs(KvStore.Key target) {
+  protected static Map<String, AttrValue> pointerAttrs(KvStore.Key target) {
     return Map.of(
-        TARGET_PARTITION_KEY, target.partitionKey(),
-        TARGET_SORT_KEY, target.sortKey());
+        TARGET_PARTITION_KEY, AttrValue.of(target.partitionKey()),
+        TARGET_SORT_KEY, AttrValue.of(target.sortKey()));
   }
 
   protected static String normalize(String s) {
@@ -117,14 +117,16 @@ public abstract class AbstractEntity<M extends MessageLite> implements KvAttribu
    * <p>The written record.version and protobuf message version are both set to nextVersion.
    */
   protected Uni<Optional<M>> putCanonicalCas(
-      KvStore.Key key, String kind, M message, Map<String, String> attrs, long expectedVersion) {
+      KvStore.Key key, String kind, M message, Map<String, AttrValue> attrs, long expectedVersion) {
 
     long nv = nextVersion(expectedVersion);
     M withVer = versionAccessor != null ? versionAccessor.withVersion(message, nv) : message;
     long ts = getExpiresAt(withVer);
     if (ts > 0) {
       attrs = new java.util.HashMap<>(attrs);
-      attrs.put(ATTR_EXPIRES_AT, Long.toString(ts));
+      // Written as a string, not a number: an older replica's read path drops non-string attrs, and
+      // its next write of this record would then persist it without an expiry at all.
+      attrs.put(ATTR_EXPIRES_AT, AttrValue.of(Long.toString(ts)));
     }
     var rec = new KvStore.Record(key, kind, encode(withVer), attrs, nv);
     return kv.putCas(rec, expectedVersion).map(ok -> ok ? Optional.of(withVer) : Optional.empty());
@@ -141,7 +143,11 @@ public abstract class AbstractEntity<M extends MessageLite> implements KvAttribu
    * <p>Caller supplies raw value bytes; we still write a monotonically increasing version.
    */
   protected Uni<Boolean> putRawCas(
-      KvStore.Key key, String kind, byte[] value, Map<String, String> attrs, long expectedVersion) {
+      KvStore.Key key,
+      String kind,
+      byte[] value,
+      Map<String, AttrValue> attrs,
+      long expectedVersion) {
 
     long nv = nextVersion(expectedVersion);
     var rec = new KvStore.Record(key, kind, value, attrs, nv);
@@ -157,8 +163,8 @@ public abstract class AbstractEntity<M extends MessageLite> implements KvAttribu
             optPtr -> {
               if (optPtr.isEmpty()) return Uni.createFrom().item(Optional.empty());
               var ptr = optPtr.get();
-              var tpk = ptr.attrs().get(TARGET_PARTITION_KEY);
-              var tsk = ptr.attrs().get(TARGET_SORT_KEY);
+              var tpk = AttrValue.stringOr(ptr.attrs(), TARGET_PARTITION_KEY, null);
+              var tsk = AttrValue.stringOr(ptr.attrs(), TARGET_SORT_KEY, null);
               if (tpk == null || tsk == null) return Uni.createFrom().item(Optional.empty());
               return get(new KvStore.Key(tpk, tsk));
             });
@@ -190,8 +196,11 @@ public abstract class AbstractEntity<M extends MessageLite> implements KvAttribu
                   page.items().stream()
                       .map(
                           r -> {
-                            String pk = r.attrs().get(KvAttributes.TARGET_PARTITION_KEY);
-                            String sk = r.attrs().get(KvAttributes.TARGET_SORT_KEY);
+                            String pk =
+                                AttrValue.stringOr(
+                                    r.attrs(), KvAttributes.TARGET_PARTITION_KEY, null);
+                            String sk =
+                                AttrValue.stringOr(r.attrs(), KvAttributes.TARGET_SORT_KEY, null);
                             return kv.get(new KvStore.Key(pk, sk));
                           })
                       .toList();
--- a/core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/AbstractEntity.java
+++ b/core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/AbstractEntity.java
@@ -65,9 +65,11 @@ public abstract class AbstractEntity<M extends MessageLite> implements KvAttribu
         return null;
       }
       var m = (M) defaultInstance.getParserForType().parseFrom(data);
-      long ts = AttrValue.longOr(r.attrs(), ATTR_EXPIRES_AT, 0L);
-      if (ts > 0) {
-        m = setExpiresAt(m, ts);
+      var ts = r.attrs().get(ATTR_EXPIRES_AT);
+      if (ts != null) {
+        // Accepts both the legacy string form and the native numeric one; a present but malformed
+        // stamp throws, so corrupt expiry data cannot read as "no expiry".
+        m = setExpiresAt(m, ts.asLong());
       }
       return m;
     } catch (InvalidProtocolBufferException e) {
--- a/core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/AbstractEntity.java
+++ b/core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/AbstractEntity.java
@@ -67,8 +67,8 @@ public abstract class AbstractEntity<M extends MessageLite> implements KvAttribu
       var m = (M) defaultInstance.getParserForType().parseFrom(data);
       var ts = r.attrs().get(ATTR_EXPIRES_AT);
       if (ts != null) {
-        // Accepts both the legacy string form and the native numeric one; a present but malformed
-        // stamp throws, so corrupt expiry data cannot read as "no expiry".
+        // String or numeric form both decode; a malformed stamp throws rather than reading as
+        // "no expiry".
         m = setExpiresAt(m, ts.asLong());
       }
       return m;
@@ -126,8 +126,8 @@ public abstract class AbstractEntity<M extends MessageLite> implements KvAttribu
     long ts = getExpiresAt(withVer);
     if (ts > 0) {
       attrs = new java.util.HashMap<>(attrs);
-      // Written as a string, not a number: an older replica's read path drops non-string attrs, and
-      // its next write of this record would then persist it without an expiry at all.
+      // String, not number: an older replica drops non-string attrs on read and would then
+      // rewrite the record without its expiry (see AttrWriteRules).
       attrs.put(ATTR_EXPIRES_AT, AttrValue.of(Long.toString(ts)));
     }
     var rec = new KvStore.Record(key, kind, encode(withVer), attrs, nv);
--- a/core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/KvAttributes.java
+++ b/core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/KvAttributes.java
@@ -15,6 +15,8 @@
  */
 package ai.floedb.floecat.storage.kv;
 
+import java.util.Set;
+
 public interface KvAttributes {
   String ATTR_PARTITION_KEY = "pk";
   String ATTR_SORT_KEY = "sk";
@@ -26,4 +28,11 @@ public interface KvAttributes {
 
   String TARGET_PARTITION_KEY = "targetPk";
   String TARGET_SORT_KEY = "targetSk";
+
+  /**
+   * Names a backend uses for the record's own structure. They are reserved: a record attribute may
+   * not use one, because storing it would collide with the structural field of the same name.
+   */
+  Set<String> STRUCTURAL_ATTRS =
+      Set.of(ATTR_PARTITION_KEY, ATTR_SORT_KEY, ATTR_KIND, ATTR_VALUE, ATTR_VERSION);
 }
--- a/core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/KvAttributes.java
+++ b/core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/KvAttributes.java
@@ -30,17 +30,13 @@ public interface KvAttributes {
   String TARGET_SORT_KEY = "targetSk";
 
   /**
-   * Names a backend gives its own meaning, on a record it also stores attributes on. They are
-   * reserved: a record attribute may not use one. For {@link #ATTR_PARTITION_KEY}, {@link
-   * #ATTR_SORT_KEY}, {@link #ATTR_KIND}, {@link #ATTR_VALUE} and {@link #ATTR_VERSION} the harm is
-   * a collision with the structural field of the same name. For {@link #ATTR_TTL} it is worse and
-   * quieter: {@code DynamoDbTablesBootstrap} enables DynamoDB's TTL feature on that attribute, so
-   * an ordinary numeric attr that happens to be named {@code ttl} is read by DynamoDB as the row's
-   * expiry time and the row is deleted once it passes — no error, no trace.
+   * Names the backend gives its own meaning; a record attribute may not use one. Most would collide
+   * with the structural field of the same name. {@link #ATTR_TTL} is quieter: DynamoDB's TTL
+   * feature is enabled on that attribute, so a numeric attr named {@code ttl} silently schedules
+   * the row's deletion.
    *
-   * <p>Reserved on the way in only. A row that some other writer already put one of these names on
-   * still has to be readable, so decoders drop them rather than refusing the record (see {@code
-   * DynamoDbKvStore#avToAttrs}).
+   * <p>Reserved on writes only — rows written by others may carry these names, so decoders drop
+   * them rather than refusing the record.
    */
   Set<String> RESERVED_ATTRS =
       Set.of(ATTR_PARTITION_KEY, ATTR_SORT_KEY, ATTR_KIND, ATTR_VALUE, ATTR_VERSION, ATTR_TTL);
--- a/core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/KvAttributes.java
+++ b/core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/KvAttributes.java
@@ -30,9 +30,18 @@ public interface KvAttributes {
   String TARGET_SORT_KEY = "targetSk";
 
   /**
-   * Names a backend uses for the record's own structure. They are reserved: a record attribute may
-   * not use one, because storing it would collide with the structural field of the same name.
+   * Names a backend gives its own meaning, on a record it also stores attributes on. They are
+   * reserved: a record attribute may not use one. For {@link #ATTR_PARTITION_KEY}, {@link
+   * #ATTR_SORT_KEY}, {@link #ATTR_KIND}, {@link #ATTR_VALUE} and {@link #ATTR_VERSION} the harm is
+   * a collision with the structural field of the same name. For {@link #ATTR_TTL} it is worse and
+   * quieter: {@code DynamoDbTablesBootstrap} enables DynamoDB's TTL feature on that attribute, so
+   * an ordinary numeric attr that happens to be named {@code ttl} is read by DynamoDB as the row's
+   * expiry time and the row is deleted once it passes — no error, no trace.
+   *
+   * <p>Reserved on the way in only. A row that some other writer already put one of these names on
+   * still has to be readable, so decoders drop them rather than refusing the record (see {@code
+   * DynamoDbKvStore#avToAttrs}).
    */
-  Set<String> STRUCTURAL_ATTRS =
-      Set.of(ATTR_PARTITION_KEY, ATTR_SORT_KEY, ATTR_KIND, ATTR_VALUE, ATTR_VERSION);
+  Set<String> RESERVED_ATTRS =
+      Set.of(ATTR_PARTITION_KEY, ATTR_SORT_KEY, ATTR_KIND, ATTR_VALUE, ATTR_VERSION, ATTR_TTL);
 }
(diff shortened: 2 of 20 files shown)


core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/AbstractEntity.java
```diff
@@ -65,9 +65,9 @@ protected M decode(KvStore.Record r) {
         return null;
       }
       var m = (M) defaultInstance.getParserForType().parseFrom(data);
-      var ts = r.attrs().get(ATTR_EXPIRES_AT);
-      if (ts != null) {
-        m = setExpiresAt(m, Long.parseLong(ts));
+      long ts = AttrValue.longOr(r.attrs(), ATTR_EXPIRES_AT, 0L);
```
mrouvroy-floe: This changes malformed expiry handling from fail-loud to “no expiry”: a present but unparsable timestamp now falls back to 0 via longOr(...). The change only requires accepting both S and N forms, not silently ignoring invalid values.

I’d prefer preserving the old behavior for present-but-invalid expiry attrs: accept StringValue/NumberValue when they parse, but throw on malformed timestamp so corrupt expiry data does not become an immortal record.

mivashko-floe: Agree. Pushed the fix

core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/MetadataAttrUpdates.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }
+
+  private static void checkNames(Map<String, ?> attrs, String what) {
+    for (var e : attrs.entrySet()) {
+      var name = e.getKey();
+      if (name == null || name.isBlank()) {
+        throw new IllegalArgumentException(what + " contains a blank attr name");
+      }
+      if (KvAttributes.STRUCTURAL_ATTRS.contains(name)) {
```
mrouvroy-floe: [major] **Expiry timestamp can be written as `N`**

`updateMetadataAttrsIfExists` accepts arbitrary metadata names and only rejects structural attrs here, so callers can pass `KvAttributes.ATTR_EXPIRES_AT` in `sets` with `AttrValue.of(long)` or in `increments`. On DynamoDB, that flows through `attrToAv`/`N(...)` and writes the `timestamp` attribute as native `N`.

That is the exact form the PR description says is unsafe during a rolling deploy: old replicas drop non-`S` attrs on read, and their next whole-record write can persist the pointer without any expiry attribute, making the expiry loss permanent.

Please enforce the rollout invariant in the shared validation layer. Either reject `ATTR_EXPIRES_AT` entirely for this partial-update primitive, or allow it only as a `StringValue` in `sets` and reject it in `increments`, so both stores fail synchronously before any request is issued.

---

🤖 _Generated with GPT-5.5._

core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/KvAttributes.java
```diff
@@ -26,4 +28,11 @@ public interface KvAttributes {
 
   String TARGET_PARTITION_KEY = "targetPk";
   String TARGET_SORT_KEY = "targetSk";
+
+  /**
+   * Names a backend uses for the record's own structure. They are reserved: a record attribute may
+   * not use one, because storing it would collide with the structural field of the same name.
+   */
+  Set<String> STRUCTURAL_ATTRS =
```
mrouvroy-floe: [major] **Reserve `ttl` in `STRUCTURAL_ATTRS` to prevent accidental record expiry**

`STRUCTURAL_ATTRS` omits `ATTR_TTL`, so nothing prevents callers from writing or incrementing an attribute literally named `ttl` through `updateMetadataAttrsIfExists`.

This matters because `DynamoDbTablesBootstrap` enables DynamoDB's TTL feature on `ATTR_TTL`. A bookkeeping counter or an accidental metadata key with that name will be interpreted by DynamoDB as a row-expiration timestamp, silently deleting live records. The existing `AttrWriteRulesTest.accepts_numeric_values_for_every_other_attr` test even uses `ATTR_TTL` as a sample ordinary attribute, reinforcing the false sense that it is safe.

Add `ATTR_TTL` to the shared attr-name guard in `STRUCTURAL_ATTRS` (or introduce a dedicated expiry-writing path with explicit semantics), and update `AttrWriteRulesTest` to stop treating `ATTR_TTL` as an example of an ordinary numeric attr.

---

core/storage-spi/src/test/java/ai/floedb/floecat/storage/kv/EntityContractTest.java
```diff
@@ -401,6 +459,43 @@ public Uni<Boolean> deleteCas(Key key, long expectedVersion) {
...
+      var attrs = new HashMap<>(existing.attrs());
+      attrs.putAll(sets);
+      for (var increment : increments.entrySet()) {
+        var name = increment.getKey();
+        var current = attrs.get(name);
+        // Rejected on the type, not on whether the text happens to parse — same rule as
+        // InMemoryKvStore, because DynamoDB's ADD raises ValidationException for any S-typed
+        // attribute, numeric-looking ones like "42" included. A laxer fake would let an entity
+        // test pass here and fail in production.
+        if (current != null && !(current instanceof AttrValue.NumberValue)) {
```
mrouvroy-floe, line 513: [minor] **Fake `updateMetadataAttrsIfExists` throws synchronously instead of returning a failed `Uni`**

The test fake's `updateMetadataAttrsIfExists` throws `IllegalStateException` directly when an increment targets a string attribute, so the exception escapes synchronously out of a `Uni`-returning method.

This diverges from production behaviour: `InMemoryKvStore` wraps the same failure in a deferred `Uni.createFrom().item(() -> ...)` supplier (with an explicit comment explaining the failure must arrive as a failed `Uni`, not a synchronous throw), and `DynamoDbKvStore` surfaces the equivalent `ValidationException` as a failed `Uni` as well. A future contract test that increments a string attr and calls `.await()` expecting a failed `Uni` would instead see a synchronous throw at the call site against this fake — exactly the false-pass-vs-production drift the second commit set out to eliminate.

Wrap the throw in a deferred supplier so the fake's failure mode matches production.

---

kwesterfeld-floe (approved):

core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/MetadataAttrUpdates.java
mivashko-floe: Already addressed before this comment landed — it was written against `d078f512`, and `647d6474` (the next commit, "enforce the string-typed expiry stamp on every write path") added exactly this guard.

`MetadataAttrUpdates.checkNames` rejects `ATTR_EXPIRES_AT` outright rather than per-form, i.e. the stricter of your two options: refused in `sets` in *either* form and in `increments`. The reasoning for the flat rule over a StringValue-in-sets allowance is in the comment there — an increment can only produce the numeric form, and no caller needs to bump an expiry without rewriting the record that carries it.

Pinned by `MetadataAttrUpdatesTest.rejects_expiry_stamp_in_sets_in_either_form` and `rejects_expiry_stamp_in_increments`.

core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/KvAttributes.java
mivashko-floe: Good catch — fixed in `a8fc2f01`, and it needed a second half you did not mention.

`ATTR_TTL` is now in the shared set, which is the mechanism you suggested and the right one: `Record`s constructor covers `putCas`/`txnWriteCas`, `MetadataAttrUpdates` covers `sets` and `increments`, and both stores inherit each — a write-only rule next to `checkExpiryIsString` would have needed per-write-path coverage with a hole the moment one forgot to call it.

**The read-path half:** that set is also enforced in `Record`s constructor, which runs on *decode*, and `DynamoDbKvStore.avToAttrs` listed the five names it skipped instead of deriving them. Adding `ATTR_TTL` to the set alone would have made any stored row carrying a `ttl` attr — an out-of-band writer's, say — throw `IllegalArgumentException` on every `get` and `query`. `avToAttrs` now derives its skip set, so a later addition to the set cannot become a read outage either. `DynamoDbKvStoreTest.reserved_attr_written_out_of_band_is_dropped_rather_than_failing_the_read` pins it; I verified it fails on the old listed form before keeping it.

Also renamed the set `STRUCTURAL_ATTRS` -> `RESERVED_ATTRS` (and the thrown messages to "reserved by the backend"). `ttl` is reserved because the backend gives the name its own meaning, not because it collides with a structural field, and the old name invited pruning it back out — which silently restores the data-loss path.

`AttrWriteRulesTest` no longer uses `ATTR_TTL` as its sample ordinary numeric attr, and `KvStoreRecordTest.record_rejects_a_ttl_attr` names the case separately from the set-driven loop so the *reason* survives.

One thing worth your call, no code change: `ATTR_TTL` looks write-unreachable across the whole tree. `DynamoDbTablesBootstrap:51` enables DynamoDB TTL on it, but nothing ever writes it — application expiry goes through `ATTR_EXPIRES_AT`/`"timestamp"` (`AbstractEntity:131`, `PointerStoreEntity:261`). So either the TTL enablement is vestigial, or it exists for a writer outside this repo. Reserving the name is correct either way; removing the enablement is a behavioural change that does not belong in this PR.

core/storage-spi/src/test/java/ai/floedb/floecat/storage/kv/EntityContractTest.java
mivashko-floe: Fixed in `84ded1b3`. The body moves into a deferred supplier, matching `InMemoryKvStore` exactly, with argument validation left synchronous as in both stores.

Worth recording why it survived: nothing calls this override — it exists only to satisfy the interface — so the divergence was invisible. Your "a future contract test would see a synchronous throw" was exactly right, and since nothing exercises it, the fix could regress just as quietly. So there is now a test: `fake_kv_delivers_increment_of_string_attr_as_a_failed_uni` asserts both halves, that the call itself does not throw (`assertDoesNotThrow`) and that awaiting yields the failure. Verified it catches the old behaviour — simulating a synchronous throw fails it with `Unexpected exception thrown: IllegalStateException`.

One nuance in passing: the fake is now timing-identical to `InMemoryKvStore` but not to DynamoDB, whose request is issued eagerly. The failure *delivery* matches all three, which is what the fake is for.

storage/memory/src/main/java/ai/floedb/floecat/storage/memory/InMemoryKvStore.java
```diff
@@ -67,6 +74,63 @@ public Uni<Boolean> deleteCas(Key key, long expectedVersion) {
...
+                      if (current != null && !(current instanceof AttrValue.NumberValue)) {
+                        throw new IllegalStateException(
+                            "cannot increment attr "
+                                + name
+                                + " on "
+                                + key
+                                + ": it currently holds a string value");
+                      }
+                      long base = current == null ? 0L : ((AttrValue.NumberValue) current).value();
+                      attrs.put(name, AttrValue.of(base + increment.getValue()));
```
mrouvroy-floe: [major] **Reject counter overflow consistently**

`AttrValue.NumberValue` is a `long`, but this unchecked addition wraps. For example, incrementing `Long.MAX_VALUE` by `1` silently stores `Long.MIN_VALUE` and reports a successful update.

DynamoDB `ADD` instead retains the larger decimal `N` value. On the next read this implementation degrades that value to `StringValue`, so the two `KvStore` implementations disagree both on the result and on whether the attribute remains numeric.

Define an overflow policy at the SPI boundary and enforce it atomically in both stores (for example, fail the update before a result can leave the `long` model), then add the `Long.MAX_VALUE`/`Long.MIN_VALUE` boundary cases to the shared contract tests.

---

🤖 _Generated with GPT-5.6 Terra._

mivashko-floe: Fixed in `3c578cc7`, though not the way you suggested — worth saying why.

In-memory now uses `Math.addExact`, so a wrap becomes a failed `Uni` with nothing half-applied. I left DynamoDB as-is and documented the difference instead.

Two reasons. A condition term makes overflow surface as `ConditionalCheckFailedException`, which maps to `Optional.empty()` — and `KvRemoteIndex.touch` reads empty as "no entry" and swallows it, so overflow would get quieter rather than louder. Doing it properly needs `NOT attribute_type(#a0, :n)` in the condition (a bare `#a0 <= :lim` turns the string-attr `ValidationException` into a condition failure) plus `ALL_OLD` to tell overflow from absent. That's real surgery on the path every `touch` takes, for a boundary at ~9.2e18 increments of a single counter.

The gap that's left leans strict: in-memory rejects what DynamoDB would accept, never the reverse — so it can only cause a false failure in tests, not a false pass in prod.

Happy to do the strict version on both stores as its own commit if you'd rather. Leaving this open for your call.

🤖 *Generated with Claude Opus 5.*

storage/memory/src/main/java/ai/floedb/floecat/storage/memory/InMemoryKvStore.java
```diff
@@ -67,6 +74,68 @@ public Uni<Boolean> deleteCas(Key key, long expectedVersion) {
...
+  @Override
+  public Uni<Optional<Long>> updateMetadataAttrsIfExists(
+      Key key, Map<String, AttrValue> sets, Map<String, Long> increments) {
+    MetadataAttrUpdates.validate(key, sets, increments);
+
+    // Deferred (a supplier, unlike the eager item(...) used elsewhere in this class) so that the
+    // loud non-numeric-increment failure below arrives as a failed Uni instead of a synchronous
+    // throw out of a Uni-returning method. Argument validation above stays synchronous, matching
+    // deleteCas's guard.
+    return Uni.createFrom()
```
mrouvroy-floe: [major] **Lazy `Uni` in `updateMetadataAttrsIfExists` can double-apply mutations on repeated subscription**

`updateMetadataAttrsIfExists` returns `Uni.createFrom().item(() -> ...)`, deferring the side-effecting mutation to subscription time rather than executing it eagerly.

A `Uni` can be subscribed to more than once; any caller that awaits or chains the same returned `Uni` twice will run the supplier again, incrementing the counter or version a second time. The rest of the in-memory store performs mutations eagerly before wrapping the result in `Uni.createFrom().item(...)`, and the DynamoDB path issues a single already-started `CompletionStage` — so the same caller observes exactly one update there. This asymmetry makes the atomic counter primitive backend-dependent: accidental double-subscription produces one DynamoDB `UpdateItem` but two in-memory increments, undermining the very correctness guarantee this method was introduced to provide.

Replace the lazy supplier with an eager evaluation: compute the mutation immediately and return `Uni.createFrom().item(result)` (no lambda), matching the pattern used everywhere else in this store.

---

core/storage-spi/src/test/java/ai/floedb/floecat/storage/kv/EntityContractTest.java
```diff
@@ -401,6 +483,53 @@ public Uni<Boolean> deleteCas(Key key, long expectedVersion) {
...
+                  if (current != null && !(current instanceof AttrValue.NumberValue)) {
+                    throw new IllegalStateException(
+                        "cannot increment attr "
+                            + name
+                            + " on "
+                            + key
+                            + ": it currently holds a string value");
+                  }
+                  long base = current == null ? 0L : ((AttrValue.NumberValue) current).value();
+                  attrs.put(name, AttrValue.of(base + increment.getValue()));
```
mrouvroy-floe: [nit] **Fake increment uses plain `+` instead of `Math.addExact`, diverging from real store arithmetic**

The fake computes `base + increment.getValue()`, whereas the production `InMemoryKvStore` uses `Math.addExact(base, ...)` (line 128) and DynamoDB rejects out-of-range sums. The impact is low today because no entity test drives a counter to the `long` boundary through this fake.

However, this is the same category of fake-vs-production divergence the second commit of this PR set out to eliminate: a future overflow-related entity test would see the fake silently wrap and report success while both real stores fail, exactly the false-pass that `fake_kv_delivers_increment_of_string_attr_as_a_failed_uni` was added to guard against.

Switch to `Math.addExact` here to keep the fake's arithmetic in lockstep with the real stores.

---

storage/memory/src/main/java/ai/floedb/floecat/storage/memory/InMemoryKvStore.java
mivashko-floe: Confirmed and fixed, though not quite as suggested. The double-apply is real — a probe that awaits the returned `Uni` twice takes `hits` from 0 to 2 and the version from 1 to 3, whereas the DynamoDB path issues one `UpdateItem` because `ddb.call(...)` runs before the `Uni` is constructed and `fromStage` only attaches a per-subscription callback to the already-started stage. But going straight to `Uni.createFrom().item(result)` would revert 84ded1b36: the supplier is what delivers the increment guards (`IllegalStateException` on an S-typed attr, `Math.addExact` overflow) as a *failed* `Uni` rather than a synchronous throw out of a `Uni`-returning method, which is how the real store surfaces `ValidationException` from `ADD` against a string attribute — so the plain eager form fixes double-apply by breaking failure delivery. Instead the mutation now runs eagerly in a private `applyMetadataAttrUpdates` helper and a `RuntimeException` from it is converted with `Uni.createFrom().failure(...)`, which keeps both properties: one application per call regardless of subscription count, and failures still arriving asynchronously. Added `updateMetadataAttrsIfExists_applies_once_however_often_the_uni_is_subscribed` to both the in-memory and DynamoDB contract tests to pin the guarantee on both backends. Also swept the rest of `storage/` and `core/` main sources for other side-effecting `item(() -> …)` suppliers — this was the only one. Fixed in f93578e1e.

core/storage-spi/src/test/java/ai/floedb/floecat/storage/kv/EntityContractTest.java
```diff
@@ -401,6 +483,53 @@ public Uni<Boolean> deleteCas(Key key, long expectedVersion) {
...
+    @Override
+    public Uni<Optional<Long>> updateMetadataAttrsIfExists(
+        Key key, Map<String, AttrValue> sets, Map<String, Long> increments) {
+      MetadataAttrUpdates.validate(key, sets, increments);
+
+      // Deferred (a supplier, unlike the eager item(...) used elsewhere in this fake) so that the
+      // non-numeric-increment failure below arrives as a failed Uni, the way InMemoryKvStore and
+      // DynamoDbKvStore both surface it, rather than as a synchronous throw a test awaiting the Uni
+      // would never catch. Argument validation above stays synchronous, matching them too.
+      return Uni.createFrom()
```
mrouvroy-floe: [minor] **Fake `updateMetadataAttrsIfExists` reintroduces subscription-sensitive mutation and overflow bugs**

`TestKvStore.updateMetadataAttrsIfExists` still performs the successful mutation inside `Uni.createFrom().item(() -> ...)`, so subscribing to the same returned `Uni` twice applies the update twice in the fake. This directly diverges from both real stores: `InMemoryKvStore` was changed in this PR to apply eagerly and wrap failures as `Uni.createFrom().failure(...)`, and `DynamoDbKvStore` has an already-started `UpdateItem`. The comment at line 491 claims the supplier is needed to deliver a failed `Uni`, but the sibling commit (`f93578e1e`) proved you get both eager-once semantics and a failed `Uni` from the same helper pattern, making that rationale stale.

A future entity test that exercises a successful metadata update through this fake can pass or fail depending on subscription count even though production applies the update exactly once per call. Additionally, line 521 sums with plain `+`, whereas `InMemoryKvStore` uses `Math.addExact`, leaving silent integer overflow undetected in the fake.

Mirror the production in-memory pattern: perform the mutation eagerly outside the `Uni`, convert `RuntimeException` to `Uni.createFrom().failure(...)`, and replace `+` with `Math.addExact` so the fake also protects against overflow.

---

core/storage-spi/src/main/java/ai/floedb/floecat/storage/kv/KvAttributes.java
```diff
@@ -26,4 +28,20 @@ public interface KvAttributes {
 
   String TARGET_PARTITION_KEY = "targetPk";
   String TARGET_SORT_KEY = "targetSk";
+
+  /**
+   * Names a backend gives its own meaning, on a record it also stores attributes on. They are
+   * reserved: a record attribute may not use one. For {@link #ATTR_PARTITION_KEY}, {@link
+   * #ATTR_SORT_KEY}, {@link #ATTR_KIND}, {@link #ATTR_VALUE} and {@link #ATTR_VERSION} the harm is
+   * a collision with the structural field of the same name. For {@link #ATTR_TTL} it is worse and
```
mrouvroy-floe: [minor] **LocalStack TTL attribute mismatch undermines the reserved-name guard**

The new reserved-name rule rejects `ATTR_TTL` (`ttl`) because the backend expires rows by that attribute, but the LocalStack setup paths (`Makefile` and `docker/localstack-init.sh`) still enable DynamoDB TTL on `expires_at`. If those scripts create the table before `DynamoDbTablesBootstrap.ensureTtlEnabled` runs, the bootstrap sees TTL already enabled and returns without verifying that the enabled attribute is actually `ttl`.

This leaves local/dev tables with a different backend-reserved attribute than the one this guard protects. A numeric metadata attr named `expires_at` is accepted by `Record` and `MetadataAttrUpdates`, yet LocalStack treats it as the row expiration timestamp; conversely, tests against local tables do not exercise the production `ttl` reservation at all.

Update the LocalStack bootstrap scripts to enable TTL on `ttl` (not `expires_at`) so that local and production environments share the same reserved attribute, and the guard is actually exercised by local tests.

---

mrouvroy-floe (approved):
Overall looks good, just two small minors related to tests that I think are worth addressing before merge

storage/aws/src/main/java/ai/floedb/floecat/storage/kv/dynamodb/DynamoDbKvStore.java
```diff
@@ -271,6 +296,87 @@ public Uni<Boolean> deleteCas(Key key, long expectedVersion) {
...
+            .updateExpression(expr.toString())
+            // attribute_not_exists(#value) is the attrs-only guard: a value-carrying record keeps a
+            // copy of its version inside the serialized payload, which this update cannot rewrite.
+            .conditionExpression("attribute_exists(#pk) AND attribute_not_exists(#value)")
+            .expressionAttributeNames(names)
+            .expressionAttributeValues(values)
+            .returnValues(ReturnValue.UPDATED_NEW)
+            .build();
+
+    return dynamo(client -> client.updateItem(req))
```
mrouvroy-floe, line 353: [major] **DynamoDB `ADD` silently stores values outside `long` range, violating the `KvStore` SPI contract**

**Problem:** `updateMetadataAttrsIfExists` forwards caller deltas directly to a DynamoDB `ADD` expression and treats any response where `version` parses as a success. Because DynamoDB numbers are arbitrary-precision, incrementing an attribute at `Long.MAX_VALUE` by `1` succeeds server-side and durably stores `9223372036854775808` — a value no Java `long` can represent. The `KvStore` SPI javadoc (`KvStore.java:105`) promises that increments past the range of `long` *fail the `Uni` with a store-specific error; neither overwrites nor wraps*, a guarantee upheld by `InMemoryKvStore` via `Math.addExact`. The DynamoDB implementation silently breaks that contract.

**Impact:** A caller who trusts the SPI — "if the `Uni` succeeds, my counter is a valid `long`" — is misled. The row is left in a state the SPI cannot faithfully represent: the next read degrades that DynamoDB `N` attribute to `StringValue` and `asLong()` throws, so the metadata version has already advanced but is now unreadable. The existing contract test (`updateMetadataAttrsIfExists_increment_past_long_max_does_not_wrap`) documents the broken behavior rather than asserting the correct one.

**Fix:** Add an atomic overflow guard to the DynamoDB update expression — e.g., a `ConditionExpression` that rejects the update when `attribute + delta > :long_max` — so the `UpdateItem` call fails before the out-of-range value is stored. Then update the contract test to assert that this condition triggers a failed `Uni` rather than a downstream read error. The SPI javadoc in `KvStore.java:105` should also note which specific exception type implementations are expected to surface.

---

PR #413, opened 2026-07-27
feat: Add complete connector definition export and import tooling

mcusack-floe:
## Summary

- Adds a protected RPC for exporting unmasked connector definitions and separately stored credentials.
- Adds a Java CLI for exporting, inspecting, validating, and importing connector archives.
- Converts account-scoped destination IDs into portable names and paths.
- Supports dry runs and configurable conflict handling.
- Protects archives with owner-only permissions and explicit plaintext-secret acknowledgement.
- Adds focused service, permission, and archive tests.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist

### DIFF
--- a/core/proto/src/main/proto/floecat/connector/connector.proto
+++ b/core/proto/src/main/proto/floecat/connector/connector.proto
@@ -253,6 +253,34 @@ message ValidateConnectorResponse {
   map<string,string> capabilities = 3;
 }
 
+// Administrative export payload. Unlike GetConnector/ListConnectors, this response is not
+// masked. Access is guarded by the connector.export permission.
+message ExportConnectorRequest {
+  ai.floedb.floecat.common.ResourceId connector_id = 1;
+  bool include_credentials = 2;
+}
+
+message ExportConnectorResponse {
+  Connector connector = 1;
+  optional AuthCredentials credentials = 2;
+}
+
+// Portable on-disk format used by floecat-connector-transfer. connector preserves the exact
+// source record for inspection, while portable_spec replaces account-scoped destination IDs with
+// names and paths suitable for CreateConnector in another account.
+message ConnectorTransferEntry {
+  Connector connector = 1;
+  ConnectorSpec portable_spec = 2;
+  optional AuthCredentials credentials = 3;
+}
+
+message ConnectorTransferBundle {
+  uint32 format_version = 1;
+  string source_account_id = 2;
+  google.protobuf.Timestamp exported_at = 3;
+  repeated ConnectorTransferEntry entries = 4;
+}
+
 service Connectors {
   rpc GetConnector(GetConnectorRequest) returns (GetConnectorResponse);
   rpc ListConnectors(ListConnectorsRequest) returns (ListConnectorsResponse);
@@ -261,4 +289,5 @@ service Connectors {
   rpc DeleteConnector(DeleteConnectorRequest) returns (DeleteConnectorResponse);
 
   rpc ValidateConnector(ValidateConnectorRequest) returns (ValidateConnectorResponse);
+  rpc ExportConnector(ExportConnectorRequest) returns (ExportConnectorResponse);
 }
--- a/service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
+++ b/service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
@@ -32,6 +32,8 @@ import ai.floedb.floecat.connector.rpc.CreateConnectorResponse;
 import ai.floedb.floecat.connector.rpc.DeleteConnectorRequest;
 import ai.floedb.floecat.connector.rpc.DeleteConnectorResponse;
 import ai.floedb.floecat.connector.rpc.DestinationTarget;
+import ai.floedb.floecat.connector.rpc.ExportConnectorRequest;
+import ai.floedb.floecat.connector.rpc.ExportConnectorResponse;
 import ai.floedb.floecat.connector.rpc.GetConnectorRequest;
 import ai.floedb.floecat.connector.rpc.GetConnectorResponse;
 import ai.floedb.floecat.connector.rpc.ListConnectorsRequest;
@@ -214,6 +216,58 @@ public class ConnectorsImpl extends BaseServiceImpl implements Connectors {
                   return GetConnectorResponse.newBuilder()
                       .setConnector(maskConnector(connector))
                       .build();
+                }),
+            correlationId())
+        .onFailure()
+        .invoke(L::fail)
+        .onItem()
+        .invoke(L::ok);
+  }
+
+  @Override
+  public Uni<ExportConnectorResponse> exportConnector(ExportConnectorRequest request) {
+    var L = LogHelper.start(LOG, "ExportConnector");
+
+    return mapFailures(
+            run(
+                () -> {
+                  var principalContext = principalProvider.get();
+                  var correlationId = principalContext.getCorrelationId();
+                  var accountId = principalContext.getAccountId();
+
+                  authz.require(principalContext, "connector.export");
+                  var connectorId =
+                      scopedConnectorId(accountId, request.getConnectorId(), correlationId);
+
+                  for (int attempt = 0; attempt < 3; attempt++) {
+                    var before = connectorRepo.getById(connectorId);
+                    if (before.isEmpty()) {
+                      throw GrpcErrors.notFound(
+                          correlationId,
+                          GeneratedErrorMessages.MessageKey.CONNECTOR,
+                          Map.of("id", connectorId.getId()));
+                    }
+
+                    Optional<AuthCredentials> firstCredentials = Optional.empty();
+                    Optional<AuthCredentials> secondCredentials = Optional.empty();
+                    if (request.getIncludeCredentials()) {
+                      firstCredentials = credentialResolver.resolve(accountId, connectorId.getId());
+                      secondCredentials =
+                          credentialResolver.resolve(accountId, connectorId.getId());
+                    }
+                    var after = connectorRepo.getById(connectorId);
+
+                    if (after.isPresent()
+                        && before.get().equals(after.get())
+                        && firstCredentials.equals(secondCredentials)) {
+                      var response =
+                          ExportConnectorResponse.newBuilder().setConnector(before.get());
+                      firstCredentials.ifPresent(response::setCredentials);
+                      return response.build();
+                    }
+                  }
+
+                  throw GrpcErrors.aborted(correlationId, Map.of("id", connectorId.getId()));
                 }),
             correlationId())
         .onFailure()
--- a/service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
+++ b/service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
@@ -574,63 +574,76 @@ public class ConnectorsImpl extends BaseServiceImpl implements Connectors {
                                       Map.of("id", connectorId.getId())));
 
                   FieldMask normalizedMask = normalizeMask(request.getUpdateMask());
+                  String accountId = pc.getAccountId();
                   var desired =
                       applyConnectorSpecPatch(current, request.getSpec(), normalizedMask, corr)
                           .toBuilder()
                           .setUpdatedAt(nowTs())
                           .build();
+                  if (desired.hasDestination()) {
+                    desired =
+                        desired.toBuilder()
+                            .setDestination(
+                                resolveUpdatedDestination(
+                                    desired.getDestination(), accountId, corr))
+                            .build();
+                  }
                   validateConnectorProperties(
                       desired.getKind(), desired.getPropertiesMap(), corr, "properties");
                   validatePersistedAuthConfig(desired.getAuth(), corr);
                   validateReconcilePolicy(desired.getPolicy(), corr);
-                  String accountId = pc.getAccountId();
                   String secretId = connectorId.getId();
                   boolean authTouched =
                       maskTargets(normalizedMask, "auth")
                           || maskTargets(normalizedMask, "auth.credentials");
                   boolean incomingHasCredentials = hasAuthCredentials(desired.getAuth());
-                  Optional<AuthCredentials> priorCredentials = Optional.empty();
+                  boolean shouldDeleteSecret = authTouched && !incomingHasCredentials;
+                  Optional<AuthCredentials> priorCredentials =
+                      authTouched
+                          ? credentialResolver.resolve(accountId, secretId)
+                          : Optional.empty();
                   AuthConfig storedAuth = desired.getAuth();
                   if (incomingHasCredentials) {
-                    priorCredentials = credentialResolver.resolve(accountId, secretId);
                     storedAuth =
                         storeAuthCredentials(desired.getAuth(), accountId, connectorId.getId());
                     if (storedAuth != desired.getAuth()) {
                       desired = desired.toBuilder().setAuth(storedAuth).build();
                     }
+                  } else if (shouldDeleteSecret) {
+                    credentialResolver.delete(accountId, secretId);
                   }
                   ensureNoStoredCredentials(desired.getAuth());
-                  boolean shouldDeleteSecret = authTouched && !incomingHasCredentials;
 
                   if (desired.equals(current)) {
-                    if (shouldDeleteSecret) {
-                      credentialResolver.delete(accountId, secretId);
+                    try {
+                      var metaNoop = connectorRepo.metaFor(connectorId);
+                      boolean callerCares = hasMeaningfulPrecondition(request.getPrecondition());
+                      if (callerCares && metaNoop.getPointerVersion() != meta.getPointerVersion()) {
+                        throw GrpcErrors.preconditionFailed(
+                            corr,
+                            GeneratedErrorMessages.MessageKey.VERSION_MISMATCH,
+                            Map.of(
+                                "expected", Long.toString(meta.getPointerVersion()),
+                                "actual", Long.toString(metaNoop.getPointerVersion())));
+                      }
+                      MutationOps.BaseServiceChecks.enforcePreconditions(
+                          corr, metaNoop, request.getPrecondition());
+                      return UpdateConnectorResponse.newBuilder()
+                          .setConnector(maskConnector(current))
+                          .setMeta(metaNoop)
+                          .build();
+                    } catch (RuntimeException failure) {
+                      if (authTouched) {
+                        restoreCredentials(accountId, secretId, priorCredentials);
+                      }
+                      throw failure;
                     }
-                    var metaNoop = connectorRepo.metaFor(connectorId);
-                    boolean callerCares = hasMeaningfulPrecondition(request.getPrecondition());
-                    if (callerCares && metaNoop.getPointerVersion() != meta.getPointerVersion()) {
-                      throw GrpcErrors.preconditionFailed(
-                          corr,
-                          GeneratedErrorMessages.MessageKey.VERSION_MISMATCH,
-                          Map.of(
-                              "expected", Long.toString(meta.getPointerVersion()),
-                              "actual", Long.toString(metaNoop.getPointerVersion())));
-                    }
-                    MutationOps.BaseServiceChecks.enforcePreconditions(
-                        corr, metaNoop, request.getPrecondition());
-                    return UpdateConnectorResponse.newBuilder()
-                        .setConnector(maskConnector(current))
-                        .setMeta(metaNoop)
-                        .build();
                   }
 
                   try {
                     boolean ok = connectorRepo.update(desired, meta.getPointerVersion());
                     if (!ok) {
                       var nowMeta = connectorRepo.metaForSafe(connectorId);
-                      if (incomingHasCredentials) {
-                        restoreCredentials(accountId, secretId, priorCredentials);
-                      }
                       throw GrpcErrors.preconditionFailed(
                           corr,
                           GeneratedErrorMessages.MessageKey.VERSION_MISMATCH,
@@ -639,7 +652,7 @@ public class ConnectorsImpl extends BaseServiceImpl implements Connectors {
                               "actual", Long.toString(nowMeta.getPointerVersion())));
                     }
                   } catch (BaseResourceRepository.NameConflictException nce) {
-                    if (incomingHasCredentials) {
+                    if (authTouched) {
                       restoreCredentials(accountId, secretId, priorCredentials);
                     }
                     throw GrpcErrors.alreadyExists(
@@ -648,7 +661,7 @@ public class ConnectorsImpl extends BaseServiceImpl implements Connectors {
                         Map.of("display_name", desired.getDisplayName()));
                   } catch (BaseResourceRepository.PreconditionFailedException pfe) {
                     var nowMeta = connectorRepo.metaForSafe(connectorId);
-                    if (incomingHasCredentials) {
+                    if (authTouched) {
                       restoreCredentials(accountId, secretId, priorCredentials);
                     }
                     throw GrpcErrors.preconditionFailed(
@@ -657,10 +670,11 @@ public class ConnectorsImpl extends BaseServiceImpl implements Connectors {
                         Map.of(
                             "expected", Long.toString(meta.getPointerVersion()),
                             "actual", Long.toString(nowMeta.getPointerVersion())));
-                  }
-
-                  if (shouldDeleteSecret) {
-                    credentialResolver.delete(accountId, secretId);
+                  } catch (RuntimeException failure) {
+                    if (authTouched) {
+                      restoreCredentials(accountId, secretId, priorCredentials);
+                    }
+                    throw failure;
                   }
 
                   var outMeta = connectorRepo.metaForSafe(connectorId);
@@ -826,6 +840,43 @@ public class ConnectorsImpl extends BaseServiceImpl implements Connectors {
     }
     return auth.getCredentials().getCredentialCase()
         != AuthCredentials.CredentialCase.CREDENTIAL_NOT_SET;
+  }
+
+  private DestinationTarget resolveUpdatedDestination(
+      DestinationTarget destination, String accountId, String corr) {
+    var resolved = destination.toBuilder();
+    if (destination.hasCatalogDisplayName()) {
+      String displayName = destination.getCatalogDisplayName().trim();
+      var catalog = catalogRepo.getByName(accountId, displayName);
+      if (catalog.isEmpty()) {
+        throw GrpcErrors.notFound(
+            corr,
+            GeneratedErrorMessages.MessageKey.CONNECTOR_DESTINATION_CATALOG_NOT_FOUND,
+            Map.of("display_name", displayName));
+      }
+      resolved.setCatalogId(catalog.get().getResourceId());
+    }
+    if (resolved.hasCatalogId() && destination.hasNamespace()) {
+      namespaceRepo
+          .getByPath(
+              accountId,
+              resolved.getCatalogId().getId(),
+              destination.getNamespace().getSegmentsList())
+          .ifPresent(namespace -> resolved.setNamespaceId(namespace.getResourceId()));
+    }
+    if (resolved.hasCatalogId() && resolved.hasNamespaceId() && destination.hasTableDisplayName()) {
+      String displayName = destination.getTableDisplayName().trim();
+      tableRepo
+          .getByName(
+              accountId,
+              resolved.getCatalogId().getId(),
+              resolved.getNamespaceId().getId(),
+              displayName)
+          .ifPresentOrElse(
+              table -> resolved.setTableId(table.getResourceId()),
+              () -> resolved.setTableDisplayName(displayName));
+    }
+    return resolved.build();
   }
 
   private void restoreCredentials(
--- a/service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
+++ b/service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
@@ -300,6 +300,7 @@ public class ConnectorsImpl extends BaseServiceImpl implements Connectors {
 
                   var display = mustNonEmpty(spec.getDisplayName(), "display_name", corr);
                   var uri = mustNonEmpty(spec.getUri(), "uri", corr);
+                  var initialState = initialConnectorState(spec.getState(), corr);
                   validateConnectorProperties(spec.getKind(), spec.getPropertiesMap(), corr);
                   validatePersistedAuthConfig(spec.getAuth(), corr);
                   validateReconcilePolicy(spec.getPolicy(), corr);
@@ -430,7 +431,7 @@ public class ConnectorsImpl extends BaseServiceImpl implements Connectors {
                           .setPolicy(spec.getPolicy())
                           .setCreatedAt(tsNow)
                           .setUpdatedAt(tsNow)
-                          .setState(ConnectorState.CS_ACTIVE);
+                          .setState(initialState);
 
                   if (spec.hasDescription()) builder.setDescription(spec.getDescription());
                   if (spec.hasSource()) builder.setSource(spec.getSource());
@@ -848,6 +849,16 @@ public class ConnectorsImpl extends BaseServiceImpl implements Connectors {
         != AuthCredentials.CredentialCase.CREDENTIAL_NOT_SET;
   }
 
+  static ConnectorState initialConnectorState(ConnectorState requestedState, String corr) {
+    return switch (requestedState) {
+      case CS_UNSPECIFIED, CS_ACTIVE -> ConnectorState.CS_ACTIVE;
+      case CS_PAUSED -> ConnectorState.CS_PAUSED;
+      case CS_DELETING, UNRECOGNIZED ->
+          throw GrpcErrors.invalidArgument(
+              corr, GeneratedErrorMessages.MessageKey.FIELD, Map.of("field", "state"));
+    };
+  }
+
   private DestinationTarget resolveUpdatedDestination(
       DestinationTarget destination, SourceSelector source, String accountId, String corr) {
     var resolved = destination.toBuilder();
@@ -1225,21 +1236,12 @@ public class ConnectorsImpl extends BaseServiceImpl implements Connectors {
       boolean hasCatalogRef =
           inDst.hasCatalogId()
               || (inDst.hasCatalogDisplayName() && !inDst.getCatalogDisplayName().isBlank());
-      boolean hasNamespaceRef =
-          inDst.hasNamespaceId()
-              || (inDst.hasNamespace() && inDst.getNamespace().getSegmentsCount() > 0);
 
       if (!hasCatalogRef) {
         throw GrpcErrors.invalidArgument(
             corr,
             GeneratedErrorMessages.MessageKey.CONNECTOR_MISSING_DESTINATION_CATALOG,
             Map.of("field", "destination.catalog"));
-      }
-      if (!hasNamespaceRef) {
-        throw GrpcErrors.invalidArgument(
-            corr,
-            GeneratedErrorMessages.MessageKey.FIELD,
-            Map.of("field", "destination.namespace"));
       }
       if (inDst.hasCatalogId()) {
         ensureKind(
@@ -1419,14 +1421,6 @@ public class ConnectorsImpl extends BaseServiceImpl implements Connectors {
             corr,
             GeneratedErrorMessages.MessageKey.CONNECTOR_MISSING_DESTINATION_CATALOG,
             Map.of("field", "destination.catalog"));
-      }
-      boolean hasNamespaceRef =
-          d.hasNamespaceId() || (d.hasNamespace() && d.getNamespace().getSegmentsCount() > 0);
-      if (!hasNamespaceRef) {
-        throw GrpcErrors.invalidArgument(
-            corr,
-            GeneratedErrorMessages.MessageKey.FIELD,
-            Map.of("field", "destination.namespace"));
       }
     }
 
--- a/service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
+++ b/service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
@@ -378,7 +378,10 @@ public class ConnectorsImpl extends BaseServiceImpl implements Connectors {
                   if (destB.hasCatalogId()
                       && destB.hasNamespaceId()
                       && dest.hasTableDisplayName()
-                      && !dest.hasTableId()) {
+                      && !dest.hasTableId()
+                      && spec.hasSource()
+                      && spec.getSource().hasTable()
+                      && !spec.getSource().getTable().isBlank()) {
                     String dTbl = dest.getTableDisplayName().trim();
                     var tblOpt =
                         tableRepo.getByName(
@@ -580,12 +583,15 @@ public class ConnectorsImpl extends BaseServiceImpl implements Connectors {
                           .toBuilder()
                           .setUpdatedAt(nowTs())
                           .build();
-                  if (desired.hasDestination()) {
+                  boolean destinationTouched =
+                      maskTargets(normalizedMask, "destination")
+                          || maskTargetsUnder(normalizedMask, "destination");
+                  if (destinationTouched && desired.hasDestination()) {
                     desired =
                         desired.toBuilder()
                             .setDestination(
                                 resolveUpdatedDestination(
-                                    desired.getDestination(), accountId, corr))
+                                    desired.getDestination(), desired.getSource(), accountId, corr))
                             .build();
                   }
                   validateConnectorProperties(
@@ -843,7 +849,7 @@ public class ConnectorsImpl extends BaseServiceImpl implements Connectors {
   }
 
   private DestinationTarget resolveUpdatedDestination(
-      DestinationTarget destination, String accountId, String corr) {
+      DestinationTarget destination, SourceSelector source, String accountId, String corr) {
     var resolved = destination.toBuilder();
     if (destination.hasCatalogDisplayName()) {
       String displayName = destination.getCatalogDisplayName().trim();
@@ -864,7 +870,11 @@ public class ConnectorsImpl extends BaseServiceImpl implements Connectors {
               destination.getNamespace().getSegmentsList())
           .ifPresent(namespace -> resolved.setNamespaceId(namespace.getResourceId()));
     }
-    if (resolved.hasCatalogId() && resolved.hasNamespaceId() && destination.hasTableDisplayName()) {
+    if (resolved.hasCatalogId()
+        && resolved.hasNamespaceId()
+        && destination.hasTableDisplayName()
+        && source.hasTable()
+        && !source.getTable().isBlank()) {
       String displayName = destination.getTableDisplayName().trim();
       tableRepo
           .getByName(
--- a/service/src/main/java/ai/floedb/floecat/service/security/RolePermissions.java
+++ b/service/src/main/java/ai/floedb/floecat/service/security/RolePermissions.java
@@ -47,6 +47,7 @@ public final class RolePermissions {
           "view.read",
           "view.write",
           "connector.manage",
+          "connector.export",
           "system-objects.read",
           "account.delete");
   private static final List<String> PLATFORM_PERMS =
--- a/service/src/test/java/ai/floedb/floecat/service/connector/impl/ConnectorsImplExportConnectorTest.java
+++ b/service/src/test/java/ai/floedb/floecat/service/connector/impl/ConnectorsImplExportConnectorTest.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2026 Yellowbrick Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ai.floedb.floecat.service.connector.impl;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import ai.floedb.floecat.common.rpc.PrincipalContext;
+import ai.floedb.floecat.common.rpc.ResourceId;
+import ai.floedb.floecat.common.rpc.ResourceKind;
+import ai.floedb.floecat.connector.rpc.AuthConfig;
+import ai.floedb.floecat.connector.rpc.AuthCredentials;
+import ai.floedb.floecat.connector.rpc.Connector;
+import ai.floedb.floecat.connector.rpc.ExportConnectorRequest;
+import ai.floedb.floecat.connector.spi.CredentialResolver;
+import ai.floedb.floecat.service.repo.impl.ConnectorRepository;
+import ai.floedb.floecat.service.security.impl.Authorizer;
+import ai.floedb.floecat.service.security.impl.PrincipalProvider;
+import java.lang.reflect.Field;
+import java.util.Optional;
+import org.junit.jupiter.api.Test;
+
+class ConnectorsImplExportConnectorTest {
+  @Test
+  void exportsUnmaskedConnectorAndSeparateCredentials() throws Exception {
+    var service = new ConnectorsImpl();
+    service.connectorRepo = mock(ConnectorRepository.class);
+    service.principalProvider = mock(PrincipalProvider.class);
+    service.authz = mock(Authorizer.class);
+    service.credentialResolver = mock(CredentialResolver.class);
+    installBasePrincipal(service, service.principalProvider);
+
+    var connectorId =
+        ResourceId.newBuilder()
+            .setAccountId("acct")
+            .setId("connector-1")
+            .setKind(ResourceKind.RK_CONNECTOR)
+            .build();
+    var connector =
+        Connector.newBuilder()
+            .setResourceId(connectorId)
+            .setDisplayName("upstream")
+            .setAuth(AuthConfig.newBuilder().putProperties("client_id", "not-masked"))
+            .build();
+    var credentials =
+        AuthCredentials.newBuilder()
+            .setBearer(AuthCredentials.BearerToken.newBuilder().setToken("secret-token"))
+            .build();
+    var principal =
+        PrincipalContext.newBuilder()
+            .setAccountId("acct")
+            .setCorrelationId("corr-1")
+            .addPermissions("connector.export")
+            .build();
+
+    when(service.principalProvider.get()).thenReturn(principal);
+    when(service.connectorRepo.getById(connectorId)).thenReturn(Optional.of(connector));
+    when(service.credentialResolver.resolve("acct", "connector-1"))
+        .thenReturn(Optional.of(credentials));
+
+    var response =
+        service
+            .exportConnector(
+                ExportConnectorRequest.newBuilder()
+                    .setConnectorId(connectorId)
+                    .setIncludeCredentials(true)
+                    .build())
+            .await()
+            .indefinitely();
+
+    assertThat(response.getConnector()).isEqualTo(connector);
+    assertThat(response.getCredentials()).isEqualTo(credentials);
+    verify(service.authz).require(principal, "connector.export");
+  }
+
+  private static void installBasePrincipal(
+      ConnectorsImpl service, PrincipalProvider principalProvider) {
+    try {
+      Field field =
+          ai.floedb.floecat.service.common.BaseServiceImpl.class.getDeclaredField("principal");
+      field.setAccessible(true);
+      field.set(service, principalProvider);
+    } catch (ReflectiveOperationException e) {
+      throw new AssertionError("Failed to inject BaseServiceImpl principal provider", e);
+    }
+  }
+}
--- a/tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorArchive.java
+++ b/tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorArchive.java
@@ -1,0 +1,196 @@
+/*
+ * Copyright 2026 Yellowbrick Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ai.floedb.floecat.tools.transfer;
+
+import ai.floedb.floecat.connector.rpc.ConnectorTransferBundle;
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.util.JsonFormat;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.nio.file.attribute.PosixFilePermission;
+import java.util.EnumSet;
+import java.util.Set;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+import java.util.zip.ZipOutputStream;
+
+final class ConnectorArchive {
+  static final int FORMAT_VERSION = 1;
+  private static final int MAX_BUNDLE_BYTES = 64 * 1024 * 1024;
+
+  private ConnectorArchive() {}
+
+  static void write(Path output, ConnectorTransferBundle bundle, boolean force) throws IOException {
+    Path absolute = output.toAbsolutePath();
+    if (Files.exists(absolute) && !force) {
+      throw new IOException("output already exists (use --force): " + absolute);
+    }
+    Path parent = absolute.getParent();
+    if (parent != null) Files.createDirectories(parent);
+    Path temporary = Files.createTempFile(parent, ".floecat-connectors-", ".tmp");
+    try {
+      setOwnerOnly(temporary);
+      try (var zip = new ZipOutputStream(Files.newOutputStream(temporary))) {
+        put(zip, "bundle.pb", bundle.toByteArray());
+        put(zip, "manifest.json", manifest(bundle).getBytes(StandardCharsets.UTF_8));
+        for (var entry : bundle.getEntriesList()) {
+          String base =
+              "connectors/" + safeSegment(entry.getConnector().getResourceId().getId()) + "/";
+          put(zip, base + "connector.pb", entry.getConnector().toByteArray());
+          put(
+              zip,
+              base + "portable-spec.json",
+              protobufJson(entry.getPortableSpec()).getBytes(StandardCharsets.UTF_8));
+          if (entry.hasCredentials()) {
+            put(zip, base + "credentials.pb", entry.getCredentials().toByteArray());
+          }
+        }
+      }
+      if (force) {
+        Files.move(
+            temporary,
+            absolute,
+            StandardCopyOption.REPLACE_EXISTING,
+            StandardCopyOption.ATOMIC_MOVE);
+      } else {
+        Files.move(temporary, absolute, StandardCopyOption.ATOMIC_MOVE);
+      }
+      setOwnerOnly(absolute);
+    } finally {
+      Files.deleteIfExists(temporary);
+    }
+  }
+
+  static ConnectorTransferBundle read(Path input) throws IOException {
+    try (var zip = new ZipFile(input.toFile())) {
+      ZipEntry entry = zip.getEntry("bundle.pb");
+      if (entry == null || entry.isDirectory()) {
+        throw new IOException("connector archive is missing bundle.pb");
+      }
+      byte[] bytes;
+      try (InputStream in = zip.getInputStream(entry)) {
+        bytes = readBounded(in, MAX_BUNDLE_BYTES);
+      }
+      try {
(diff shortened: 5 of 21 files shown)


tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
```diff
@@ -0,0 +1,433 @@
...
+              client.connectors.validateConnector(
+                  ValidateConnectorRequest.newBuilder().setSpec(spec).build());
+          if (!validation.getOk()) {
+            throw new IllegalArgumentException(
+                "connector validation failed for "
+                    + spec.getDisplayName()
+                    + ": "
+                    + validation.getSummary());
+          }
+          if (dryRun) {
```
mrouvroy-floe: [major] **`--dry-run` ignores conflicts**

`--dry-run` exits the loop immediately after connector validation, before the existing-name conflict handling runs. With the default `--conflict FAIL`, a dry run prints `VALID ...` for a connector whose display name already exists even though the real import would throw `connector already exists`.

That makes dry runs non-predictive for the main destructive decision this command is supposed to preview. Operators can get a clean dry-run result and then have the real import fail, or miss that `--conflict REPLACE` would delete an existing connector.

Move the conflict lookup/handling ahead of the `dryRun` `continue`, and make dry-run output reflect the selected conflict mode, for example `would fail`, `would skip`, or `would replace`, without mutating anything.

---

🤖 _Generated with GPT-5.5._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
```diff
@@ -0,0 +1,433 @@
...
+    if (!missing.isEmpty()) {
+      throw new IllegalArgumentException("unknown connector(s): " + String.join(", ", missing));
+    }
+    return selected;
+  }
+
+  private static String idempotencyKey(
+      ConnectorTransferBundle bundle, ConnectorTransferEntry entry, String replacementId)
+      throws Exception {
+    String source =
```
mrouvroy-floe, line 451: [major] **Stable idempotency key can skip creates**

`idempotencyKey(...)` is stable for the same source account, source connector ID, display name, and replacement ID. For an import into an empty target, `replacementId` is `""`, so re-importing the same archive within the server idempotency TTL reuses the previous `CreateConnector` key.

`CreateConnector` returns the cached successful payload for a repeated idempotency key/request instead of running the creator again. If the connector from the first import was deleted, the CLI sees no existing connector, calls `createConnector`, receives the cached old connector, prints success, and leaves the target account without the connector. If the archive content changed under the same key, the server returns an idempotency mismatch instead.

Do not use a deterministic cross-run key here, or include an import-run nonce and retry only within that run. If the tool keeps server idempotency, verify that the returned connector actually exists in the target before counting the entry as imported.

---

🤖 _Generated with GPT-5.5._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
mcusack-floe: Fixed in 91809c5c. Conflict evaluation now happens before the dry-run branch. Dry runs report WOULD CREATE, WOULD FAIL, WOULD SKIP, or WOULD REPLACE; maintain a simulated name map across archive entries; perform no delete/create operations; and return a nonzero status when FAIL conflicts are found. Added focused conflict-mode unit coverage, and the connector-transfer tests pass (5/5).

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
mcusack-floe: Fixed in 91809c5c. Each import invocation now generates a UUID nonce that is included in the hashed CreateConnector idempotency key. Keys remain stable within one invocation but differ across separate imports, preventing a later run from receiving a cached payload for a connector that has since been deleted. Added a unit test covering same-run stability and cross-run uniqueness; the connector-transfer tests pass (5/5).

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
```diff
@@ -0,0 +1,483 @@
...
+                throw new IllegalArgumentException(
+                    "connector already exists: " + spec.getDisplayName());
+            case SKIP -> {
+              System.out.println("Skipped existing connector " + spec.getDisplayName());
+              skipped++;
+              continue;
+            }
+            case REPLACE -> {
+              replacementId = current.getResourceId().getId();
+              client.connectors.deleteConnector(
```
mrouvroy-floe: [major] **Make `REPLACE` failure-safe**

`--conflict REPLACE` deletes the existing connector before `createConnector` and `restoreState` have succeeded.

That leaves a real data-loss path: `validateConnector` does not cover all `CreateConnector` checks, and the later create/state-restore can still fail because of missing destination objects, required source/destination invariants, repository conflicts, credential storage errors, or a transient RPC failure. In all of those cases the target connector and its credentials have already been removed.

Use an update-in-place replacement path, or otherwise make the delete/create sequence failure-safe with full preflight validation and rollback before deleting the existing connector.

---

🤖 _Generated with GPT-5.5._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/PortableConnectorSpecs.java
```diff
@@ -0,0 +1,93 @@
...
+      spec.setDestination(portableDestination(connector.getDestination(), directory));
+    }
+    return spec.build();
+  }
+
+  private static DestinationTarget portableDestination(
+      DestinationTarget destination, DirectoryServiceGrpc.DirectoryServiceBlockingStub directory) {
+    var portable = DestinationTarget.newBuilder();
+    if (destination.hasCatalogId()) {
+      portable.setCatalogDisplayName(
```
mrouvroy-floe: [major] **Fail on unresolved destination refs**

`portableDestination` assumes every reverse lookup succeeds and immediately copies the returned name/path into the archive.

The directory lookup RPCs return empty responses when a `catalog_id`, `namespace_id`, or `table_id` cannot be resolved. This code then records an empty `catalog_display_name`, empty namespace path, or empty table name, so export appears successful but produces a non-portable archive that either fails later during import or imports with a degraded destination.

Check each lookup response for a nonblank catalog display name, `hasRef()` with nonempty namespace segments, and `hasName()` with a nonblank table name. If any ID cannot be resolved, fail the export with a clear error instead of writing a broken archive.

---

🤖 _Generated with GPT-5.5._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
mcusack-floe: Fixed in c749addd. REPLACE no longer deletes the existing connector: it sends one full-field UpdateConnector request against the existing resource ID, including state and credentials. The import now preflights every archive entry and every conflict before performing any mutation. The shared update service also resolves portable destination references before persistence and rolls credential replacement/removal back if the connector repository update fails. Added focused replacement, unresolved-destination, and credential rollback tests. Connector-transfer tests pass (12/12) and service safety tests pass (3/3).

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/PortableConnectorSpecs.java
mcusack-floe: Fixed in c749addd. Export now requires catalog lookups to return a nonblank display name, namespace lookups to return a populated ref with nonblank segments, and table lookups to return a nonblank table name. Any unresolved destination ID aborts export before an archive is written. Existing name/path references are also checked for blank values. Added positive conversion coverage plus focused unresolved catalog, namespace, and table tests; connector-transfer tests pass (12/12).

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
```diff
@@ -520,63 +574,76 @@ public Uni<UpdateConnectorResponse> updateConnector(UpdateConnectorRequest reque
                                       Map.of("id", connectorId.getId())));
 
                   FieldMask normalizedMask = normalizeMask(request.getUpdateMask());
+                  String accountId = pc.getAccountId();
                   var desired =
                       applyConnectorSpecPatch(current, request.getSpec(), normalizedMask, corr)
                           .toBuilder()
                           .setUpdatedAt(nowTs())
                           .build();
+                  if (desired.hasDestination()) {
```
mrouvroy-floe: [major] **Do not resolve an untouched destination**

This runs for every update on a connector that has a destination, even when the update mask only changes an unrelated field. It consequently turns retained name references into resource IDs as a side effect of, for example, a `uri` update.

For a namespace-discovery connector, reconciliation can create the destination table while the connector still stores `destination.table_display_name`. The next unrelated update finds that table and changes the stored destination to `table_id`; the reconciler then rejects it because such connectors intentionally have no `source.table`.

This is already failing `ConnectorIT.icebergFixtureIncrementalReconcileCapturesStatsForNewSnapshot` and `...CapturesDeleteStatsForNewSnapshot` in both failing CI jobs, with `Pinned destination table id requires connector.source.table`. Only resolve/materialize destination references when the update mask changes destination fields (and preserve name-based targets for namespace-discovery connectors).

---

🤖 _Generated with GPT-5.6 Terra._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
```diff
@@ -0,0 +1,520 @@
...
+        String importRunNonce = UUID.randomUUID().toString();
+        List<PreparedImport> prepared = new ArrayList<>();
+        for (var entry : bundle.getEntriesList()) {
+          ConnectorSpec spec = specWithCredentials(entry);
+          if (spec.getState() == ConnectorState.CS_DELETING) {
+            throw new IllegalArgumentException(
+                "cannot import connector in deleting state: " + spec.getDisplayName());
+          }
+          var validation =
+              client.connectors.validateConnector(
```
mrouvroy-floe: [major] **Validate portable destinations during dry runs**

`--dry-run` reports an import action after calling `ValidateConnector`, but that RPC validates the upstream connector configuration only; it never reads or resolves `spec.destination`. A bundle whose `catalog_display_name` does not exist in the target account can therefore be reported as `WOULD CREATE` (or `WOULD REPLACE`) and then fail as soon as the live import reaches `CreateConnector`/`UpdateConnector`.

That makes the advertised dry run non-predictive for one of the main portability concerns: resolving destination names in the target account.

Add a destination-resolution preflight before producing dry-run actions, ideally by sharing the service-side destination validation or explicitly resolving the catalog/namespace/table through the directory API, so the dry-run result matches the mutation path.

---

🤖 _Generated with GPT-5.6 Terra._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
```diff
@@ -0,0 +1,520 @@
...
+            skipped++;
+            continue;
+          }
+
+          Connector importedConnector;
+          if (item.action() == ImportAction.REPLACE) {
+            importedConnector =
+                client
+                    .connectors
+                    .updateConnector(replacementRequest(item.current(), item.spec()))
```
mrouvroy-floe, line 302: [major] **Allow replacement of catalog-only destinations**

The `REPLACE` path always sends the full `destination` mask. `UpdateConnector` rejects a full destination without a namespace, but `CreateConnector` only requires a destination catalog. Consequently, an archive exported from a valid catalog-only connector imports successfully when the target name is new, yet fails when the same name exists and `--conflict REPLACE` is selected.

This is a real valid connector shape: the reconciler falls back to the source namespace when no destination namespace is specified, and `CreateConnector` explicitly permits it.

Make full-destination update validation accept the same catalog-only form as create, or build the replacement mask so absent optional destination fields are preserved rather than treated as an invalid complete replacement.

---

🤖 _Generated with GPT-5.6 Terra._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
```diff
@@ -0,0 +1,520 @@
...
+          .build();
+
+  @CommandLine.Option(names = "--host", description = "gRPC host (default: ${DEFAULT-VALUE})")
+  String host = env("FLOECAT_GRPC_HOST", "localhost");
+
+  @CommandLine.Option(names = "--port", description = "gRPC port (default: ${DEFAULT-VALUE})")
+  int port = envInt("FLOECAT_GRPC_PORT", 9100);
+
+  @CommandLine.Option(names = "--tls", description = "Use TLS for the gRPC connection")
+  boolean tls;
```
mrouvroy-floe: [major] **Bearer token and plaintext credentials sent over unencrypted channel by default**

The channel defaults to `usePlaintext()` because `--tls` is off by default (`ConnectorTransferCli.java:88`, `Client` at `:487`). The export path always requests `setIncludeCredentials(true)`, so unmasked `AuthCredentials` — including bearer/session tokens and plaintext connector secrets — traverse the network in the clear when the tool is pointed at a non-local host.

This directly contradicts the care taken elsewhere in this PR (owner-only archive permissions, the `--plaintext-secrets` acknowledgement). The README examples explicitly use `--host target.example` without `--tls`, making this the de-facto default path for remote use.

Default `--tls` to on, or at minimum refuse a plaintext channel when the host is not loopback and credentials are being transferred.

---

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorArchive.java
```diff
@@ -0,0 +1,229 @@
...
+    }
+    return out.toString();
+  }
+
+  private static void setOwnerOnly(Path path) {
+    try {
+      Set<PosixFilePermission> permissions =
+          EnumSet.of(PosixFilePermission.OWNER_READ, PosixFilePermission.OWNER_WRITE);
+      Files.setPosixFilePermissions(path, permissions);
+    } catch (IOException | UnsupportedOperationException ignored) {
```
mrouvroy-floe: [major] **Archive permission failures are silently ignored**

`setOwnerOnly(...)` swallows both `IOException` and `UnsupportedOperationException`. If the filesystem does not support POSIX permissions or if setting them fails, the command still writes an archive that may contain plaintext credentials without the owner-only protection the tool promises.

The caller sees a successful export with no signal that file permissions were not tightened — this is a fail-open security issue for a secret-bearing archive.

Fail closed when owner-only permissions cannot be applied, or implement a platform-appropriate fallback (e.g. Windows ACLs) and emit a clear, fatal error when no protection mechanism is available.

---

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
mrouvroy-floe: [major] **Import preflight validation runs before conflict resolution, breaking `--conflict SKIP` and `--dry-run`**

`import` runs `ValidateConnector` on every archive entry before determining the conflict action. This means a connector that already exists and would be skipped under `--conflict SKIP` can still fail the import if its archived upstream credentials no longer validate — a connector that should be left untouched is rejected.

`--dry-run` is similarly affected: `ValidateConnector` does not resolve `spec.destination`, so a bundle with a missing `catalog_display_name` can report `WOULD CREATE` or `WOULD REPLACE` and then fail in the real `CreateConnector`/`UpdateConnector` call, undermining the predictive value of the dry run.

Compute the conflict action first. Skip validation entirely for entries whose action is `SKIP`. For entries that will be created or replaced, scope the validation to what the corresponding RPC actually checks.

---

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
mrouvroy-floe, line 302: [major] **Catalog-only replacement fails under `--conflict REPLACE`**

`REPLACE` always sends a full `destination` update through `replacementRequest(...)`, which routes through `UpdateConnector`'s full-destination validation. That validation requires a namespace, whereas `CreateConnector` only requires a destination catalog.

The result is that an archive exported from a valid catalog-only connector imports successfully when the connector name is new, but fails with a validation error when the same name already exists and `--conflict REPLACE` is active. The create-time contract (catalog-only is valid) is not preserved at update time.

Either make `UpdateConnector`'s full-destination validation accept the same catalog-only shape as `CreateConnector`, or have `replacementRequest(...)` omit absent optional destination fields rather than treating the portable spec as a complete-but-invalid destination.

---

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
```diff
@@ -0,0 +1,520 @@
...
+          existing.put(connector.getDisplayName(), connector);
+        }
+
+        int imported = 0;
+        int skipped = 0;
+        int wouldCreate = 0;
+        int wouldReplace = 0;
+        int wouldSkip = 0;
+        int wouldFail = 0;
+        String importRunNonce = UUID.randomUUID().toString();
```
mrouvroy-floe, line 240: [minor] **Randomized idempotency key per run defeats cross-run deduplication**

`importRunNonce` is a fresh `UUID` generated once per invocation (`:238`) and mixed into every entry's `idempotencyKey` (`:441`-`:455`). Because this value changes on every run, the server-side idempotency mechanism can never deduplicate the same archive entry across two separate runs.

Concretely: if a `createConnector` call fails at the network layer after the server committed it, re-running the import finds the connector already exists and, under the default `--conflict FAIL`, errors out — whereas a deterministic key derived from archive content would allow the retried create to no-op. Repeatability currently rests entirely on the pre-loop display-name existence check, which is also subject to a TOCTOU race.

If per-run isolation is genuinely intended, document it explicitly. Otherwise, derive the idempotency key deterministically from the archive content (e.g. a hash of the entry's serialized bytes) so that retried runs are truly safe to repeat.

---

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorArchive.java
```diff
@@ -0,0 +1,229 @@
...
+  private static String manifest(ConnectorTransferBundle bundle) {
+    var out = new StringBuilder();
+    out.append("{\n  \"formatVersion\": ").append(bundle.getFormatVersion());
+    out.append(",\n  \"sourceAccountId\": \"")
+        .append(jsonEscape(bundle.getSourceAccountId()))
+        .append("\",");
+    out.append("\n  \"exportedAt\": \"")
+        .append(bundle.getExportedAt().getSeconds())
+        .append(".")
+        .append(bundle.getExportedAt().getNanos())
```
mrouvroy-floe: [minor] **`exportedAt` in manifest.json emits malformed fractional seconds**

The manifest timestamp concatenates `seconds`, a literal `.`, and raw `nanos` (`:174`-`:178`). `nanos` is not zero-padded to 9 digits, so `nanos = 42000000` renders as `...123.42000000Z` and `nanos = 5` renders as `...123.5Z` — both misrepresent the sub-second value and produce strings that are neither valid RFC-3339 nor internally consistent.

This is cosmetic (the authoritative timestamp lives in `bundle.pb`), but the informational JSON field is misleading.

Fix with `String.format("%d.%09dZ", seconds, nanos)`, or reuse the `Timestamps`/`JsonFormat` protobuf rendering already relied on elsewhere in the archive.

---

jchauzi-floe (commented):
I found four more issues after checking these against the existing review threads.

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
```diff
@@ -0,0 +1,520 @@
...
+
+  private static List<Connector> select(List<Connector> all, List<String> requested) {
+    all = all.stream().sorted(Comparator.comparing(Connector::getDisplayName)).toList();
+    if (requested == null || requested.isEmpty()) return all;
+    Set<String> wanted = new HashSet<>(requested);
+    List<Connector> selected =
+        all.stream()
+            .filter(
+                connector ->
+                    wanted.contains(connector.getDisplayName())
```
jchauzi-floe: [major] **A selector can match two connectors**

Each requested value is checked against both the display name and resource ID. Display names only need to be nonblank, so one connector can be named after another connector's UUID. In that case, `--connector <uuid>` selects both and puts both sets of credentials in the archive. That's a nasty surprise for a command meant to limit an export.

I think an ID match should win. If the value matches an ID, select only that connector. Otherwise, use the display-name match. Rejecting the ambiguity before writing would work too.

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
```diff
@@ -0,0 +1,520 @@
...
+                    .connectors
+                    .createConnector(
+                        CreateConnectorRequest.newBuilder()
+                            .setSpec(item.spec())
+                            .setIdempotency(
+                                IdempotencyKey.newBuilder()
+                                    .setKey(idempotencyKey(importRunNonce, bundle, item.entry())))
+                            .build())
+                    .getConnector();
+            restoreState(client.connectors, importedConnector, item.spec().getState());
```
jchauzi-floe: [major] **A failed state restore leaves the connector active**

`CreateConnector` always persists `CS_ACTIVE`. For an archived `CS_PAUSED` connector, this second call is the only thing that restores its state. If it fails after create has committed, the CLI exits with an error and leaves the connector active. And the default retry hits the name conflict, so the operator cannot simply rerun the import. The scheduler may reconcile a connector that was deliberately paused at the source.

Can `CreateConnector` honor the requested paused state atomically? Otherwise this path needs recovery that cannot strand the new connector as active.

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorArchive.java
```diff
@@ -0,0 +1,229 @@
...
+    if (Files.exists(absolute) && !force) {
+      throw new IOException("output already exists (use --force): " + absolute);
+    }
+    Path parent = absolute.getParent();
+    if (parent != null) Files.createDirectories(parent);
+    Path temporary = Files.createTempFile(parent, ".floecat-connectors-", ".tmp");
+    try {
+      setOwnerOnly(temporary);
+      try (var zip = new ZipOutputStream(Files.newOutputStream(temporary))) {
+        put(zip, "bundle.pb", bundle.toByteArray());
```
jchauzi-floe, line 57: [moderate] **The writer can produce an archive the reader rejects**

`write()` has no check matching the reader's 64 MiB limit. A default export can collect enough connectors and credentials for `bundle.pb` to exceed `MAX_BUNDLE_BYTES`, publish the zip successfully, and then fail every `inspect` or `import` in `readBounded()`.

Please enforce the same limit before moving the temporary file, perhaps with `bundle.getSerializedSize()`, or support the larger bundle in both directions. A successful export should be readable by the same version of the tool.

service/src/main/java/ai/floedb/floecat/service/security/RolePermissions.java
```diff
@@ -47,6 +47,7 @@ public final class RolePermissions {
           "view.read",
           "view.write",
           "connector.manage",
+          "connector.export",
```
jchauzi-floe, line 50: [minor] **The fixed-role documentation is now stale**

`connector.export` is added to the administrator and developer grants here, but `docs/fixed-roles.md` still lists the exact permissions for those roles without it. That document is linked as the permission contract from `docs/api-access.md`, so operators will get the wrong answer about who can call this secret-bearing RPC.

Please update both role rows. The `Connectors` summaries in `docs/proto.md` and `docs/service.md` could mention `ExportConnector` as well.

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
mcusack-floe: Fixed in 70d24528a. Import now preflights destination catalog resolution through DirectoryService for CREATE and REPLACE actions before reporting dry-run results, so a missing target catalog fails the dry run instead of producing WOULD CREATE or WOULD REPLACE. Added focused resolvable and unresolved destination-preflight coverage.

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
mcusack-floe: Fixed in 70d24528a. Full destination updates now require the same minimum shape as create: a catalog reference, with namespace and table remaining optional. Catalog-only connectors can therefore be replaced in place. Added a service regression test for a full catalog-only destination update.

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
mcusack-floe: Fixed in 70d24528a. TLS is now the default transport. Plaintext requires an explicit --plaintext opt-in intended for local development, and the README examples and security guidance were updated. Added CLI option coverage for the secure default and explicit opt-in.

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorArchive.java
mcusack-floe: Fixed in 70d24528a. Owner-only permission failures are no longer swallowed: IOException propagates and unsupported POSIX permissions produce a fatal IOException before archive publication. The temporary and final archive paths are both protected, with regression coverage asserting owner read/write only.

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
mcusack-floe: Fixed in 70d24528a. Conflict action is computed before mutation validation. FAIL and SKIP entries no longer run upstream or destination validation, while CREATE and REPLACE entries run destination preflight plus ValidateConnector. Added focused tests proving skipped and failed entries do not invoke mutation validation.

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
mcusack-floe: Fixed in 70d24528a. UpdateConnector full-destination validation now accepts catalog-only targets, matching CreateConnector. The REPLACE request can keep its complete destination mask without rejecting a valid connector shape. Covered by a catalog-only full-update regression test.

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorArchive.java
mcusack-floe: Fixed in 70d24528a. manifest.json now renders exportedAt with protobuf Timestamps.toString, preserving nanosecond precision and producing canonical RFC 3339 output. Added coverage for a five-nanosecond timestamp.

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
mcusack-floe: Fixed in 70d24528a. Selector resolution now checks connector IDs first and falls back to display names only when no ID matches. Selections are deduplicated, and a regression test covers a display name equal to another connector ID.

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
mcusack-floe: Fixed in 70d24528a. CreateConnector now persists CS_PAUSED atomically when requested and rejects CS_DELETING. The CLI follow-up state-restoration call was removed, eliminating the active-connector window and second-call failure path. Added service coverage that captures the persisted connector and verifies the response remains paused.

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorArchive.java
mcusack-floe: Fixed in 70d24528a. ConnectorArchive.write now checks the protobuf serialized size against the same 64 MiB limit used by the reader before creating or publishing the archive. Added boundary-failure coverage.

service/src/main/java/ai/floedb/floecat/service/security/RolePermissions.java
mcusack-floe: Fixed in 70d24528a. The administrator and developer rows in docs/fixed-roles.md now include connector.export. The Connectors summaries in docs/proto.md and docs/service.md also document ExportConnector and its dedicated permission.

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/PortableConnectorSpecs.java
```diff
@@ -0,0 +1,161 @@
...
+    if (connector.hasAuth()) spec.setAuth(connector.getAuth());
+    if (connector.hasDestination()) {
+      spec.setDestination(portableDestination(connector.getDestination(), directory));
+    }
+    return spec.build();
+  }
+
+  static void validateForImport(
+      ConnectorSpec spec, DirectoryServiceGrpc.DirectoryServiceBlockingStub directory) {
+    if (!spec.hasDestination()) {
```
mrouvroy-floe, line 55: [major] **Validate required `source.namespace`**

`validateForImport(...)` validates only the destination, while `ValidateConnector` skips source checks when `spec.hasSource()` is false. An archive whose portable spec omits `source.namespace` therefore reaches the dry-run output as `WOULD CREATE` or `WOULD REPLACE` when its upstream connection succeeds.

Both mutation paths reject that same spec: `CreateConnector` requires a populated source namespace, and the full-field replacement mask makes `UpdateConnector` enforce the same invariant. This leaves the advertised dry run non-predictive for malformed, hand-built, or older archives.

Add the same `source.namespace` presence check to import preflight, or centralize the create/replace structural validation so `ValidateConnector` and the mutation RPCs cannot drift.

---

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/PortableConnectorSpecs.java
```diff
@@ -0,0 +1,161 @@
...
+      String displayName = destination.getCatalogDisplayName().trim();
+      if (displayName.isBlank()) {
+        throw new IllegalArgumentException("destination catalog display name is blank");
+      }
+      var response =
+          directory.resolveCatalog(
+              ResolveCatalogRequest.newBuilder()
+                  .setRef(NameRef.newBuilder().setCatalog(displayName))
+                  .build());
+      if (!response.hasResourceId() || response.getResourceId().getId().isBlank()) {
```
jchauzi-floe, line 69: [minor] **Missing-catalog check is dead code: `ResolveCatalog` throws `NOT_FOUND`, never returns empty**

`validateForImport` treats an unresolvable destination catalog as a `ResolveCatalogResponse` without a `resource_id` and converts it into the friendly `IllegalArgumentException`. But `DirectoryServiceImpl.resolveCatalog` never returns an empty response — for a missing catalog it throws `GrpcErrors.notFound(...)` (`DirectoryServiceImpl.java:83-88`), which the blocking stub surfaces as a `StatusRuntimeException`.

Two consequences:

- Against a real server, a missing target catalog aborts `import`/`--dry-run` with a raw gRPC `NOT_FOUND` error from inside the prepare loop instead of the intended `destination catalog could not be resolved: <name>` message. In dry-run mode this also means the run dies on the first bad entry rather than completing the per-entry `WOULD ...` report the dry run promises.
- The unit test `rejectsUnresolvedCatalogBeforeImport` mocks a response shape (`ResolveCatalogResponse.getDefaultInstance()`) the server can't produce, so it green-lights a code path that never executes in production.

Still fails safe (no mutation happens either way), but the fix is small: catch `StatusRuntimeException` with `Status.Code.NOT_FOUND` around the `resolveCatalog` call and rethrow the friendly `IllegalArgumentException`, keeping the empty-response check as a fallback. Ideally `ImportCommand` would also collect per-entry preflight failures during `--dry-run` and report them as a `WOULD FAIL validation` line instead of aborting on the first one.

---

🤖 _Generated with Fable 5._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
```diff
@@ -0,0 +1,536 @@
...
+    }
+    return new PreparedImport(entry, spec, current, action, validationSummary);
+  }
+
+  private static String validateImport(Client client, ConnectorSpec spec) {
+    if (spec.getState() == ConnectorState.CS_DELETING) {
+      throw new IllegalArgumentException(
+          "cannot import connector in deleting state: " + spec.getDisplayName());
+    }
+    PortableConnectorSpecs.validateForImport(spec, client.directory);
```
jchauzi-floe, line 413: [minor] **Import preflight does not cover the `source.namespace` invariant**

`validateImport` checks the state, the destination (`PortableConnectorSpecs.validateForImport`), and upstream connectivity (`ValidateConnector`). None of those validate `source`: `ValidateConnector` only uses `source` opportunistically for a best-effort `describe`, and `validateForImport` looks at the destination only.

Both real mutations do enforce it: `CreateConnector` rejects a spec without a non-empty `source.namespace` (`CONNECTOR_MISSING_SOURCE_NAMESPACE`), and the REPLACE path always includes `source` in `REPLACE_UPDATE_MASK`, so `applyConnectorSpecPatch` rejects it too. An archive entry lacking `source.namespace` (hand-assembled or edited — server-exported archives always carry it) is therefore reported as `WOULD CREATE`/`WOULD REPLACE` by `--dry-run`, and a real import fails mid-loop after earlier entries have already been created or replaced — exactly the partial-import outcome the preflight was introduced to prevent.

Adding a cheap client-side check in `validateImport` (non-empty `source.namespace` segments, mirroring the server rule) closes the parity gap:

```java
if (!spec.hasSource()
    || spec.getSource().getNamespace().getSegmentsCount() == 0) {
  throw new IllegalArgumentException(
      "connector source namespace is required: " + spec.getDisplayName());
}
```

---

🤖 _Generated with Fable 5._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
```diff
@@ -0,0 +1,536 @@
...
+    Client(
+        String host, int port, boolean tls, String token, String sessionToken, String accountId) {
+      var builder = ManagedChannelBuilder.forAddress(host, port);
+      if (tls) builder.useTransportSecurity();
+      else builder.usePlaintext();
+      managedChannel = builder.build();
+      Channel channel =
+          ClientInterceptors.intercept(
+              managedChannel, new GrpcHeaders(token, sessionToken, accountId));
+      connectors = ConnectorsGrpc.newBlockingStub(channel);
```
jchauzi-floe, line 505: [nit] **No deadlines on the blocking stubs**

`Client` creates `ConnectorsBlockingStub` and `DirectoryServiceBlockingStub` without any deadline, and no call site adds one. A stalled connection or hung server leaves the CLI blocked forever — worst during `import`, where a hang mid-mutation-loop gives the operator no indication of how much of the archive was applied.

A per-call deadline (e.g. `connectors.withDeadlineAfter(60, TimeUnit.SECONDS)`, longer for `ValidateConnector` since it contacts the upstream system) turns a silent hang into an actionable `DEADLINE_EXCEEDED` error.

---

🤖 _Generated with Fable 5._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/PortableConnectorSpecs.java
```diff
@@ -0,0 +1,161 @@
...
+        ConnectorSpec.newBuilder()
+            .setDisplayName(connector.getDisplayName())
+            .setKind(connector.getKind())
+            .setUri(connector.getUri())
+            .setState(connector.getState())
+            .putAllProperties(connector.getPropertiesMap());
+    if (connector.hasDescription()) spec.setDescription(connector.getDescription());
+    if (connector.hasSource()) spec.setSource(connector.getSource());
+    if (connector.hasPolicy()) spec.setPolicy(connector.getPolicy());
+    if (connector.hasAuth()) spec.setAuth(connector.getAuth());
```
jchauzi-floe, line 46: [major] **`portable_spec` retains `credentials`**

`PortableConnectorSpecs.from(...)` copies the exported connector's `AuthConfig` verbatim. `ConnectorsImpl.hasAuthCredentials(...)` treats an `AuthCredentials` message whose case is `CREDENTIAL_NOT_SET` as credential-free, so a valid create or update can persist an explicitly present `auth.credentials {}`; `ConnectorArchive.write(...)` then reports a successful export.

On read, `ConnectorArchive.validate(...)` rejects every portable spec for which `AuthConfig.hasCredentials()` is true, regardless of credential case. The resulting archive therefore cannot be inspected, dry-run, or imported by the same version of the tool.

Clear `credentials` when copying `AuthConfig` into the portable spec and keep credential material exclusively in `ConnectorTransferEntry.credentials`. Add a round-trip test using `AuthConfig.credentials = AuthCredentials.getDefaultInstance()` so a successful export is guaranteed to remain readable.

```suggestion
if (connector.hasAuth()) spec.setAuth(connector.getAuth().toBuilder().clearCredentials());
```

---

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
```diff
@@ -0,0 +1,532 @@
...
+        List<Connector> connectors = client.listConnectors();
+        connectors = select(connectors, requested);
+        if (connectors.isEmpty()) throw new IllegalArgumentException("no connectors selected");
+
+        var bundle =
+            ConnectorTransferBundle.newBuilder()
+                .setFormatVersion(ConnectorArchive.FORMAT_VERSION)
+                .setExportedAt(now());
+
+        for (Connector listed : connectors) {
```
jchauzi-floe, line 158: [minor] **Export includes `CS_DELETING` connectors, producing an archive the same tool refuses to import**

`export` archives every listed connector, and `PortableConnectorSpecs.from(...)` copies `state` verbatim — including `CS_DELETING` for a connector whose asynchronous teardown is still in flight. Nothing on the write path or in `ConnectorArchive.validate(...)` rejects that state.

On the import side, `validateImport` throws for `CS_DELETING` during the prepare loop, before any per-entry reporting. So a routine "export all" taken while one connector is being deleted yields an archive whose `import` — and even `--dry-run` — always aborts wholesale, and the operator has to re-export with explicit `--connector` selections to exclude a connector they probably never wanted archived. This is the same writer/reader-mismatch class as the bundle-size issue fixed earlier: a successful export should be importable by the same tool version.

Skip `CS_DELETING` connectors at export time with a printed notice (or fail fast when one is explicitly selected via `--connector`), mirroring the import-side check.

---

🤖 _Generated with Fable 5._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
```diff
@@ -0,0 +1,532 @@
...
+        int wouldCreate = 0;
+        int wouldReplace = 0;
+        int wouldSkip = 0;
+        int wouldFail = 0;
+        String importRunNonce = UUID.randomUUID().toString();
+        List<PreparedImport> prepared = new ArrayList<>();
+        for (var entry : bundle.getEntriesList()) {
+          Connector current = existing.get(entry.getPortableSpec().getDisplayName());
+          prepared.add(
+              prepareImport(entry, current, conflictMode, spec -> validateImport(client, spec)));
```
jchauzi-floe, line 245: [minor] **Dry run aborts on the first invalid entry; README promises per-entry "rejected" reporting**

`prepareImport` calls `validateImport`, which throws `IllegalArgumentException` on any validation failure (deleting state, unresolvable destination, `ValidateConnector` returning `ok=false`). In `--dry-run` mode that exception propagates out of the prepare loop, so a single bad entry kills the entire run — no `WOULD ...` lines for the remaining entries and no summary.

The README states: "The dry run reports whether each connector would be created, skipped, replaced, or rejected" — but there is no per-entry rejected outcome anywhere in the code; rejection is only ever a thrown exception. The existing thread on the `ResolveCatalog` `NOT_FOUND` behavior touches this as a side effect, but the gap is general: every validation failure short-circuits the report, which defeats the point of a dry run over a multi-entry archive.

In dry-run mode, catch validation failures per entry, report them as `WOULD REJECT <name>: <reason>`, continue with the remaining entries, and return a nonzero exit status — matching the documented contract. Live imports can keep the fail-fast behavior.

---

🤖 _Generated with Fable 5._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorArchive.java
```diff
@@ -0,0 +1,242 @@
...
+        put(zip, "manifest.json", manifest(bundle).getBytes(StandardCharsets.UTF_8));
+        for (var entry : bundle.getEntriesList()) {
```
jchauzi-floe, line 59: [blocker] **Archive stores every connector — including plaintext credentials — twice, and reads only one copy**

`write()` emits `bundle.pb` (which already contains every entry in full) and then re-emits each entry as `connectors/<id>/connector.pb`, `portable-spec.json`, and `credentials.pb`. `read()` only ever looks at `bundle.pb`; the per-connector tree is write-only. So the archive doubles the plaintext secret material on disk for no consumer, the two copies can silently drift, and the drift risk is what forces `validate()` to cross-check display names between `entry.connector` and `entry.portable_spec` (lines 142-145) and to police credentials leaking into the portable spec (146-149). It also drags in `safeSegment()` and `protobufJson()` purely to name and render files nobody reads.

Write `bundle.pb` and nothing else (keep `manifest.json` only if a human-readable index is genuinely wanted). Render the per-connector view on demand in `inspect` from the parsed bundle — that is already where a human looks. This deletes the loop at 59-70, `safeSegment`, `protobufJson`, and the two cross-copy checks in `validate()`, and halves the secret material at rest.

---

🤖 _Generated with Claude Opus 5._

core/proto/src/main/proto/floecat/connector/connector.proto
```diff
@@ -253,6 +253,34 @@ message ValidateConnectorResponse {
...
+
+message ExportConnectorResponse {
+  Connector connector = 1;
+  optional AuthCredentials credentials = 2;
+}
+
+// Portable on-disk format used by floecat-connector-transfer. connector preserves the exact
+// source record for inspection, while portable_spec replaces account-scoped destination IDs with
+// names and paths suitable for CreateConnector in another account.
+message ConnectorTransferEntry {
```
jchauzi-floe, line 271: [blocker] **Three overlapping representations of one connector; credentials boxed out of the shape the API accepts**

`ConnectorTransferEntry` holds the full source `Connector`, a `portable_spec`, and a third field for credentials — even though the rest of the connector API models credentials as `AuthConfig.credentials` inside the spec. Every consumer then pays for the split: `ExportConnectorResponse` needs its own `optional AuthCredentials` (263-265), `ConnectorTransferCli.specWithCredentials` (357-366) exists only to graft the credentials back into `auth` before calling `CreateConnector`, `ConnectorArchive.validate()` has to assert the two copies of the display name agree, and the import path uses `entry.connector` for nothing but a log line and the idempotency key. Three representations, one fact.

Collapse to one. Let `ExportConnector` return the `Connector` with `auth.credentials` populated when `include_credentials` is set (that is precisely the shape `CreateConnector` consumes), and reduce `ConnectorTransferEntry` to `{ string source_connector_id, ConnectorSpec spec }`. `specWithCredentials`, the separate credentials fields, and the name-agreement validation all disappear; `PortableConnectorSpecs.from` becomes a pure destination rewrite.

---

🤖 _Generated with Claude Opus 5._

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
```diff
@@ -774,6 +849,57 @@ private static boolean hasAuthCredentials(AuthConfig auth) {
...
+    return switch (requestedState) {
+      case CS_UNSPECIFIED, CS_ACTIVE -> ConnectorState.CS_ACTIVE;
+      case CS_PAUSED -> ConnectorState.CS_PAUSED;
+      case CS_DELETING, UNRECOGNIZED ->
+          throw GrpcErrors.invalidArgument(
+              corr, GeneratedErrorMessages.MessageKey.FIELD, Map.of("field", "state"));
+    };
+  }
+
+  private DestinationTarget resolveUpdatedDestination(
```
jchauzi-floe, line 862: [blocker] **resolveUpdatedDestination is a second, divergent copy of createConnector's destination resolution**

`createConnector` (lines 319-400) already resolves a destination from names to ids, and it clears the name once the id is known: `destB.setCatalogId(...); destB.clearCatalogDisplayName()`, same for `clearNamespace()` and `clearTableDisplayName()`. The new `resolveUpdatedDestination` re-implements the same three-step resolution with the same odd `source.hasTable() && !source.getTable().isBlank()` guard, the same throw-for-catalog / ignore-for-namespace / fall-back-for-table policy — but never clears. Two implementations of one rule, with different post-conditions on the persisted record.

The divergence is observable: `UpdateConnector` with `update_mask=destination` and `destination.catalog_display_name="sales"` now stores `{catalog_display_name: "sales", catalog_id: X}`. Rename the catalog to `sales_v2` and any read-modify-write update (GetConnector → resend destination) resolves the stale stored name, gets `Optional.empty()` from `catalogRepo.getByName`, and fails NOT_FOUND — while `catalog_id: X` sits valid in the same message. Worse, `createConnector` explicitly rejects a spec carrying both an id and a non-matching display name (`CONNECTOR_DESTINATION_CATALOG_MISMATCH`, 336-354), so update now persists records that create would refuse as input.

Extract one `ConnectorDestinationResolver.resolve(destination, source, accountId, corr)` with a single documented post-condition — resolved refs become ids and the name is cleared — and call it from both `createConnector` and `updateConnector`. That deletes ~40 lines from `createConnector`, deletes this helper, and makes the stored destination shape a single invariant instead of a per-RPC accident.

---

🤖 _Generated with Claude Opus 5._

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
```diff
@@ -222,6 +224,58 @@ public Uni<GetConnectorResponse> getConnector(GetConnectorRequest request) {
...
+                () -> {
+                  var principalContext = principalProvider.get();
+                  var correlationId = principalContext.getCorrelationId();
+                  var accountId = principalContext.getAccountId();
+
+                  authz.require(principalContext, "connector.export");
+                  var connectorId =
+                      scopedConnectorId(accountId, request.getConnectorId(), correlationId);
+
+                  for (int attempt = 0; attempt < 3; attempt++) {
```
jchauzi-floe, line 242: [major] **Untested read-twice-thrice-then-ABORT loop in exportConnector buys no guarantee**

`exportConnector` wraps its body in a three-attempt loop that reads the connector, resolves the credentials *twice*, re-reads the connector, and retries unless all copies match — then throws ABORTED. This is a hand-rolled pseudo-transaction across two stores that cannot deliver the atomicity it mimics: the second connector read happens after both credential reads, so a write interleaved after `after` is still missed, and the record's `updated_at` is bumped by every `updateConnector` anyway, which makes the double credential resolve redundant. The tool's own README already states the two stores "cannot provide a distributed transactional snapshot", so the loop defends a property the feature does not claim. It costs up to 6 secret-manager reads and 6 KV reads per connector, introduces an ABORTED failure mode callers must handle, and `ConnectorsImplExportConnectorTest` exercises none of it — no retry, no torn read, no abort.

Delete the loop: read the connector, resolve credentials once, return, and keep the README's non-atomicity note. If a consistency fence is actually wanted, use the existing one — read `connectorRepo.metaFor(connectorId).getPointerVersion()`, resolve credentials, re-read the version, and fail if it moved. That is one extra cheap read instead of three extra secret reads, and it says what it means.

---

🤖 _Generated with Claude Opus 5._

service/src/main/java/ai/floedb/floecat/service/connector/impl/ConnectorsImpl.java
```diff
@@ -520,63 +578,79 @@ public Uni<UpdateConnectorResponse> updateConnector(UpdateConnectorRequest reque
...
                   validateReconcilePolicy(desired.getPolicy(), corr);
-                  String accountId = pc.getAccountId();
                   String secretId = connectorId.getId();
                   boolean authTouched =
                       maskTargets(normalizedMask, "auth")
                           || maskTargets(normalizedMask, "auth.credentials");
                   boolean incomingHasCredentials = hasAuthCredentials(desired.getAuth());
-                  Optional<AuthCredentials> priorCredentials = Optional.empty();
+                  boolean shouldDeleteSecret = authTouched && !incomingHasCredentials;
+                  Optional<AuthCredentials> priorCredentials =
```
jchauzi-floe, line 608: [major] **Credential rollback duplicated at five exit points; restoreCredentials cannot restore "absent"**

`if (authTouched) restoreCredentials(accountId, secretId, priorCredentials); throw ...` now appears at five separate exits of `updateConnector` (the nested try inside the no-op branch at 624-645, and the `NameConflictException`, `PreconditionFailedException`, and catch-all handlers at 656-684). Three of those handlers previously did nothing but map an error to a gRPC status; they now also carry compensation, and the no-op branch grew an entire nested try/catch to host a copy. The compensation is a cross-cutting concern smeared across the error mapping.

Hoist it once: resolve `priorCredentials`, then wrap everything from the first credential write to the successful return in a single `try { ... } catch (RuntimeException failure) { if (authTouched) restoreCredentials(...); throw failure; }`. The specific catches shrink back to pure error mapping, the nested try disappears, and the invariant "any throw after the secret write restores the secret" is stated in one place instead of being re-proved five times. Separately, `restoreCredentials` (903-907) only acts when `priorCredentials.isPresent()`: when there was no prior secret and the record update fails, the freshly written secret is left orphaned. Model the compensation as restoring the prior *state* — present → put, absent → delete — rather than "put if present".

---

🤖 _Generated with Claude Opus 5._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
```diff
@@ -0,0 +1,532 @@
...
+
+    @CommandLine.Option(
+        names = "--connector",
+        description = "Connector display name or ID to include; repeatable (default: all)")
+    List<String> requested = new ArrayList<>();
+
+    @CommandLine.Option(names = "--force", description = "Replace an existing output archive")
+    boolean force;
+
+    @CommandLine.Option(
```
jchauzi-floe, line 137: [major] **--plaintext-secrets is unreachable ceremony and include_credentials is a dead RPC knob**

`--plaintext-secrets` is declared `required = true` with arity 0, so picocli refuses to run the command unless the flag is present, and a present flag is `true` — which makes the guard at 145-147 unreachable and the "acknowledgement" a mandatory incantation rather than a choice. Meanwhile the export call hard-codes `setIncludeCredentials(true)` (line 163), so the new `include_credentials` field on `ExportConnectorRequest` has exactly one caller that always passes the same value: unnecessary optionality on a brand-new API surface, and there is no way to take a credential-free export at all.

Make the two mean something together: drop `required = true`, pass `setIncludeCredentials(plaintextSecrets)`, and keep a single runtime check only if credentials are requested. The flag becomes a real decision (secret-bearing export vs. definition-only export), the RPC field gains a second caller that justifies it, and the dead branch goes away.

---

🤖 _Generated with Claude Opus 5._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
```diff
@@ -0,0 +1,532 @@
...
+      replacement = replacement.toBuilder().setState(ConnectorState.CS_ACTIVE).build();
+    }
+    return UpdateConnectorRequest.newBuilder()
+        .setConnectorId(current.getResourceId())
+        .setSpec(replacement)
+        .setUpdateMask(REPLACE_UPDATE_MASK)
+        .build();
+  }
+
+  static String idempotencyKey(
```
jchauzi-floe, line 448: [major] **Idempotency key is a hashed random nonce whose only effect is opting into an undocumented server branch**

`idempotencyKey` SHA-256s a per-run `UUID.randomUUID()` together with the account, source id, and display name. Because the nonce is fresh every run (and `ConnectorTransferCliTest.idempotencyKeyIsStableWithinRunAndUniqueAcrossRuns` locks that in), the key never hits `IdempotencyGuard`'s stored record — each entry is created once per run anyway, so the guard is never consulted. Its real effect is indirect and unwritten: passing *any* non-empty key steers `createConnector` away from its `getByName` pre-check into the name-conflict path at 486-521, where a matching `canonicalFingerprint` is returned as an idempotent success. That is what makes re-imports "safely repeatable", and nothing in the CLI, the README, or a comment says so. The SHA-256 machinery is theatre around a coin flip on a server branch nobody named.

Either derive the key from stable bundle identity (`source_account_id` + `exported_at` + source connector id) so re-running the same archive genuinely short-circuits in `IdempotencyGuard` and a *changed* archive correctly reports a fingerprint conflict — and say so in a one-line comment — or drop the key entirely and let the server's fingerprint comparison be reached by an explicit, documented server-side rule instead of a client-side nonce trick.

---

🤖 _Generated with Claude Opus 5._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/GrpcHeaders.java
```diff
@@ -0,0 +1,65 @@
...
+package ai.floedb.floecat.tools.transfer;
+
+import io.grpc.CallOptions;
+import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
+import io.grpc.ForwardingClientCall;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+
+final class GrpcHeaders implements ClientInterceptor {
```
jchauzi-floe, line 26: [major] **Fork of client-cli's AuthHeaderInterceptor rather than reuse**

`GrpcHeaders` is `client-cli/src/main/java/ai/floedb/floecat/client/cli/util/AuthHeaderInterceptor.java` with the header names inlined — same `ForwardingClientCall` shape, same three headers, same `regionMatches(true, 0, "bearer ", 0, 7)` bearer-prefix trick. `ConnectorTransferCli.env`/`envInt` (469-477) likewise re-derive the `FLOECAT_TOKEN` / `FLOECAT_SESSION_TOKEN` / `FLOECAT_ACCOUNT` / `FLOECAT_GRPC_HOST` / `FLOECAT_GRPC_PORT` defaults that `Shell.initAuthConfig` and `Shell.resolveGrpcEndpointOverride` already own. This is now the third place in the tree that builds an authenticated Floecat channel (client-cli, the Trino client, this), each with its own defaults — so the auth header contract can drift per tool.

Move `AuthHeaderInterceptor` and the env-default resolution into a dependency-light shared module (client-cli's Quarkus deps rule out depending on it directly) and have both client-cli and this tool consume it. If that is judged out of scope for this PR, at minimum delete `GrpcHeaders` in favour of `AuthHeaderInterceptor` with header names supplied as constants, so there is one implementation of the header contract.

---

🤖 _Generated with Claude Opus 5._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
```diff
@@ -0,0 +1,532 @@
...
+    @Override
+    public Integer call() throws Exception {
+      ConnectorTransferBundle bundle = ConnectorArchive.read(input);
+      try (var client = parent.client()) {
+        Map<String, Connector> existing = new HashMap<>();
+        for (Connector connector : client.listConnectors()) {
+          existing.put(connector.getDisplayName(), connector);
+        }
+
+        int imported = 0;
```
jchauzi-floe, line 234: [major] **ImportCommand: six parallel counters, a double dispatch over the same enum, and a shadow enum**

`call()` is ~100 lines carrying `imported`, `skipped`, `wouldCreate`, `wouldReplace`, `wouldSkip`, `wouldFail` — six ints that are just the sizes of four groups — then switches over `item.action()` once to count and print, and again 20 lines later to execute. `ImportAction` is `ConflictMode` plus `CREATE`, and `importAction()` (427-434) is a three-arm identity switch between them, i.e. a whole enum and a mapper earning nothing. `wouldFail > 0` then re-scans `prepared` to recover the display name it already printed.

Group once and derive the counts: `Map<ImportAction, List<PreparedImport>> byAction = prepared.stream().collect(groupingBy(PreparedImport::action))`. The dry-run report becomes a loop over the map (sizes instead of counters, and the FAIL entry is right there), and the apply pass becomes two loops over `byAction.get(SKIP)` / the rest. Split the method along its real seams — `plan(bundle, existing, mode)`, `report(plan)`, `apply(plan)` — and collapse the two enums: keep one four-valued `ImportAction` and give `--conflict` explicit `completionCandidates`, or drop `ImportAction` and let the plan carry `Connector current` (nullable) plus the mode. Either way the identity switch goes.

---

🤖 _Generated with Claude Opus 5._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorArchive.java
```diff
@@ -0,0 +1,242 @@
...
+  private static String protobufJson(com.google.protobuf.MessageOrBuilder message)
+      throws IOException {
+    try {
+      return JsonFormat.printer().includingDefaultValueFields().print(message) + "\n";
+    } catch (InvalidProtocolBufferException e) {
+      throw new IOException("failed to render connector protobuf as JSON", e);
+    }
+  }
+
+  private static String manifest(ConnectorTransferBundle bundle) {
```
jchauzi-floe, line 182: [minor] **Hand-rolled JSON writer and escaper in a file that already depends on JsonFormat**

`manifest()` concatenates JSON by hand — literal braces, manual comma placement, a trailing-newline fixup at 203 — and `jsonEscape()` (212-230) hand-implements string escaping including `\u%04x` control-character encoding. Forty lines of bug-prone serialisation sitting six lines below `protobufJson()`, which uses `JsonFormat` from a dependency this module already declares.

Define the manifest as a small proto (or reuse `ConnectorTransferBundle` with `entries[].credentials` cleared) and print it with `JsonFormat.printer()`. `manifest`, `jsonEscape`, and the escaping test surface all go away, and the manifest gains guaranteed-valid JSON for free.

---

🤖 _Generated with Claude Opus 5._

tools/connector-transfer/src/main/java/ai/floedb/floecat/tools/transfer/ConnectorTransferCli.java
```diff
@@ -0,0 +1,532 @@
...
+    ImportAction action = importAction(current != null, conflictMode);
+    String validationSummary = "";
+    if (action == ImportAction.CREATE || action == ImportAction.REPLACE) {
+      validationSummary = validator.validate(spec);
+    }
+    return new PreparedImport(entry, spec, current, action, validationSummary);
+  }
+
+  private static String validateImport(Client client, ConnectorSpec spec) {
+    if (spec.getState() == ConnectorState.CS_DELETING) {
```
jchauzi-floe, line 409: [minor] **Connector-state validation scattered across three sites for one rule**

The rule "an importable connector is ACTIVE or PAUSED" is enforced in three places: `validateImport` rejects `CS_DELETING` (409-412), `replacementRequest` silently coerces `CS_UNSPECIFIED` to `CS_ACTIVE` (438-440), and the server rejects `CS_UNSPECIFIED` on update (1403-1407) and `CS_DELETING` on create (`initialConnectorState`, 852-860). `ConnectorArchive.validate()` — the gate that already rejects blank names, duplicate ids, and embedded credentials — checks state not at all, so a malformed archive is caught late and inconsistently, and the client-side coercion quietly rewrites data the archive claimed to preserve.

Check it once in `ConnectorArchive.validate()`: reject any entry whose state is not `CS_ACTIVE` or `CS_PAUSED`. Then delete the `CS_DELETING` check in `validateImport` and the coercion in `replacementRequest`, which becomes a two-line request builder.

---

🤖 _Generated with Claude Opus 5._

service/src/test/java/ai/floedb/floecat/service/connector/impl/ConnectorsImplExportConnectorTest.java
```diff
@@ -0,0 +1,103 @@
...
+                    .build())
+            .await()
+            .indefinitely();
+
+    assertThat(response.getConnector()).isEqualTo(connector);
+    assertThat(response.getCredentials()).isEqualTo(credentials);
+    verify(service.authz).require(principal, "connector.export");
+  }
+
+  private static void installBasePrincipal(
```
jchauzi-floe, line 92: [minor] **Third and fourth copies of the installBasePrincipal reflection hack; create tests filed under update**

`installBasePrincipal` — reflectively setting `BaseServiceImpl.principal` — is copied verbatim into both new test files, joining the copies already in `ConnectorsImplDeleteConnectorTest` and `StorageAuthorityServiceImplTest`. Five identical reflection helpers now encode the same test-boundary workaround, so making `principal` injectable (or adding a package-private setter / shared `TestSupport` helper) requires touching five files. Separately, `ConnectorsImplUpdateConnectorTest` opens with `createStatePreservesPausedAndDefaultsUnspecifiedToActive` and `createStateRejectsDeleting`, which test `initialConnectorState` on the create path and belong in `ConnectorsImplCreateConnectorTest`.

Extract the helper once into the existing `service/src/test/.../util/TestSupport` and have all call sites use it (ideally alongside a package-private seam on `BaseServiceImpl` so the reflection disappears entirely), and move the two create-state tests to the create test class.

---

🤖 _Generated with Claude Opus 5._

PR #414, opened 2026-07-27
GetUserObjects - 1 : prepare deterministic parallel pin planning

mrouvroy-floe:
## Summary

Separate per-input pin computation from the ordered merge, preserving deterministic first-touch pin semantics. 
Add thread-safe diagnostics aggregation for work that will run off the request thread.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist


service/src/main/java/ai/floedb/floecat/service/query/resolver/QueryInputResolver.java
```diff
@@ -222,38 +222,24 @@ public ResolutionResult resolveInputs(
...
-          addResolvedAndPins(state, in.getTableId(), override);
-        }
-
-        case VIEW_ID -> {
-          diag.count("pin.view_id_inputs");
-          addResolvedAndPins(state, in.getViewId(), override);
-        }
-
-        default -> throw GrpcErrors.invalidArgument(correlationId, QUERY_INPUT_INVALID, Map.of());
+      plans.add(planInput(state, in, resolvedNames));
```
jchauzi-floe: [major] **`InputPlan` pins lose construction-time GC protection**

`planInput` now constructs every `TablePin` into `plans`, but `registerResolvingPinBlobs` is not called until the later `mergePin` pass. This violates the `QueryContextStore` contract that a pin becomes a GC root at construction time, which exists specifically to close the resolve-to-persist window.

While a later input or view expansion performs metadata I/O—and especially once these plans run in parallel—a concurrent table update plus blob GC can supersede and delete an earlier pin's root before line 242 registers it. `resolveInputs` can then return a pin whose referenced blobs are already gone; registering it before return does not repair that race.

Register each candidate pin's transient roots immediately after `tablePinFor` returns, track ownership for the resolution attempt, and release registrations for discarded compatible candidates or failed planning. Keep only the merge/deduplication ordered, and add a test that blocks a later plan while verifying the completed pin is already registered.

---

service/src/main/java/ai/floedb/floecat/service/query/resolver/QueryInputResolver.java
```diff
@@ -222,38 +222,24 @@ public ResolutionResult resolveInputs(
...
+    }
+
+    // Merge the plans into the pin set in input order. Order decides two things: `resolved` mirrors
+    // the request order, and mergePin keeps the first pin seen for a table (rooting its blobs)
+    // while
+    // raising QUERY_TABLE_PIN_CONFLICT when a later reference pins the same table with an
+    // incompatible snapshot/as-of. Every kept pin is GC-rooted here, before resolveInputs returns
+    // and therefore before the caller reads any schema/stats or persists the context.
+    for (InputPlan plan : plans) {
```
jchauzi-floe: [minor] **`QUERY_TABLE_PIN_CONFLICT` can be masked**

All `planInput` calls now finish before the first `mergePin`. For an input sequence such as `[table@888, table@999, empty-target]`, the previous loop merged the second input and raised `QUERY_TABLE_PIN_CONFLICT` before inspecting the third; the new loop plans the third input first and raises `QUERY_INPUT_INVALID`, so the earlier conflict is never evaluated.

This changes the returned gRPC status/error code based on an unrelated later input, despite the refactor's no-behavior-change goal and its claim that ordered conflict detection is preserved. The same issue remains when planning is later fanned out if all plans are gathered before merging.

Join/gather plans in input order and merge each plan before observing a later plan's failure, or define and enforce an explicit validation precedence before pin work. Add a regression test where an earlier conflicting pair is followed by a malformed or unresolved input.

---

service/src/main/java/ai/floedb/floecat/service/query/resolver/QueryInputResolver.java
```diff
@@ -210,82 +265,120 @@ public ResolutionResult resolveInputs(
...
+        };
 
+    List<TablePin> pins = new ArrayList<>();
     if (rid.getKind() == ResourceKind.RK_VIEW) {
       // Views are not pinned directly. We only pin their base tables.
       // Reject snapshot_id overrides for views; allow AS-OF and apply it to dependency pins.
       validateViewOverride(state.correlationId, rid, override);
-      collectBaseTables(state, rid, effectiveAsOf(override, state.asOfDefault), new HashSet<>());
-      return;
+      collectBaseTablePins(
```
jchauzi-floe: [minor] **`InputPlan` defers view conflicts**

`planInput` expands every base table of a view before `mergePlan` reconciles any of the resulting pins. A failure from a later base can therefore mask a conflict established by an earlier base against a pin from a preceding input.

For example, consider an explicit `table@888` input followed by an `AS_OF` view whose first base resolves that table to snapshot `999` and whose second base has no snapshot at that time. The base implementation reconciled the first base immediately and returned `QUERY_TABLE_PIN_CONFLICT`; this version continues planning and instead returns `QUERY_SNAPSHOT_NOT_FOUND_AT_TIME` from the second base. That changes deterministic error precedence despite the no-behavior-change goal and is the same class of masking that the top-level plan/merge loop now avoids.

Preserve the successful pin prefix and its terminal planning failure in `InputPlan`, then merge that prefix before rethrowing the failure, or otherwise reconcile pins incrementally in dependency order. Add a regression test covering a conflicting first view base followed by a failing later base.

---

jchauzi-floe (approved):

PR #415, opened 2026-07-27
GetUserObjects - 2: resolve pins concurrently and safely

mrouvroy-floe:
## Summary

Resolve independent input pins concurrently through a bounded, cancellation-aware fan-out. 
Preserve request context, transient pin roots, single-flight current-snapshot lookup, 
failure diagnostics, and prompt cancellation cleanup.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist

- [x]  Documentation updated where needed

### DIFF
--- a/core/scanner/src/main/java/ai/floedb/floecat/scanner/spi/CatalogOverlay.java
+++ b/core/scanner/src/main/java/ai/floedb/floecat/scanner/spi/CatalogOverlay.java
@@ -198,7 +198,8 @@ public interface CatalogOverlay {
    *
    * <p>The default preserves compatibility for existing overlays, whose lifecycle state may be tied
    * to one request thread. Implementations backed by thread-safe services may opt in to concurrent
-   * resolution.
+   * resolution. Opting in permits {@link #catalog}, {@link #resolve}, {@code resolveName(s)}, and
+   * {@link #tablePinFor} callbacks to execute concurrently and off the caller thread.
    */
   default boolean supportsConcurrentResolution() {
     return false;
--- a/core/scanner/src/main/java/ai/floedb/floecat/scanner/spi/CatalogOverlay.java
+++ b/core/scanner/src/main/java/ai/floedb/floecat/scanner/spi/CatalogOverlay.java
@@ -199,7 +199,11 @@ public interface CatalogOverlay {
    * <p>The default preserves compatibility for existing overlays, whose lifecycle state may be tied
    * to one request thread. Implementations backed by thread-safe services may opt in to concurrent
    * resolution. Opting in permits {@link #catalog}, {@link #resolve}, {@code resolveName(s)}, and
-   * {@link #tablePinFor} callbacks to execute concurrently and off the caller thread.
+   * {@link #tablePinFor} callbacks to execute concurrently and off the caller thread. It also
+   * permits GetUserObjects relation construction to run off-thread, including schema/name reads,
+   * pin validation, stats access, and the engine metadata decorator lifecycle. Implementations
+   * opting in must therefore make all of those callbacks thread-safe and must not depend on custom
+   * caller-thread state that the service context propagation does not capture.
    */
   default boolean supportsConcurrentResolution() {
     return false;
--- a/core/scanner/src/main/java/ai/floedb/floecat/scanner/spi/CatalogOverlay.java
+++ b/core/scanner/src/main/java/ai/floedb/floecat/scanner/spi/CatalogOverlay.java
@@ -199,11 +199,13 @@ public interface CatalogOverlay {
    * <p>The default preserves compatibility for existing overlays, whose lifecycle state may be tied
    * to one request thread. Implementations backed by thread-safe services may opt in to concurrent
    * resolution. Opting in permits {@link #catalog}, {@link #resolve}, {@code resolveName(s)}, and
-   * {@link #tablePinFor} callbacks to execute concurrently and off the caller thread. It also
-   * permits GetUserObjects relation construction to run off-thread, including schema/name reads,
-   * pin validation, stats access, and the engine metadata decorator lifecycle. Implementations
-   * opting in must therefore make all of those callbacks thread-safe and must not depend on custom
-   * caller-thread state that the service context propagation does not capture.
+   * {@link #tablePinFor} callbacks to execute concurrently and off the caller thread, together with
+   * overlay-owned schema/name callbacks used while assembling GetUserObjects relations ({@link
+   * #schemaFor}, {@link #tableSchema}, {@link #tableName}, and {@link #viewName}). It does not
+   * change the caller-thread contract of separately injected stats, pin-validation, or
+   * engine-decoration collaborators. Implementations opting in must make the listed overlay
+   * callbacks thread-safe and must not depend on custom caller-thread state that service context
+   * propagation does not capture.
    */
   default boolean supportsConcurrentResolution() {
     return false;
--- a/core/scanner/src/main/java/ai/floedb/floecat/scanner/spi/CatalogOverlay.java
+++ b/core/scanner/src/main/java/ai/floedb/floecat/scanner/spi/CatalogOverlay.java
@@ -194,6 +194,17 @@ public interface CatalogOverlay {
   }
 
   /**
+   * Whether independent resolution callbacks may run concurrently on this overlay instance.
+   *
+   * <p>The default preserves compatibility for existing overlays, whose lifecycle state may be tied
+   * to one request thread. Implementations backed by thread-safe services may opt in to concurrent
+   * resolution.
+   */
+  default boolean supportsConcurrentResolution() {
+    return false;
+  }
+
+  /**
    * Resolves a relation (table or view) by name reference using an explicit engine context.
    *
    * <p>Prefer this overload wherever the caller already holds the request's engine context — see
--- a/service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
+++ b/service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
@@ -99,7 +99,9 @@ public final class BoundedFanout {
    * As {@link #mapOrdered(List, int, Executor, Function)}, but {@code cancelled} is polled so a
    * cancelled stream interrupts every submitted task and returns without waiting for a slow task's
    * completion. Tasks must cooperate with interruption while blocked in downstream calls. The
-   * permit wait is interruptible for the same reason.
+   * permit wait is interruptible for the same reason. A non-cancellation task failure waits for
+   * every already-submitted sibling before surfacing, so its latency is bounded by the slowest
+   * sibling; this completion guarantee lets callers safely release task-owned resources.
    */
   public static <I, O> List<O> mapOrdered(
       List<I> items,
@@ -164,7 +166,9 @@ public final class BoundedFanout {
 
   /**
    * Cancellation-aware ordered result consumption. A consumer failure has the same ordered
-   * precedence as a task failure, while submitted work is cancelled promptly when requested.
+   * precedence as a task failure, while submitted work is cancelled promptly when requested. A
+   * non-cancellation failure waits for already-submitted siblings before it surfaces, preserving
+   * the task-resource completion guarantee at the cost of slowest-sibling failure latency.
    */
   public static <I, O> void forEachOrdered(
       List<I> items,
--- a/service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
+++ b/service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
@@ -418,6 +418,7 @@ public final class BoundedFanout {
     return new CancellationException("fan-out cancelled");
   }
 
+  /** One submitted input and the future through which its completion is observed. */
   private static final class CompletionSlot<O> {
     private final int index;
     private CompletableFuture<O> completion;
@@ -428,6 +429,7 @@ public final class BoundedFanout {
     }
   }
 
+  /** The successful result or unwrapped failure captured for one input. */
   private record TaskOutcome<O>(O result, Throwable failure) {
     private static <O> TaskOutcome<O> success(O result) {
       return new TaskOutcome<>(result, null);
--- a/service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
+++ b/service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
@@ -159,11 +159,11 @@ public final class BoundedFanout {
         active.remove(slot);
         TaskOutcome<O> outcome = completedOutcome(slot.completion);
         outcomes.set(slot.index, outcome);
-        onCompletion.accept(slot.index, outcome);
         if (next < items.size()) {
           active.add(
               submitCompletionTask(next, items.get(next++), executor, context, task, completions));
         }
+        onCompletion.accept(slot.index, outcome);
       }
     } catch (RuntimeException | Error submissionFailure) {
       awaitCompletedTasks(active);
@@ -196,7 +196,6 @@ public final class BoundedFanout {
         CompletionSlot<O> slot = takeCancellableCompletion(completions, active, cancelled);
         TaskOutcome<O> outcome = completedOutcome(slot.task, active, cancelled);
         outcomes.set(slot.index, outcome);
-        onCompletion.accept(slot.index, outcome);
         active.remove(slot);
         if (next < items.size()) {
           checkCancelled(cancelled, active);
@@ -204,6 +203,7 @@ public final class BoundedFanout {
               submitCancellableTask(
                   next, items.get(next++), executor, context, task, cancelled, completions));
         }
+        onCompletion.accept(slot.index, outcome);
       }
     } catch (CancellationException cancellationFailure) {
       cancelSubmittedTasks(active);
--- a/service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
+++ b/service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2026 Yellowbrick Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ai.floedb.floecat.service.concurrent;
+
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.Scope;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Semaphore;
+import java.util.function.Function;
+
+/** Runs independent, mostly-blocking tasks on a shared executor with a concurrency bound. */
+public final class BoundedFanout {
+
+  private BoundedFanout() {}
+
+  /**
+   * Apply {@code task} to each item on {@code executor}, at most {@code permits} running at once,
+   * and return the results in input order. Each task runs under the caller's OpenTelemetry context.
+   * A task failure surfaces unwrapped — its original {@link RuntimeException} or {@link Error},
+   * never a {@link CompletionException} wrapper — and the first such failure propagates to the
+   * caller once its future is joined.
+   */
+  public static <I, O> List<O> mapOrdered(
+      List<I> items, int permits, Executor executor, Function<I, O> task) {
+    Semaphore gate = new Semaphore(permits);
+    Context otelContext = Context.current();
+    List<CompletableFuture<O>> futures =
+        items.stream()
+            .map(
+                item ->
+                    CompletableFuture.supplyAsync(
+                        () -> {
+                          gate.acquireUninterruptibly();
+                          try (Scope ignored = otelContext.makeCurrent()) {
+                            return task.apply(item);
+                          } finally {
+                            gate.release();
+                          }
+                        },
+                        executor))
+            .toList();
+
+    List<O> results = new ArrayList<>(items.size());
+    for (CompletableFuture<O> future : futures) {
+      results.add(join(future));
+    }
+    return results;
+  }
+
+  private static <O> O join(CompletableFuture<O> future) {
+    try {
+      return future.join();
+    } catch (CompletionException ce) {
+      Throwable cause = ce.getCause();
+      if (cause instanceof RuntimeException re) {
+        throw re;
+      }
+      if (cause instanceof Error e) {
+        throw e;
+      }
+      throw new IllegalStateException("unexpected checked exception from parallel task", cause);
+    }
+  }
+}
--- a/service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
+++ b/service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
@@ -65,51 +65,13 @@ public final class BoundedFanout {
   }
 
   /**
-   * As {@link #mapOrdered(List, int, Executor, Function)}, but consumes each successful result on
-   * the caller thread before observing the next result. This preserves input-order validation
-   * precedence when a result consumer can itself fail, while still awaiting all submitted tasks
-   * before surfacing that failure so callers can safely release task-owned resources.
-   */
-  public static <I, O> void forEachOrdered(
-      List<I> items,
-      int permits,
-      Executor executor,
-      Function<I, O> task,
-      Consumer<? super O> consumer) {
-    List<TaskOutcome<O>> outcomes = emptyOutcomes(items.size());
-    OrderedOutcomeConsumer<O> ordered = new OrderedOutcomeConsumer<>(outcomes, consumer);
-    completeAll(items, permits, executor, task, outcomes, ordered::accept);
-  }
-
-  /**
-   * As {@link #mapOrdered(List, int, Executor, Function)}, but {@code cancelled} is polled so a
-   * cancelled stream interrupts every submitted task and returns without waiting for a slow task's
-   * completion. Tasks must cooperate with interruption while blocked in downstream calls. The
-   * submission window observes cancellation before it submits more work. A non-cancellation task
-   * failure waits for every input task before surfacing, so callers can safely release task-owned
-   * resources. {@code cancelled} may be read concurrently by the scheduler and workers; it must be
-   * non-blocking and thread-safe (typically {@link java.util.concurrent.atomic.AtomicBoolean#get}).
-   * Observed cancellation throws {@link CancellationException}.
-   */
-  public static <I, O> List<O> mapOrdered(
-      List<I> items,
-      int permits,
-      ExecutorService executor,
-      Function<I, O> task,
-      BooleanSupplier cancelled) {
-    List<TaskOutcome<O>> outcomes = emptyOutcomes(items.size());
-    completeAllCancellable(
-        items, permits, executor, task, cancelled, outcomes, (index, outcome) -> {});
-    return orderedResults(outcomes);
-  }
-
-  /**
    * Cancellation-aware ordered result consumption. A consumer failure has the same ordered
    * precedence as a task failure, while submitted work is cancelled promptly when requested. A
    * non-cancellation failure waits for already-submitted siblings before it surfaces, preserving
    * the task-resource completion guarantee at the cost of slowest-sibling failure latency. {@code
-   * cancelled} has the same concurrent-read contract and {@link CancellationException} behavior as
-   * the cancellable {@code mapOrdered} overload.
+   * cancelled} may be read concurrently by the scheduler and workers, so it must be non-blocking
+   * and thread-safe (typically {@link java.util.concurrent.atomic.AtomicBoolean#get}). Observed
+   * cancellation throws {@link CancellationException}.
    */
   public static <I, O> void forEachOrdered(
       List<I> items,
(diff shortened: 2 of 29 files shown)


service/src/main/java/ai/floedb/floecat/service/query/resolver/QueryInputResolver.java
```diff
@@ -263,6 +395,52 @@ public ResolutionResult resolveInputs(
...
+   * conflict detection).
+   */
+  private List<InputPlan> planInputsConcurrently(
+      ResolutionState state,
+      List<QueryInput> inputs,
+      Map<NameRef, Optional<ResourceId>> resolvedNames) {
+    AggregatingPhaseDiagnostics taskDiagnostics = new AggregatingPhaseDiagnostics();
+    ResolutionState taskState = state.withDiagnostics(taskDiagnostics);
+    try {
+      return BoundedFanout.mapOrdered(
```
jchauzi-floe: [major] **Propagate `GetUserObjects` cancellation**

`planInputsConcurrently` calls the four-argument, non-cancellable `BoundedFanout.mapOrdered`; the cancellation-aware overload is referenced only by its unit test. The `GetUserObjects` termination hook only publishes telemetry, so no cancellation signal reaches this resolver.

If a client cancels while a metadata-store lookup is blocked, the resolver still waits for every submitted input and continues its transient-root work before returning. This defeats the PR's promised prompt cancellation and can keep database calls and request resources alive after the subscriber has gone away.

Thread a cancellation flag from `UserObjectBundleIterator` through `collectChunkPins` and `resolveInputs`, use an `ExecutorService` with the cancellation-aware overload here, and check the same flag around the serial and pre-fan-out resolution stages.

---

service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
```diff
@@ -0,0 +1,270 @@
...
+   */
+  public static <I, O> List<O> mapOrdered(
+      List<I> items, int permits, Executor executor, Function<I, O> task) {
+    validatePermits(permits);
+    Semaphore gate = new Semaphore(permits);
+    PropagatedContext context = PropagatedContext.capture();
+    List<CompletableFuture<O>> futures = new ArrayList<>(items.size());
+    try {
+      for (I item : items) {
+        futures.add(
```
jchauzi-floe: [major] **Bound submitted work, not only task bodies**

The semaphore is acquired inside `runTask`, after this loop has submitted a future for every item. With `QueryInputResolver`'s virtual-thread-per-task executor, each input therefore creates and starts a virtual thread; items beyond `permits` merely block those threads on the gate.

`DescribeInputs` has no item-count validation, and the server accepts inbound messages up to 64 MiB. A large request can consequently allocate O(N) futures and virtual threads before producing a result, exhausting heap or scheduler resources even though the configured concurrency is small. The same helper also eagerly queues every namespace scan when used by `NameResolver`.

Implement a rolling window or submit only `min(permits, items.size())` worker loops that claim input indices and write into ordered result slots. That keeps both active and queued work bounded while preserving result order.

---

service/src/main/java/ai/floedb/floecat/service/context/PropagatedContext.java
```diff
@@ -0,0 +1,87 @@
...
+      return ResolvedCallContexts.callWith(
+          call,
+          () -> {
+            // MDC is a projection of the call context; re-derive it so off-thread log lines carry
+            // the request's ids, then clear so a pooled worker does not leak them to the next task.
+            InboundContextInterceptor.populateMdc(call);
+            try {
+              return body.get();
+            } finally {
+              InboundContextInterceptor.clearMdc();
```
jchauzi-floe: [minor] **Restore the prior MDC**

`supply` unconditionally removes all Floecat MDC keys on exit, including `floecat_component` and `floecat_operation`, which it never populated. Unlike the OpenTelemetry `Scope` and `ResolvedCallContexts.callWith`, it does not restore the ambient state that existed before entering the propagated scope.

When the supplied executor runs inline or already propagates MDC—as the `ManagedExecutor` used by `NameResolver` may—the task enters with existing values. Clearing them corrupts the caller or executor context for nested scopes and for continuations that run after the supplier completes on the same runnable.

Snapshot the relevant MDC values (or the whole MDC map) before `populateMdc`, then restore them in `finally`, removing only keys that were absent before entry. Add a same-thread or nested-scope test that verifies the original values survive.

---

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -378,8 +392,10 @@ private RelationPinSet collectChunkPins(
             asOfDefault,
             Optional.of(ctx.getQueryDefaultCatalogId()),
             currentSnapshotPinCache,
-            diagnostics);
+            diagnostics,
+            cancelled);
```
jchauzi-floe, line 554: [major] **`resolveInputs` overload bypasses test resolvers**

Passing `cancelled` selects the new eight-argument `resolveInputs` overload, but the PR only updated `TestQueryInputResolver` and the anonymous `emptyResolver` to override the seven-argument overload. Their overrides are therefore bypassed, and the base implementation runs with the `null` `metadataGraph` supplied by those fakes.

This is the cause of the current red `ci-core` check: 39 `UserObjectBundleServiceTest` cases fail at `withMetadataGraph` with `Cannot invoke CatalogOverlay.supportsConcurrentResolution() because this.metadataGraph is null`. The same mismatch remains on the stacked #416 branch.

Update every resolver fake to override the cancellable signature (including `BooleanSupplier`), or structure the overloads so the cancellable entry point preserves virtual dispatch through the existing override contract. Add a test that exercises `UserObjectBundleService` through the fake after this call-site change.

---

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -378,8 +392,10 @@ private RelationPinSet collectChunkPins(
...
     diagnostics.nanos("pin.resolver", System.nanoTime() - resolverStartNs);
+    throwIfCancelled(cancelled);
```
jchauzi-floe: [major] **Cancellation abandons transient pin roots**

`resolveInputs` may return successfully after registering every resolved pin through `registerResolvingPinBlobs`, and cancellation can become visible at this check. Throwing here drops `resolution` before its pins are committed or released. The second cancellation check in `fillPending` has the same ownership gap after `collectChunkPins` returns and before `accumulateChunkPins` takes ownership.

Those registrations then remain GC roots until the resolving-pin grace expires (10 minutes by default), so cancelled streams can retain obsolete table/snapshot blobs and grow the per-query resolving entry despite the PR's prompt-cleanup guarantee.

Keep explicit ownership of the returned `RelationPinSet` through the cancellation checks: if cancellation wins before `accumulateChunkPins`, call `releaseResolvingPinBlobs` for exactly those roots before rethrowing. Refactor the existing merge-failure release at `accumulateChunkPins` as needed to avoid a double release, and add a test that cancels immediately after the resolver returns.

---

service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
```diff
@@ -0,0 +1,440 @@
...
+          firstErrorFailure = e;
+        }
+      }
+      if (firstRuntimeFailure == null && firstErrorFailure == null && nextItem < items.size()) {
+        if (cancelled.getAsBoolean()) {
+          cancelSubmittedTasks(futures);
+          throw cancelled();
+        }
+        I item = items.get(nextItem++);
+        futures.add(executor.submit(() -> runTask(gate, context, task, item, cancelled)));
```
jchauzi-floe: [minor] **Replacement rejection skips task cleanup**

Only the initial submission loop is protected by the submission-failure handlers. If `executor.submit` rejects a replacement task here after the rolling window has begun advancing, the exception escapes immediately while the other futures in the window may still be running. The same unguarded replacement submission exists in the other three collection loops.

This violates `BoundedFanout`'s cleanup contract and lets metadata work continue after the caller has already unwound its state; it is particularly relevant for the shared `ManagedExecutor` used by `NameResolver`, which can reject work under saturation or shutdown.

Handle replacement-submission failures with the same policy as initial failures: await already-submitted tasks for non-cancellable variants and cancel them for cancellation-aware variants. Add a regression test whose executor accepts the initial window and rejects the first replacement.

---

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -106,7 +119,103 @@
 public class UserObjectBundleService {
 
   private static final int MAX_RESOLUTIONS_PER_CHUNK = 25;
+  private static final int MAX_CONCURRENT_METADATA_LOOKUPS = 64;
```
jchauzi-floe: [major] **Two independent 64-permit pools both claim to be the global metadata I/O bound**

`UserObjectBundleService` creates its own `metadataLookupPermits` (`MAX_CONCURRENT_METADATA_LOOKUPS = 64`) and executor, while `QueryInputResolver` independently creates `concurrentInputResolutionPermits` (`MAX_CONCURRENT_INPUT_RESOLUTIONS = 64`, `QueryInputResolver.java:112`) and its own executor. Both are documented as *the* application-wide cap on concurrent metadata-store I/O.

Because a single `GetUserObjects` request calls into both — `awaitCancellableMetadataLookup` on the bundle service *and* `collectChunkPins` → `QueryInputResolver` — up to 128 concurrent metadata-store RPCs can be in flight against a bound that was designed to allow only 64. Under load, this doubles store pressure and defeats the documented backpressure guarantee.

Unify the two pools into a single shared semaphore and executor, or explicitly document the intended combined cap and adjust both limits accordingly.

---

🤖 _Generated with Claude Opus 5 and Claude Sonnet 5._

core/scanner/src/main/java/ai/floedb/floecat/scanner/spi/CatalogOverlay.java
```diff
@@ -193,6 +193,18 @@ default java.util.Map<NameRef, Optional<ResourceId>> resolveNames(
...
 
+  /**
+   * Whether independent resolution callbacks may run concurrently on this overlay instance.
+   *
+   * <p>The default preserves compatibility for existing overlays, whose lifecycle state may be tied
+   * to one request thread. Implementations backed by thread-safe services may opt in to concurrent
+   * resolution. Opting in permits {@link #catalog}, {@link #resolve}, {@code resolveName(s)}, and
+   * {@link #tablePinFor} callbacks to execute concurrently and off the caller thread.
+   */
+  default boolean supportsConcurrentResolution() {
```
jchauzi-floe, line 210: [major] **`supportsConcurrentResolution()` authorizes far more off-thread work than it documents**

The javadoc says opting in "permits `catalog`, `resolve`, `resolveName(s)`, and `tablePinFor` callbacks to execute concurrently and off the caller thread". In practice, `UserObjectBundleService` gates much more on the same flag: `awaitCancellableMetadataLookup` (`:1988`) drives `buildRelationCancellably` (`:2068`), so opting in also moves `overlay.tableSchema`, `overlay.schemaFor`, `overlay.tableName`/`viewName`, `pinValidator.validate`, `statsProvider.tableStats`, and the entire `EngineMetadataDecorator` chain (`decorateRelation`, `decorateView`, `decorateColumn`, `completeRelation`) onto `floecat-bundle-metadata-*` platform threads.

Those threads only get what `PropagatedContext` re-establishes. Any overlay implementation that relies on caller-thread state (security context, tenant context, MDC) and opts in based solely on the javadoc contract will silently lose that state for all the undocumented callbacks.

Either expand the javadoc to enumerate every callback that executes off-thread, or restructure so the flag's actual scope matches its documented scope.

---

🤖 _Generated with Claude Opus 5 and Claude Sonnet 5._

service/src/main/java/ai/floedb/floecat/service/query/resolver/QueryInputResolver.java
```diff
@@ -188,15 +428,29 @@ synchronized void register(TablePin pin) {
...
+      }
       List<String> roots = rootsByPin.get(pin);
       if (roots != null) {
         queryStore.releaseResolvingPinBlobs(queryId, roots);
         rootsByPin.remove(pin);
       }
     }
 
+    /**
+     * Release every registration still owned by this failed attempt. A store failure stops release
```
jchauzi-floe, line 443: [minor] **Test-only `Map` overload and its ~200-line compatibility layer appear to have no production caller**

The `resolveInputs(..., Map<ResourceId, TablePin> currentSnapshotPinCache, ...)` overload (`:447`) and its supporting machinery — `resolveInputsViaLegacyOverride`, `CurrentSnapshotCacheOwnership`-aware snapshot/copy-back/rollback (`snapshotCompletedPinCache`, `toSingleFlightCache`, `restoreCompletedPinCache`, `copyCompletedPins`, `completedPins`) — add roughly 200+ lines of intricate concurrency-compensation logic.

Both production call sites that formerly passed a plain `Map` (`QueryInputMetadataAssembler.java`, `QuerySchemaServiceImpl.java`) were migrated in this PR to the `ConcurrentMap` overload. The only remaining exerciser of the plain-`Map` overload is `UserObjectBundleTestSupport` / `TestQueryInputResolver`. Keeping this dead production path costs a `ThreadLocal<ResolutionInvocation>`, four public `resolveInputs` overloads whose interactions are hard to trace, and a subtle abandonment risk: when copy-back fails and rollback cannot be proven, the code deliberately leaves the abandoned pin in the cache.

Drop the `Map` overload and its supporting machinery; if the test seam is still needed, provide it via a purpose-built test helper rather than a production virtual-dispatch seam.

---

🤖 _Generated with Claude Opus 5 and Claude Sonnet 5._

service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
```diff
@@ -0,0 +1,465 @@
...
+  /**
+   * Wait for submitted tasks after scheduling or ordered consumption fails, while still returning
+   * promptly when the stream is cancelled. Without cancellation the original failure remains
+   * caller-visible.
+   */
+  private static void awaitSubmittedTasks(
+      List<? extends CompletionSlot<?>> active, BooleanSupplier cancelled) {
+    for (CompletionSlot<?> slot : active) {
+      while (true) {
+        checkCancelled(cancelled, active);
```
jchauzi-floe: [minor] **Cleanup path replaces the caller's real failure with `CancellationException`**

`completeAllCancellable` catches a task/consumer failure and calls `awaitSubmittedTasks(active, cancelled)` before rethrowing (`:214`). But `awaitSubmittedTasks` starts each wait with `checkCancelled(cancelled, active)`, which throws `cancelled()` — so when the stream is cancelled while this drain is in progress, the `CancellationException` escapes from inside the `catch` block and `processingFailure` is never thrown or recorded as a suppressed cause.

This inverts the precedence `StreamTelemetryGate` was built to enforce: a genuine resolution error (a pin conflict, a `NOT_FOUND`) that races a client disconnect is reported to `UserObjectBundleService` as a `CancellationException`, so `next()` leaves `terminalOutcome == NONE` and the stream is published as `"cancelled"` instead of the real error. Debugging becomes much harder.

In the failure-drain path, either suppress the `CancellationException` and continue draining, or attach it as a suppressed cause on `processingFailure` before rethrowing.

---

🤖 _Generated with Claude Opus 5 and Claude Sonnet 5._

service/src/main/java/ai/floedb/floecat/service/query/resolver/QueryInputResolver.java
```diff
@@ -81,24 +106,139 @@
...
+    }
+    return clamped;
+  }
+
+  /**
+   * Create the owned planning and metadata executors after validating their admission invariant.
+   */
+  @PostConstruct
+  void postConstruct() {
+    if (MAX_CONCURRENT_INPUT_RESOLUTIONS > METADATA_IO_POOL_SIZE) {
```
jchauzi-floe: [nit] **Admission invariant guard can never fire**

`METADATA_IO_POOL_SIZE` is defined as `MAX_CONCURRENT_INPUT_RESOLUTIONS` (`:116`), so the guard `MAX_CONCURRENT_INPUT_RESOLUTIONS > METADATA_IO_POOL_SIZE` compares a constant with itself and is always `false`. The same dead guard exists in `UserObjectBundleService.newMetadataLookupExecutor()` (`UserObjectBundleService.java:255`).

If someone lowers the pool size or queue capacity independently, nothing trips. Either check something that can actually diverge (`METADATA_IO_QUEUE_CAPACITY`, or the pool's `getCorePoolSize()` after construction), or drop the guard and leave only the comment as the contract.

---

🤖 _Generated with Claude Opus 5 and Claude Sonnet 5._

service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
```diff
@@ -0,0 +1,465 @@
...
+   * As {@link #mapOrdered(List, int, Executor, Function)}, but {@code cancelled} is polled so a
+   * cancelled stream interrupts every submitted task and returns without waiting for a slow task's
+   * completion. Tasks must cooperate with interruption while blocked in downstream calls. The
+   * submission window observes cancellation before it submits more work. A non-cancellation task
+   * failure waits for every input task before surfacing, so callers can safely release task-owned
+   * resources. {@code cancelled} may be read concurrently by the scheduler and workers; it must be
+   * non-blocking and thread-safe (typically {@link java.util.concurrent.atomic.AtomicBoolean#get}).
+   * Observed cancellation throws {@link CancellationException}.
+   */
+  public static <I, O> List<O> mapOrdered(
```
jchauzi-floe: [nit] **Two of four public entry points are exercised only by tests**

Production uses exactly two `BoundedFanout` entry points: `mapOrdered(List, int, Executor, Function)` from `NameResolver.parallelScan` and `forEachOrdered(..., BooleanSupplier)` from `QueryInputResolver.planInputsConcurrently`. The cancellable `mapOrdered` (`:94`) and the non-cancellable `forEachOrdered` (`:73`) have no caller outside `BoundedFanoutTest`.

For a new shared concurrency primitive that is hard to reason about, speculative surface means speculative javadoc contracts — the `orderedResults` failure-precedence guarantee for the cancellable map and `awaitCompletedTasks` for the non-cancellable consumer are asserted but never load-bearing in production. Drop the two unused overloads until a second caller needs them; they are thin wrappers and easy to restore.

---

🤖 _Generated with Claude Opus 5 and Claude Sonnet 5._

service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
```diff
@@ -0,0 +1,403 @@
...
+    validatePermits(permits);
+    List<CompletionSlot<O>> active = new ArrayList<>(permits);
+    int next = 0;
+    try {
+      while (next < items.size() && active.size() < permits) {
+        runtime.beforeSubmit(active);
+        active.add(runtime.submit(next, items.get(next++)));
+      }
+      while (!active.isEmpty()) {
+        CompletionSlot<O> slot = runtime.take(active);
```
jchauzi-floe, line 119: [blocker] **Fan-out scheduler blocks unmanaged on the executor it submits to — deadlocks on `ForkJoinPool`**

`completeAll` blocks the calling thread in `runtime.take(active)` — `LinkedBlockingQueue.take()` for `mapOrdered` (line 186) and `poll()` for `forEachOrdered` (line 266) — while the tasks it is waiting for were submitted to the *same* `executor`. Neither is a `ForkJoinPool.ManagedBlocker`, so a `ForkJoinPool` worker that becomes a scheduler consumes a parallelism slot without compensation and its own submitted tasks can never run.

Both callers document `ForkJoinPool.commonPool()` as their executor fallback: `QueryInputResolver.blockingExecutor` (`QueryInputResolver.java:138`, used until `@PostConstruct` installs the virtual-thread pool) and `NameResolver.blockingExecutor` (`NameResolver.java:80`, used when no `ManagedExecutor` bean is present). I reproduced a hard hang:

```
mvn -o -pl service surefire:test \
  -Dtest='QueryInputResolverTest#boundsMetadataPinResolutionAcrossConcurrentRequests' \
  -DargLine="-Djava.util.concurrent.ForkJoinPool.common.parallelism=1"
```

ran for 598s with no progress. `jstack`:

```
"ForkJoinPool.commonPool-worker-1" TIMED_WAITING
  at java.util.concurrent.LinkedBlockingQueue.poll
  at BoundedFanout$CancellableTaskRuntime.take(BoundedFanout.java:266)
  at BoundedFanout.completeAll(BoundedFanout.java:119)
  at QueryInputResolver.planInputsConcurrently(QueryInputResolver.java:868)
```

This is also a regression for `NameResolver.parallelScan`. The code it replaced used `CompletableFuture.join()`, which *is* a `ManagedBlocker` (`CompletableFuture$Signaller`) and additionally calls `ForkJoinPool.helpAsyncBlocker`, so it both compensated and help-ran pending `AsyncSupply` tasks inline. `completions.take()` does neither, so the same-pool case that used to be safe now starves.

Pick one of:

- Reject `ForkJoinPool` (or any executor that can be the caller's own pool) at the `BoundedFanout` entry points, the way `CancellableCallRunner.SubmittedCall.run` already rejects a same-thread executor — and change both fallbacks to a dedicated pool.
- Wrap the queue wait in a `ForkJoinPool.ManagedBlocker` so a `ForkJoinPool` scheduler compensates.

The second is the smaller change and keeps the documented fallbacks working.

---

🤖 _Generated with Claude Opus 5._

service/src/test/java/ai/floedb/floecat/service/concurrent/BoundedFanoutTest.java
```diff
@@ -0,0 +1,587 @@
...
+                    },
+                    value -> {
+                      if (value == 0) {
+                        throw new IllegalStateException("ordered merge failed");
+                      }
+                    },
+                    () -> false);
+              });
+
+      assertThat(activeSiblingsStarted.await(1, TimeUnit.SECONDS)).isTrue();
```
jchauzi-floe, line 247: [blocker] **`ci-localstack` is red: two racy assertions in `BoundedFanoutTest`**

`gh pr checks 415` shows `ci-localstack` failing on this head with `Tests run: 1729, Failures: 1, Errors: 1`, both in code this PR adds:

```
[ERROR] BoundedFanoutTest.orderedConsumerFailureCancelsActiveSiblingsAndReturnsPromptly:247
        Expecting value to be true but was false
[ERROR] QueryInputResolverTest.boundsMetadataPinResolutionAcrossConcurrentRequests
        AssertionError: test did not release the pin lookups
```

Two distinct races in this file, both caused by asserting that a task body definitely ran when cancellation can win:

- Line 247 waits for `activeSiblingsStarted` (2 counts, from items 1 and 2). `completeAll` refills the window *before* calling `onCompletion` (line 125 before line 127), so item 2 is submitted and then the ordered consumer throws for item 0, and `afterFailure` → `cancelSubmittedTasks` can `FutureTask.cancel` item 2 before its pool thread picks it up. `highestStarted` is then 1, and line 252's `isEqualTo(2)` is racy for the same reason.
- `submissionFailureDoesNotWaitForAlreadySubmittedTasks:404` waits for `taskStarted`, but `afterFailure` → `cancelCompletionTasks` cancels the `supplyAsync` future; `CompletableFuture$AsyncSupply.run` checks `d.result == null` before invoking the supplier, so a cancellation that lands first means the task body never runs. I reproduced this one locally — it failed on 1 of 5 consecutive `BoundedFanoutTest` runs.

The assertions should be one-sided: assert the failure propagates promptly and that *any* sibling that did start got interrupted, rather than that a specific sibling started. `QueryInputResolverTest.boundsMetadataPinResolutionAcrossConcurrentRequests` passes locally on a 10-core machine but is the commonPool-starvation case from the other finding — fixing that should fix it.

---

🤖 _Generated with Claude Opus 5._

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -1710,19 +1987,165 @@ private void fillPending() {
...
+     * Concurrent overlays use bounded off-thread I/O and return promptly when the subscriber
+     * cancels. An overlay that opts out stays on the producer thread because it may own
+     * thread-confined state; cancellation is then observed immediately before and after its call.
+     */
+    private <T> T awaitCancellableMetadataLookup(Supplier<T> lookup) {
+      if (!overlay.supportsConcurrentResolution()) {
+        return metadataIoRunner.callOnCallerThread(
+            this::isCancelled, lookup, METADATA_LOOKUP_FAILURES);
+      }
+      return metadataIoRunner.call(this::isCancelled, lookup, METADATA_LOOKUP_FAILURES);
```
jchauzi-floe, line 2033: [major] **In-memory graph reads now hop through the global metadata-I/O pool per relation**

`awaitCancellableMetadataLookup` routes every overlay-owned lookup in relation assembly through `metadataIoRunner.call(...)`: `canonicalName` → `overlay.tableName`/`viewName`, plus `overlay.tableSchema` or `overlay.schemaFor`. For the production overlay these are pure in-memory reads — `MetaGraph.tableName` is `userGraph.tableName(id)` then `systemGraph.tableName(id, ctx)`; `MetaGraph.tableSchema` is `resolve(id)` plus a schema map. No blocking I/O.

Each now costs a semaphore acquire, a `FutureTask` submission to the bounded platform pool, a `CompletableFuture.get` poll, and a return hop — roughly two round trips per relation. The bigger problem is that the admission semaphore is process-wide and shared with genuinely slow work: `floecat.query.metadata_io.max_concurrency` defaults to 64 for the whole JVM, and `CancellableCallRunner` holds a permit for the full duration of the downstream call. A burst of slow blob-backed `schemaFor` reads can hold all 64 permits, so a `tableName` lookup that would have returned in nanoseconds queues behind them. That is head-of-line blocking introduced into a latency path this PR is meant to speed up.

Suggestions, in order of preference:

- Only dispatch the lookups that can actually block (`schemaFor`, and `tableSchema` for overlays that fetch), keeping the graph name lookups on the producer thread.
- Or give cheap graph reads their own admission bucket so they cannot be starved by blob I/O.

Either way, please add the per-relation hop count to the diagnostics so this is visible if it does regress.

---

🤖 _Generated with Claude Opus 5._

service/src/main/java/ai/floedb/floecat/service/query/resolver/QueryInputResolver.java
```diff
@@ -135,46 +271,146 @@ private static final class ResolutionState {
...
+    /** Forget a holder that its lookup failure already evicted from the caller cache. */
+    synchronized void forget(ResourceId tableId, CompletableFuture<TablePin> holder) {
+      owned.remove(tableId, holder);
+    }
+
+    /**
+     * Retire a completed CURRENT entry before its pin relinquishes transient-root ownership. The
+     * holder lock makes cache removal atomic with a waiter's published-entry check.
+     */
+    synchronized void retire(ResourceId tableId, TablePin pin) {
```
jchauzi-floe, line 361: [minor] **`retire` evicts a stream-scoped CURRENT entry it does not own**

`discardCompatiblePin` calls `retire(tableId, pin)`, which removes the completed single-flight holder from `state.currentSnapshotPinCache`. In `UserObjectBundleService` that map is the iterator's `currentSnapshotPinCache`, which lives for the whole `GetUserObjects` stream — not for one `resolveInputs` call — and the holder may have been inserted by an earlier chunk (`owned.remove(tableId, holder)` is a no-op in that case, which is the tell).

So when table `T` is pinned in one chunk both explicitly and via CURRENT, and the CURRENT pin loses first-touch, `T`'s CURRENT entry disappears from the stream cache. A later chunk referencing `T` via CURRENT re-issues a `tablePinFor` round trip, registers a fresh transient root for the new pin, and then has that pin silently dropped by `mergeRelationPins` at commit time because the committed pin wins — leaving the new root registered until the resolving-pin grace period expires. Before this PR the entry stayed cached and the second lookup never happened.

The eviction also looks unnecessary: `QueryPins.compatible` requires a matching `snapshot_id` (or an incoming CURRENT, which is unconditionally compatible), the winning pin registered the same blob URIs, and `registerResolvingPinBlobs` is refcounted per URI — so the cached pin stays rooted through the winner. Dropping `retire` and keeping only `resolvingPinRoots.discard(pin)` in `discardCompatiblePin` restores the cross-chunk single flight.

---

🤖 _Generated with Claude Opus 5._

service/src/main/java/ai/floedb/floecat/service/concurrent/MetadataIoRunner.java
```diff
@@ -0,0 +1,228 @@
...
+    private synchronized void start() {
+      if (executor == null || executor.isShutdown()) {
+        executor =
+            MetadataIoExecutors.newBoundedDaemonPool(capacity, capacity, "floecat-metadata-io-");
+      }
+    }
+
+    private ExecutorService executor() {
+      ExecutorService current = executor;
+      if (current == null || current.isShutdown()) {
```
jchauzi-floe, line 211: [minor] **`executor()` silently resurrects a pool that `@PreDestroy` closed**

`RuntimeState.close()` nulls `executor`, and `executor()` treats `null || isShutdown()` as "start a new one". Because the `@ApplicationScoped` bean and every `MetadataIoRunner.shared()` facade wrap the same `SharedRuntime.RUNTIME`, bean destruction closes the pool that the non-CDI compatibility constructors still hold — and the next call from any of them creates a brand-new 64-thread pool during shutdown.

CDI destruction order across `UserObjectBundleService`, `QueryInputResolver`, and `MetadataIoRunner` is not guaranteed, so this is reachable on a normal shutdown: `close()` logs its 5s timeout warning, then a straggler request spins the pool back up and the JVM only exits because the threads are daemons.

Make closure terminal — a `closed` flag that makes `executor()` throw (or that `call`/`callWithoutCancellation` translate into the existing `"resolver metadata I/O executor shut down"` failure) rather than restart. `start()` can stay idempotent for the repeated-lifecycle case it already handles.

---

🤖 _Generated with Claude Opus 5._

mrouvroy-floe:
Superseded by a clean re-cut of this work into a linear, capability-scoped stack, rebased onto `main`:

- #432 — concurrency: bounded ordered fan-out with request-context propagation
- #433 — concurrency: bounded metadata-I/O admission with cancellable calls
- #434 — concurrency: automatic metadata-I/O admission + admission-free fan-out stage
- #435 — resolver: resolve a query's pins concurrently and safely
- #436 — catalog bundle: decompose chunk assembly into structural stages

The new stack re-divides the same body of work so that each PR builds and tests independently, moves metadata-I/O admission to an automatic, unbypassable repository-layer interceptor (rather than a per-call opt-in that was easy to forget and impossible to enforce), and drops the throwaway intermediate churn — `MetadataFanout` is now born admission-free instead of being introduced two-tier and rewritten a commit later.

Closing in favour of the new stack. Thanks! 🙏

PR #416, opened 2026-07-27
GetUserObjects - 3: parallelize ordered candidate selection

mrouvroy-floe:
## Summary

Restructure chunk filling into plan, select, and gather stages. 
Select inputs concurrently while preserving candidate fallback order, deterministic output ordering,
and spillover across chunk boundaries.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
+++ b/service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
@@ -101,6 +101,7 @@ import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BooleanSupplier;
 import java.util.function.Function;
+import java.util.function.LongConsumer;
 import java.util.function.Supplier;
 import org.eclipse.microprofile.config.Config;
 import org.eclipse.microprofile.config.ConfigProvider;
@@ -1601,9 +1602,13 @@ public class UserObjectBundleService {
 
     // Maintains the order inputs were resolved so the emitted chunk mirrors the request order.
     private final List<PendingItem> pending = new ArrayList<>(MAX_RESOLUTIONS_PER_CHUNK);
+    // Per-request resolution memos. Both resolve each key at most once: a repeated lookup returns
+    // the stored Optional (present or empty). See memoize() / resolveNameCached /
+    // resolveNodeCached.
     private final Map<NormalizedNameRef, Optional<ResourceId>> nameResolutionCache =
-        new HashMap<>();
-    private final Map<ResourceId, Optional<GraphNode>> nodeResolutionCache = new HashMap<>();
+        new ConcurrentHashMap<>();
+    private final Map<ResourceId, Optional<GraphNode>> nodeResolutionCache =
+        new ConcurrentHashMap<>();
     private final ArrayDeque<EagerBaseCursor> eagerBaseQueue = new ArrayDeque<>();
     private final Set<String> eagerBaseSeen = new HashSet<>();
     private final Map<RelationCacheKey, RelationInfo> relationInfoCache = new HashMap<>();
@@ -1651,7 +1656,9 @@ public class UserObjectBundleService {
       this.correlationId = correlationId;
       this.ctx = ctx;
       this.tables = tables;
-      this.knownBlobVersions = knownBlobVersions;
+      // Read-only for the life of the iterator (consulted by the identity-only fast path); copy so
+      // that stays true regardless of what the caller does with its set afterwards.
+      this.knownBlobVersions = Set.copyOf(knownBlobVersions);
       this.resolutionCount = tables.size();
       this.defaultCatalogId = ctx.getQueryDefaultCatalogId();
       this.statsProvider = statsFactory.forQuery(ctx, correlationId);
@@ -2172,46 +2179,65 @@ public class UserObjectBundleService {
     }
 
     private Optional<ResourceId> resolveNameCached(NameRef ref) {
-      NormalizedNameRef key = normalizedNameRef(ref);
-      // Stored values are non-null Optional instances; null means cache miss.
-      Optional<ResourceId> cached = nameResolutionCache.get(key);
-      if (cached != null) {
-        nameResolutionCacheHits++;
-        return cached;
-      }
-      long startNs = System.nanoTime();
-      try {
-        // Pass the engine captured at iterator construction: re-reading it from the request
-        // context per lookup is fragile across executor hops, and an empty engine silently
-        // un-resolves engine-gated system objects (eng-floe/floecat#361).
-        Optional<ResourceId> resolved =
-            overlay.resolveName(correlationId, ref, resolutionContext.engineContext());
-        nameResolutionCache.put(key, resolved);
+      // Engine captured at iterator construction is passed through: re-reading it from the request
+      // context per lookup is fragile across executor hops, and an empty engine silently
+      // un-resolves engine-gated system objects (eng-floe/floecat#361).
+      Memoized<ResourceId> m =
+          memoize(
+              nameResolutionCache,
+              normalizedNameRef(ref),
+              () -> overlay.resolveName(correlationId, ref, resolutionContext.engineContext()),
+              nanos -> nameResolveNanos += nanos);
+      if (m.resolved()) {
         nameResolutionCacheMisses++;
-        return resolved;
-      } finally {
-        nameResolveNanos += System.nanoTime() - startNs;
+      } else {
+        nameResolutionCacheHits++;
       }
+      return m.value();
     }
 
     private Optional<GraphNode> resolveNodeCached(ResourceId id) {
-      // Stored values are non-null Optional instances; null means cache miss.
-      Optional<GraphNode> cached = nodeResolutionCache.get(id);
-      if (cached != null) {
-        nodeResolutionCacheHits++;
-        return cached;
-      }
-      long startNs = System.nanoTime();
-      try {
-        // Pass the engine captured at iterator construction: re-reading it from the request
-        // context per lookup is fragile across executor hops (eng-floe/floecat#361).
-        Optional<GraphNode> resolved = overlay.resolve(id, resolutionContext.engineContext());
-        nodeResolutionCache.put(id, resolved);
+      Memoized<GraphNode> m =
+          memoize(
+              nodeResolutionCache,
+              id,
+              () -> overlay.resolve(id, resolutionContext.engineContext()),
+              nanos -> nodeResolveNanos += nanos);
+      if (m.resolved()) {
         nodeResolutionCacheMisses++;
-        return resolved;
-      } finally {
-        nodeResolveNanos += System.nanoTime() - startNs;
+      } else {
+        nodeResolutionCacheHits++;
       }
+      return m.value();
+    }
+
+    /**
+     * A memoized value together with whether this call resolved it (a miss) vs. found it cached.
+     */
+    private record Memoized<V>(Optional<V> value, boolean resolved) {}
+
+    /**
+     * Return {@code cache.get(key)}, resolving and storing it once if absent. The resolve runs at
+     * most once per key even under concurrent callers (computeIfAbsent single-flight); its elapsed
+     * time is reported to {@code addNanos}. {@link Memoized#resolved()} is true when this call ran
+     * the resolve.
+     */
+    private <K, V> Memoized<V> memoize(
+        Map<K, Optional<V>> cache, K key, Supplier<Optional<V>> resolve, LongConsumer addNanos) {
+      boolean[] resolvedHere = {false};
+      Optional<V> value =
+          cache.computeIfAbsent(
+              key,
+              k -> {
+                resolvedHere[0] = true;
+                long startNs = System.nanoTime();
+                try {
+                  return resolve.get();
+                } finally {
+                  addNanos.accept(System.nanoTime() - startNs);
+                }
+              });
+      return new Memoized<>(value, resolvedHere[0]);
     }
 
     private Optional<ResolvedRelation> selectResolvedRelationTimed(
--- a/service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
+++ b/service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
@@ -1763,7 +1763,6 @@ public class UserObjectBundleService {
           plan.add(planInput(nextInputIndex + i));
         }
         nextInputIndex += planCount;
-        seedNameResolutions(plan);
         // Resolve the planned inputs concurrently; results come back in plan order so the gather
         // below stays deterministic. resolveNanos is the wall-clock of this stage.
         long selectStageStartNs = System.nanoTime();
--- a/service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
+++ b/service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
@@ -1723,7 +1723,10 @@ public class UserObjectBundleService {
         return headerChunk(ctx.getQueryId(), seq++);
       }
 
-      if (pending.isEmpty() && (nextInputIndex < resolutionCount || !eagerBaseQueue.isEmpty())) {
+      if (pending.isEmpty()
+          && (nextInputIndex < resolutionCount
+              || !eagerBaseQueue.isEmpty()
+              || !resolvedSpillover.isEmpty())) {
         fillPending();
       }
 
--- a/service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
+++ b/service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
@@ -82,7 +82,6 @@ import ai.floedb.floecat.types.LogicalTypeFormat;
 import io.opentelemetry.api.trace.Span;
 import io.smallrye.mutiny.Multi;
 import jakarta.enterprise.context.ApplicationScoped;
-import jakarta.enterprise.inject.Instance;
 import jakarta.inject.Inject;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
@@ -100,16 +99,17 @@ import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.function.BooleanSupplier;
 import java.util.concurrent.atomic.LongAdder;
+import java.util.function.BooleanSupplier;
 import java.util.function.Function;
 import java.util.function.LongConsumer;
 import java.util.function.Supplier;
 import org.eclipse.microprofile.config.Config;
 import org.eclipse.microprofile.config.ConfigProvider;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
-import org.eclipse.microprofile.context.ManagedExecutor;
 import org.jboss.logging.Logger;
 
 @ApplicationScoped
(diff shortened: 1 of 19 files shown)


service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -1698,15 +1737,36 @@ public UserObjectsBundleChunk next() {
...
+        gather(resolvedSpillover.removeFirst(), toPin);
+      }
+      // Then newly requested inputs, up to this chunk's remaining capacity. Selecting is separated
+      // from gathering so the select step can later run in parallel; today it is a serial loop.
+      int budget = MAX_RESOLUTIONS_PER_CHUNK - pending.size();
+      if (budget > 0 && nextInputIndex < resolutionCount) {
+        int planCount = Math.min(budget, resolutionCount - nextInputIndex);
+        List<PlannedInput> plan = new ArrayList<>(planCount);
+        for (int i = 0; i < planCount; i++) {
+          plan.add(planInput(nextInputIndex + i));
```
jchauzi-floe: [major] **Defer failures with spilled inputs**

`fillPending()` plans and joins every input allowed by the pre-expansion budget before it gathers the first result. If input 0 is a view whose eager bases fill the chunk and input 1's `planInput()` or `selectOne()` throws, that later failure aborts the call here; the view and its bases are never gathered or emitted.

Before this change, the serial loop stopped as soon as the view filled the chunk, emitted that ordered prefix, and attempted input 1 only on the next pull. An input that should spill into the next chunk can therefore overtake and suppress earlier output, breaking the promised deterministic spillover behavior.

Capture planning and selection failures as ordered outcomes. Gather those outcomes in request order, put unconsumed outcomes—including a failure—into `resolvedSpillover`, and rethrow only when the failed outcome reaches the front on a later chunk. Please add a regression test with a chunk-filling view followed by an invalid or throwing candidate.

---

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -1698,15 +1737,36 @@ public UserObjectsBundleChunk next() {
...
+        }
+        nextInputIndex += planCount;
+        // Resolve the planned inputs concurrently; results come back in plan order so the gather
+        // below stays deterministic. resolveNanos is the wall-clock of this stage.
+        long selectStageStartNs = System.nanoTime();
+        List<PendingItem> selected =
+            BoundedFanout.mapOrdered(
```
jchauzi-floe: [major] **Propagate stream cancellation into `BoundedFanout`**

This calls the `BoundedFanout.mapOrdered()` overload with no cancellation signal; that overload waits for every submitted future. The stream's `onTermination()` callback only publishes telemetry, so cancelling the subscription never tells the iterator or these tasks to stop.

If a client disconnects or its deadline expires during slow name/node reads, as many as 25 submitted tasks continue running (eight concurrently), and the driver remains blocked until all of them finish. Repeated cancelled RPCs can consume executor and metadata-store capacity long after their callers are gone.

Set an atomic flag from `.onCancellation()`, use the cancellation-aware `BoundedFanout` overload (or equivalently cancel the submitted futures), and add a blocking-resolver test that verifies cancellation interrupts the work and terminates promptly.

---

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -134,6 +141,21 @@ public class UserObjectBundleService {
 
   @Inject Observability observability;
 
+  // Caps how many of a chunk's relations resolve concurrently. Each is an independent, mostly
+  // store-bound resolution; a small fan-out overlaps their round-trips without flooding the store.
+  private static final int MAX_PARALLEL_RELATION_TASKS = 8;
+
+  // Runs per-relation resolution off the request thread. The Quarkus ManagedExecutor is injected in
```
jchauzi-floe: [major] **Keep blocking fan-out off the driver pool**

The stream body runs on Mutiny's default executor (`BaseServiceImpl.runStreamEmitter()`), while this field is replaced in production with Quarkus's shared `ManagedExecutor`. `BoundedFanout.mapOrdered()` then blocks the stream-driver thread while joining tasks submitted to that managed worker pool.

When those executors share the application worker pool, enough concurrent streams can occupy every worker waiting for selection tasks that are queued to the same exhausted pool. The sibling `QueryInputResolver` already avoids this pattern with a virtual-thread executor because its driver also blocks while joining a fan-out.

Use a dedicated `ExecutorService` (for example, virtual threads with the existing semaphore as the per-request bound), or make the stage non-blocking so drivers do not wait on their own pool. Add a saturation test with more concurrent stream drivers than worker threads.

---

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -1713,17 +1745,55 @@ public UserObjectsBundleChunk next() {
...
+            plan.add(PlannedInput.failed(inputIndex, e));
           }
         }
+        nextInputIndex += planCount;
+        // Resolve and consume planned inputs concurrently in request order. Consuming each result
+        // as it is joined gives an earlier deferred failure precedence over a later task failure,
+        // while still letting an eager view fill this chunk and spill later requested inputs.
+        long selectStageStartNs = System.nanoTime();
+        try {
+          BoundedFanout.forEachOrdered(
```
jchauzi-floe, line 1778: [major] **Respect `CatalogOverlay` concurrency opt-in**

`BoundedFanout.forEachOrdered()` unconditionally runs several `selectOne()` calls at once. Those tasks call `overlay.resolveName()` and `overlay.resolve()` directly through `resolveNameCached()` / `resolveNodeCached()`, so distinct cache keys enter the same overlay concurrently.

`CatalogOverlay.supportsConcurrentResolution()` defaults to `false` specifically because existing implementations are not required to be thread-safe. The base branch's `QueryInputResolver.withMetadataGraph()` honors that contract, but this new path does not; a non-opted-in overlay can therefore race or corrupt its request-scoped state during `GetUserObjects`.

Gate the fan-out on `overlay.supportsConcurrentResolution()`, or serialize each overlay callback with the same compatibility lock used by `QueryInputResolver`. Add a default-`false` overlay test that detects overlapping callbacks.

---

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -143,6 +148,14 @@ private static void throwIfCancelled(BooleanSupplier cancelled) {
 
   @Inject Observability observability;
 
+  // Caps how many of a chunk's relations resolve concurrently. Each is an independent, mostly
+  // store-bound resolution; a small fan-out overlaps their round-trips without flooding the store.
+  private static final int MAX_PARALLEL_RELATION_TASKS = 8;
```
jchauzi-floe, line 153: [major] **Honor `max_parallel_relations`**

The new selection stage hard-codes its concurrency to `8`, ignoring `floecat.catalog.bundle.max_parallel_relations`. The base branch's `QueryInputResolver` already reads this property and explicitly defines it as shared with the catalog bundle's relation fan-out.

An operator who lowers the setting to protect a constrained metadata store still gets up to eight concurrent name/node resolutions per stream in this stage. That defeats the safety control precisely where this PR adds new store concurrency.

Read and clamp the existing property in `UserObjectBundleService`, use that value in `BoundedFanout.forEachOrdered()`, and cover a configured value such as `1` with an overlap test.

---

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -278,8 +291,10 @@ public Multi<UserObjectsBundleChunk> stream(
                   .iterable(() -> iterator)
                   .onFailure()
                   .invoke(ignored -> iterator.publishStreamTelemetry("failed"))
+                  .onCancellation()
+                  .invoke(iterator::cancel)
                   .onTermination()
-                  .invoke(iterator::cancel);
+                  .invoke(() -> iterator.publishStreamTelemetry("terminated"));
```
jchauzi-floe, line 297: [minor] **Preserve the `cancelled` telemetry outcome**

The outer `onTermination()` callback runs before cancellation is propagated upstream to `onCancellation()`. It therefore calls `publishStreamTelemetry("terminated")` first; that wins `telemetryPublished.compareAndSet(...)`, and the later `iterator.cancel()` cannot publish `"cancelled"`.

Every client cancellation is consequently reported as `outcome=terminated`, breaking the cancellation diagnostics this change is meant to preserve. Normal completion and failure already publish their own outcomes, so `terminated` only wins in this cancellation path.

Put `onCancellation()` outside `onTermination()`, or use the termination callback's cancellation flag to call `iterator.cancel()` before publishing a fallback outcome. Add a subscription-cancellation test that asserts the emitted outcome is `cancelled`.

---

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -131,6 +148,14 @@ public class UserObjectBundleService {
 
   @Inject Observability observability;
 
+  // Caps how many of a chunk's relations resolve concurrently. Each is an independent, mostly
+  // store-bound resolution; a small fan-out overlaps their round-trips without flooding the store.
+  private static final int MAX_PARALLEL_RELATION_TASKS = 8;
```
jchauzi-floe, line 153: [major] **Fan-out width is hardcoded and not read from the shared config key it claims to share**

`MAX_PARALLEL_RELATION_TASKS = 8` is a compile-time constant with no `ConfigProvider` read, yet `QueryInputResolver.maxParallelInputResolutions` reads `floecat.catalog.bundle.max_parallel_relations` and its field comment explicitly states it is "Shared with the catalog bundle's relation fan-out via the same config property." That claim is false: `UserObjectBundleService` never reads this key and has no tuning warning for it.

An operator adjusting `floecat.catalog.bundle.max_parallel_relations` to reduce store load will affect `QueryInputResolver` but silently leave `UserObjectBundleService` running at 8 — the opposite of the documented shared-config contract.

Additionally, the `Semaphore` inside `BoundedFanout` is constructed per `forEachOrdered()` call alongside an unbounded `Executors.newVirtualThreadPerTaskExecutor()` (line 157), so N concurrent `GetUserObjects` streams produce up to N×8 simultaneous metadata-store reads with no process-wide cap. The sibling `QueryInputResolver` treats this as a hazard and applies both a configurable width and a process-wide permit.

Fix: read the fan-out width from `floecat.catalog.bundle.max_parallel_relations` via `ConfigProvider` (with the same clamping and warning logic as `QueryInputResolver`), and add a process-wide semaphore or shared executor ceiling mirroring the sibling's `withInputResolutionPermit` pattern.

---

🤖 _Generated with Claude Opus 5 and Claude Sonnet 5._

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -1694,23 +1745,68 @@ public UserObjectsBundleChunk next() {
...
+            plan.add(PlannedInput.failed(inputIndex, e));
           }
         }
+        nextInputIndex += planCount;
+        // Resolve and consume planned inputs concurrently in request order. Consuming each result
+        // as it is joined gives an earlier deferred failure precedence over a later task failure,
+        // while still letting an eager view fill this chunk and spill later requested inputs.
+        long selectStageStartNs = System.nanoTime();
+        try {
+          BoundedFanout.forEachOrdered(
```
jchauzi-floe, line 1778: [major] **New fan-out ignores `CatalogOverlay.supportsConcurrentResolution()`**

`BoundedFanout.forEachOrdered()` runs up to eight `selectOne()` calls concurrently, each reaching `overlay.resolveName()` / `overlay.resolve()` through `resolveNameCached()` / `resolveNodeCached()`. Distinct cache keys therefore enter the same overlay instance concurrently, unconditionally.

`CatalogOverlay.supportsConcurrentResolution()` defaults to `false` because overlay implementations are not required to be thread-safe — the javadoc notes that lifecycle state may be tied to one request thread. The sibling `QueryInputResolver` honors this in two places (`planInputsConcurrently` vs. `planInputsSerially`, and `withInputResolutionPermit`); this new path does not. Today's only production implementation (`MetaGraph`) returns `true`, masking the issue in production, but any future or test overlay that returns `false` will race.

Fix: gate the concurrent fan-out path behind `overlay.supportsConcurrentResolution()` and fall back to a serial loop when it returns `false`, matching the pattern already established in `QueryInputResolver`.

---

🤖 _Generated with Claude Opus 5 and Claude Sonnet 5._

service/src/main/java/ai/floedb/floecat/service/query/resolver/QueryInputResolver.java
```diff
@@ -73,22 +86,47 @@
...
   // Registers each resolved pin's blobs as a transient GC root at construction time (see
   // QueryContextStore.registerResolvingPinBlobs). Null in unit tests that construct the resolver
   // without a store — registration is simply skipped then.
   private final QueryContextStore queryStore;
 
+  // Runs per-input resolution off the request thread on virtual threads, not the shared Quarkus
+  // worker pool: the driver blocks joining this fan-out and must not contend with it for the same
+  // pool. The semaphore in BoundedFanout bounds concurrency; OTel context is re-established per
+  // task.
+  private final ExecutorService blockingExecutor = Executors.newVirtualThreadPerTaskExecutor();
```
jchauzi-floe, line 115: [major] **New virtual-thread executor has no `@PreDestroy` shutdown (lifecycle regression)**

`blockingExecutor` is now a bare `Executors.newVirtualThreadPerTaskExecutor()` instance field with no `@PreDestroy`. The pre-PR version explicitly shut down its executors in a `@PreDestroy` method (`ownedBlockingExecutor.shutdownNow()`, `MetadataIoExecutors.shutdownNowAndAwait(...)`) — that lifecycle management was dropped along with the old pool machinery. `UserObjectBundleService.blockingExecutor` (also newly added in this PR) has the same gap.

Leaving these unmanaged means CDI bean destruction (hot redeploy in dev mode, test-container teardown that creates many bean instances) leaks a `VirtualThreadPerTaskExecutor` and any threads still parked in it, rather than the previous explicit-shutdown behavior.

Fix: add a `@PreDestroy` method to both classes that calls `blockingExecutor.shutdownNow()` (and awaits termination if needed), restoring the lifecycle contract that existed before this PR.

---

🤖 _Generated with Claude Opus 5 and Claude Sonnet 5._

service/src/main/java/ai/floedb/floecat/service/query/catalog/UserObjectBundleService.java
```diff
@@ -2137,59 +2300,65 @@ private String defaultCatalogForDiagnostics() {
...
+     * Return {@code cache.get(key)}, resolving and storing it once if absent. The resolve runs at
+     * most once per key even under concurrent callers (computeIfAbsent single-flight); its elapsed
+     * time is reported to {@code addNanos}. {@link Memoized#resolved()} is true when this call ran
+     * the resolve.
+     */
+    private <K, V> Memoized<V> memoize(
+        Map<K, Optional<V>> cache, K key, Supplier<Optional<V>> resolve, LongConsumer addNanos) {
+      boolean[] resolvedHere = {false};
+      Optional<V> value =
+          cache.computeIfAbsent(
```
jchauzi-floe, line 2350: [minor] **Blocking overlay I/O inside `ConcurrentHashMap.computeIfAbsent` serializes concurrent tasks**

`memoize()` executes `resolve.get()` — an `overlay.resolveName()` / `overlay.resolve()` metadata-store round-trip — inside the mapping function of `computeIfAbsent`. `ConcurrentHashMap` holds the bin lock for the entire mapping function; its javadoc requires that "the computation should be short and simple."

Two consequences working against the parallelism this PR adds:
- Two concurrent tasks resolving **different** keys that hash to the same bin serialize on a full network round-trip. With the default 16-bin table and eight concurrent tasks, a collision is probable rather than exotic.
- `computeIfAbsent` triggers resize as the map grows past 12 and 24 entries (a chunk resolves up to 25 keys). A thread helping `transfer` must take each bin lock, stalling behind whatever resolution is in flight.

Fix: use a two-step pattern — check the map first, then compute outside the lock, then `putIfAbsent` — or replace with a `ConcurrentHashMap`-based single-flight map (similar to the pin-lookup cache already in `QueryInputResolver`) that starts the I/O outside any bin lock.

---

🤖 _Generated with Claude Opus 5 and Claude Sonnet 5._

service/src/main/java/ai/floedb/floecat/service/query/resolver/QueryInputResolver.java
```diff
@@ -312,19 +633,43 @@ private TablePin pinForTable(
...
+          // Never cache a failure: drop the placeholder so a retry re-resolves, and release any
+          // callers already awaiting this id with the same error.
+          state.currentSnapshotPinCache.remove(rid, holder);
+          holder.completeExceptionally(e);
+          throw e;
+        }
+        holder.complete(pin);
+        state.diagnostics.count("pin.current_snapshot_cache_misses");
+      } else {
+        pin = Futures.join(inflight);
```
jchauzi-floe, line 668: [minor] **Single-flight pin lookup blocks in `Futures.join`, which ignores thread interruption**

`BoundedFanout`'s cancellable overloads document that cancelled streams interrupt every submitted task, and that tasks must cooperate with interruption while blocked. But the loser side of the single-flight pin-cache blocks in `Futures.join(inflight)` (`QueryInputResolver.java:668`), which delegates to `CompletableFuture.join()` — a call that does not respond to `Thread.interrupt()`.

When a stream is cancelled while thread A holds the placeholder mid-lookup and thread B is blocked joining A's future, interrupting B's carrier thread has no effect: B keeps waiting for A's real completion rather than aborting promptly. This doesn't deadlock — A will eventually finish or fail — but it defeats the cancellation contract, extending the tail latency of cancelled requests.

Fix: replace `Futures.join(inflight)` with a `inflight.get(timeout, unit)` call inside a loop that checks `Thread.interrupted()`, or use `CompletableFuture.orTimeout` combined with an explicit cancellation check so the loser can bail out when its carrier thread is interrupted.

---

🤖 _Generated with Claude Opus 5 and Claude Sonnet 5._

jchauzi-floe (commented):
**[blocker] Test double bypassed: `TestQueryInputResolver` overrides the wrong overload after 8-arg signature change** — `service/src/test/java/ai/floedb/floecat/service/query/catalog/testsupport/UserObjectBundleTestSupport.java:406`

`collectChunkPins` now calls the **8-argument** `resolveInputs(..., diagnostics, cancelled)` overload (`UserObjectBundleService.java:403` → `QueryInputResolver.java:389`), but `TestQueryInputResolver` only overrides the **7-argument** overload (`QueryInputResolver.java:321`). Adding a parameter at the call site silently orphaned the override — it still compiles and `@Override` still matches the older method, so nothing flags it.

The result is that the real `QueryInputResolver` runs instead of the fake, with `metadataGraph == null` from `super(null)`, causing:
```
java.lang.NullPointerException: Cannot invoke "CatalogOverlay.supportsConcurrentResolution()"
    because "this.metadataGraph" is null
    at QueryInputResolver.withMetadataGraph(QueryInputResolver.java:549)
```
This produces 42 test failures on the current head.

Fix: override the 8-argument overload in `TestQueryInputResolver`, replacing or augmenting the 7-argument one.

---

🤖 _Generated with Claude Opus 5 and Claude Sonnet 5._

mrouvroy-floe:
Superseded by a clean re-cut of this work into a linear, capability-scoped stack, rebased onto `main`:

- #432 — concurrency: bounded ordered fan-out with request-context propagation
- #433 — concurrency: bounded metadata-I/O admission with cancellable calls
- #434 — concurrency: automatic metadata-I/O admission + admission-free fan-out stage
- #435 — resolver: resolve a query's pins concurrently and safely
- #436 — catalog bundle: decompose chunk assembly into structural stages

The new stack re-divides the same body of work so that each PR builds and tests independently, moves metadata-I/O admission to an automatic, unbypassable repository-layer interceptor (rather than a per-call opt-in that was easy to forget and impossible to enforce), and drops the throwaway intermediate churn — `MetadataFanout` is now born admission-free instead of being introduced two-tier and rewritten a commit later.

Closing in favour of the new stack. Thanks! 🙏

PR #419, opened 2026-07-29
Nested types: carry the full ARRAY/MAP/STRUCT type tree from Iceberg and Delta

jincongho-floe:
## Summary

 `LogicalType` now carries a recursive type tree, both connectors populate it faithfully, 
  and `SchemaColumn` delivers it to consumers — while keeping the flat tag for legacy 
  readers. Also fixes four pre-existing nested-type bugs found while auditing the type path.

  Previously `array<int>`, `array<string>`, and `map<string,double>` were byte-identical
  in the derived schema (`logical_type = "ARRAY"` / `"MAP"`); element, key, and value
  types were unrecoverable outside the raw schema JSON.

  ## Changes

  ### Type model (`core/types`)
  - `LogicalType` gains a recursive tree: `element()` + `elementNullable()` for ARRAY,
    `key()`/`value()` + `valueNullable()` for MAP, ordered `LogicalField`s for STRUCT.
    Bare container tags stay valid as the legacy form (`hasTypeTree()` distinguishes).
  - `LogicalTypeFormat` gains the matching grammar with a recursive-descent parser:
    `ARRAY<INT>`, `MAP<STRING, DOUBLE>`, `ARRAY<STRUCT<sku: STRING, quantities: ARRAY<INT>>>`.
    Struct field names are case-preserved and quotable. `formatTag()` produces the flat
    legacy spelling.

  ### Wire (`core/proto`)
  - `SchemaColumn.logical_type` keeps the flat container tag for legacy consumers.
  - New `SchemaColumn.logical_type_full` carries the complete grammar string for complex
    columns (empty for scalars / legacy producers). Additive; no breaking change.

  ### Connectors
  - **Iceberg**: `IcebergTypeMappings.toLogical` recurses into LIST/MAP/STRUCT, keeping
    element/key/value/field types and nullability (`element-required`/`value-required`).
    Map key types — previously never captured — are included.
  - **Delta**: `DeltaTypeMapper` and both branches of `DeltaSchemaMapper` (kernel + JSON
    fallback) recurse likewise, preserving `containsNull`/`valueContainsNull`.
  - **Iceberg REST gateway views**: `ViewMetadataService` now converts view schema types
    recursively and emits `logical_type_full`, field IDs, and the leaf flag (previously:
    top-level flattening to bare tags only).

  ### Bug fixes
  - **UC views with complex columns failed schema resolution**: `buildSchemaJson` wrote
    UC's `type_text` display string (e.g. `array<string>`), which no Delta parser accepts.
    Now prefers `type_json` (the serialized StructField), falling back to `type_text`.
  - **Iceberg stats paths disagreed with schema paths**: the stats traversal appended the
    synthetic `element`/`key`/`value` field names on top of the `[]`/`{}` suffix, yielding
    `arr[].element.x` → canonicalized to `arr[][].x`, which never matched the schema's
    `arr[].x`. Paths now match the schema mapper exactly.
  - **`GenericSchemaMapper` marked containers `leaf=true`**, sending them into min/max
    encoding which rejects complex kinds. Containers are now `leaf=false` (VARIANT stays
    leaf), matching the other mappers.
  - **Scanner array detection was dead code**: `RowStreamToArrowBatchAdapter` looked for a
    `"[]"` suffix no producer emits, so real ARRAY columns threw in `ArrowSchemaUtil`
    instead of being stringified. It now recognizes canonical `ARRAY`/`ARRAY<...>`
    spellings; the byte-identical duplicate class in `scanner/utils` is removed.

  ### Docs
  - `docs/types.md`: complex kinds are no longer "non-parameterised in v1" — documents the
    tree, the grammar, and the `logical_type` / `logical_type_full` split.
  - `docs/connectors-delta.md`: corrected the overstated `DeltaTypeMapper` claim.

  ## Compatibility

  - `logical_type` output is unchanged for every existing column shape; the tree rides in
    a new field, so mixed-version consumers are unaffected.
  - `LogicalTypeFormat.parse` accepts everything it accepted before, plus the new grammar.
  - Raw `schema_json` persistence is untouched (it was always lossless at rest).

  ## Testing

  New unit tests: type-tree model + grammar round-trips (16), Iceberg tree mapping,
  Delta tree mapping (kernel + fallback), UC `type_json` view resolution, stats-path
  canonical forms, generic-mapper leaf semantics, scanner stringification of canonical
  arrays, gateway nested view types. Full test runs green on `core/types`,
  `core/connectors/common`, `core/scanner`, `core/arrow`, `core/catalog`, `core/stats`,
  `core/engine-utils`, `connectors/catalogs/iceberg`, `connectors/catalogs/delta`, and
  the touched gateway module.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [ ] `make fmt`
- [ ] `make verify`

- [ ] No breaking changes

## Checklist

## Additional Notes


core/proto/src/main/proto/floecat/query/schema.proto
```diff
@@ -40,14 +40,24 @@ import "floecat/query/pinning.proto";
...
   bool partition_key = 6; // True if this is a partition column.
   int64 id = 7; // Internal unique column ID.
   int32 ordinal = 8; // Ordinal position in the schema (1-based within the parent struct).
   bool leaf = 9; // True if this is a leaf (stats-capable); false for containers.
+
+  // Full recursive canonical type for complex columns, in the nested grammar of
+  // LogicalTypeFormat, e.g. ARRAY<INT>, MAP<STRING, DOUBLE>,
+  // ARRAY<STRUCT<sku: STRING, quantities: ARRAY<INT>>>. Empty when logical_type is already
+  // complete (scalar columns) or when the source tree is unknown (legacy producers).
+  string logical_type_full = 10;
```
jincongho-floe: @mrouvroy-floe What do you think of handling backward-compatibility of new field?

core/types/src/main/java/ai/floedb/floecat/types/LogicalTypeFormat.java
```diff
@@ -83,15 +104,68 @@ public static String format(LogicalType t) {
...
+  }
+
+  private static String formatFieldName(String name) {
+    if (BARE_FIELD_NAME_RE.matcher(name).matches()) {
+      return name;
+    }
+    return '"' + name.replace("\"", "\"\"") + '"';
+  }
+
   public static LogicalType parse(String s) {
```
mrouvroy-floe, line 146: Bound recursive text parsing if it remains an external-input parser

Even after removing the string from the protobuf contract, LogicalTypeFormat.parse() may still parse generic declarations or external input.

It currently recurses once per nesting level with no cap.

A malformed value such as many thousands of nested arrays can cause:
```
StackOverflowError
```
instead of a normal validation error. Existing callers catch IllegalArgumentException, not StackOverflowError.

Fix:
Thread a nesting-depth counter through parseTree().
Reject depth > a reasonable limit, for example 64, with IllegalArgumentException.

Test:
A deeply nested ARRAY<ARRAY<...>> fails cleanly with IllegalArgumentException.

core/proto/src/main/proto/floecat/query/schema.proto
mrouvroy-floe: I would prefer if we didn't make logical_type_full a string

Current design:
```protobuf
SchemaColumn.logical_type      = "ARRAY"          // legacy flat tag
SchemaColumn.logical_type_full = "ARRAY<INT>"     // new string
```
Problem: the in-memory LogicalType object now records nested nullability, but the string grammar does not.

For example:
```
Source type: ARRAY<INT NOT NULL>
LogicalType: array(INT, elementNullable = false)

Current wire string: ARRAY<INT>
Parsed again: array(INT, elementNullable = true)
```
The same loss occurs for:
```
MAP<STRING, INT NOT NULL>
STRUCT<sku: STRING NOT NULL>
```
This defeats the goal of the PR: the connector correctly discovers the information, then the wire format silently discards it.

Fix this structurally, not by inventing more string syntax. Keep the old flat string for legacy consumers, but add a typed recursive protobuf field:
```
message LogicalType {
  Kind kind = 1;
  optional int32 precision = 2;
  optional int32 scale = 3;
  optional int32 temporal_precision = 4;

  oneof shape {
    ArrayShape array = 20;
    MapShape map = 21;
    StructShape struct = 22;
  }
}

message ArrayShape {
  LogicalType element = 1;
  bool element_nullable = 2;
}

message MapShape {
  LogicalType key = 1;
  LogicalType value = 2;
  bool value_nullable = 3;
}

message StructShape {
  repeated LogicalField fields = 1;
}

message LogicalField {
  string name = 1;
  LogicalType type = 2;
  bool nullable = 3;
}
```
Then SchemaColumn should look conceptually like:
```
string logical_type = 2;                 // existing legacy flat value
optional LogicalType logical_type_tree = 10;  // new lossless value
```

Keep LogicalTypeFormat for textual inputs, docs, debugging, and generic JSON declarations. 
Do not use it as the source-of-truth planner wire format.

Tests to add:
```
LogicalType → protobuf → LogicalType round trips for:
- ARRAY<INT NOT NULL>
- MAP<STRING, DECIMAL(12,2) NOT NULL>
- STRUCT<sku: STRING NOT NULL, quantities: ARRAY<INT>>
- an explicitly known empty STRUCT
```

mrouvroy-floe: I would not keep the old flat string and a new typed tree as permanent parallel representations.

That gives us two sources of truth:
```
logical_type = "ARRAY"
logical_type_tree = ARRAY<INT NOT NULL>
```
Sooner or later one writer updates one and not the other. We already see this risk with logical_type_full: its string form loses nullability even though the in-memory tree has it.

I recommend a clean, versioned breaking migration:
```
message SchemaColumn {
  reserved 2;
  reserved "logical_type";

  // The only authoritative semantic type for every column.
  LogicalType type = 10;
}
```

So: 
1. Declare this as a protocol/schema breaking change.
2. Port all in-repository writers and readers atomically.
3. Rebuild/reconcile persisted schema descriptors from upstream metadata,
   or provide a one-time transparent migration step in-code while the old field is still readable
   on staging / prod envs

mrouvroy-floe (changes requested):
### 1. Complex view output columns are dropped during reconciliation

leaf currently means roughly “this is a scalar leaf eligible for leaf statistics.” That is reasonable.
But view reconciliation uses leaf to decide whether a schema column is a view output. Complex top-level columns are leaf=false, so they disappear.

Example:
```
View:
  SELECT tags  -- tags is ARRAY<STRING>

Mapped schema:
  tags, type=ARRAY, leaf=false

Reconciliation:
  filters leaf=true

Result:
  tags is absent from the registered view schema.
```
A view whose only output is an ARRAY can therefore appear to have no output columns. STRUCT outputs can be replaced by flattened child leaves instead of retaining the top-level struct.

Fix:
Do not use leaf/stat eligibility as the criterion for view output membership.
Preserve top-level ARRAY/MAP/STRUCT columns in the view schema. Keep leaf filtering only in the stats path.
When copying reconciled columns, preserve the typed SchemaColumn.type as well as name, nullability, ordinal, field ID, and relevant physical path.

Tests:
- view with one ARRAY<STRING> output;
- view with one STRUCT<a: INT, b: STRING> output;
- mixed scalar + ARRAY + STRUCT outputs;
- generic-format view with ARRAY<STRING>.

### 2. Generic container columns have the same view-drop problem

GenericSchemaMapper now marks ARRAY/MAP/STRUCT as leaf=false. That is correct for min/max stats, but not if the reconciler interprets leaf=false as “not a view column.”

This is a second manifestation of the same overloaded-flag problem.

Fix:
Separate “visible output column” from “stats leaf”.
Do not solve the view issue by reverting the correct leaf=false stats behaviour.

### 3. Schema paths and stats paths use different recursive walks

There are two nested-type traversals:
- one creates SchemaColumn paths;
- one creates field-ID / stats paths.

They do not recurse identically.
Example:
```
Source: LIST<LIST<INT>>

Stats path walk can produce:
  matrix[]
  matrix[][]

Schema path walk currently may only emit:
  matrix

The stats then refer to a field/path with no matching schema column.
```
Fix:
Extract one shared recursive nested-path traversal.
Use it for both schema construction and field-ID/stats construction.

Tests:
- LIST<LIST<INT>>
- MAP<STRING, LIST<STRUCT<x: INT>>>
- ARRAY<STRUCT<a: ARRAY<INT>>>

Each test should compare the complete emitted schema-path set with the complete stats-path set.

This is also a DRY issue: two independent recursive implementations for the same canonical-path contract will drift again.


### 4. Fix or remove IcebergTypeMappings.toCanonical()

Its public name and documentation imply a flat canonical type string. It now returns parameterised strings for complex types because it calls the full formatter.

Before:
```
toCanonical(list<string>) == "ARRAY"
```
After:
```
toCanonical(list<string>) == "ARRAY<STRING>"
```
Fix one of:

- delete it if it is unused;
- change it to return LogicalTypeFormat.formatTag(toLogical(type)).

Do not keep a public helper whose name/documentation says one thing and whose output changed meaning.

core/proto/src/main/proto/floecat/query/schema.proto
jincongho-floe: Reserved the old field and unify all reader/writer to the new field.

core/types/src/main/java/ai/floedb/floecat/types/LogicalTypeFormat.java
jincongho-floe: Set the recurse limit to 64 depth level.

core/proto/src/main/proto/floecat/query/schema.proto
jincongho-floe: The new type will be proper hierarchical type now

core/proto/src/main/proto/floecat/types/types.proto
```diff
@@ -59,7 +59,40 @@ message LogicalType {
   optional int32 interval_leading_precision = 6;
   // Fractional seconds precision for INTERVAL (applies when range includes SECOND).
   optional int32 interval_fractional_precision = 7;
-  map<string,string> params = 99;
+
+  // Recursive shape for complex kinds. Set for TK_ARRAY/TK_MAP/TK_STRUCT when the nested
+  // structure is known; a complex kind without a shape is the legacy non-parameterised
+  // container tag. A StructShape with zero fields is an explicitly known empty struct
+  // (distinct from an unset shape). Never set for scalar kinds or TK_VARIANT.
+  oneof shape {
```
mrouvroy-floe, line 67: ### Nullability booleans default to the unsafe direction
- line 79: bool element_nullable = 2;
- line 85: bool value_nullable = 3;
- line 95: bool nullable = 3; (on LogicalField)

A plain (non-optional) proto3 bool has no presence tracking. There is no way to distinguish "the producer set this to false" from "the producer never set it". Both decode as false.

So if any producer — a new connector, a test fixture, a hand-built message, a future service — forgets to set these, the reader sees nullable = false, i.e. "this field is NOT NULL". That's the dangerous direction: a downstream consumer that trusts non-nullability can legitimately skip null checks, drop null-handling branches, or simplify a predicate. A wrong true costs a little performance; a wrong false produces wrong answers.

Two ways to fix, either is fine:

- Invert the polarity to element_required / value_required / required. Absent then means "nullable", which is the safe default. This also matches the upstream vocabulary you're already reading from — Iceberg spells it element-required / value-required, and this PR's own mapper reads exactly those keys.
- Mark them optional bool so absence is detectable and the decoder can decide.
I'd prefer the first: it removes the failure mode instead of making it detectable.

Note MapShape correctly has no key_nullable — map keys are always required, and the Java constructor documents it.

core/types/src/main/java/ai/floedb/floecat/types/LogicalType.java
```diff
@@ -61,6 +73,14 @@ public final class LogicalType {
   public final Integer intervalLeadingPrecision;
   public final Integer intervalFractionalPrecision;
 
+  // Nested type tree (null on non-parameterised complex types and all scalar kinds).
+  public final LogicalType element;
+  public final Boolean elementNullable;
```
mrouvroy-floe: ### Boolean elementNullable is a three-state where two would do. 

The constructor already enforces that element and elementNullable are set together (line ~113), so element != null already tells you whether nullability is known — the extra null state of the Boolean is unreachable but still forces every caller to think about it. 
toProto then calls t.elementNullable() and auto-unboxes, which is an NPE if that invariant ever slips. 
A primitive boolean plus the existing invariant is simpler and can't NPE. Same for valueNullable.

core/types/src/main/java/ai/floedb/floecat/types/LogicalTypeFormat.java
```diff
@@ -83,15 +104,75 @@ public static String format(LogicalType t) {
...
+    }
+    return parseScalar(s);
+  }
+
+  /**
+   * Maximum container nesting depth accepted by {@link #parse}. Parsing recurses once per level;
+   * without a cap a malformed deeply nested input would overflow the stack instead of failing with
+   * the {@link IllegalArgumentException} callers handle.
+   */
+  static final int MAX_NESTING_DEPTH = 64;
```
mrouvroy-floe: ### Duplicate nesting-depth constant. 

MAX_NESTING_DEPTH = 64 is defined in LogicalTypeFormat.java:169 and independently redefined in ViewMetadataService.java:398. Two constants that must agree, in different modules, with no test tying them together. Share one.
Not sure if we want to make this actually configurable for strange datasets. Hardcoding seems fine for now

core/types/src/main/java/ai/floedb/floecat/types/LogicalTypeProtoAdapter.java
```diff
@@ -132,6 +132,218 @@ public static UpstreamStamp upstreamStamp(
...
+   * typed migration. The legacy {@code logical_type} string (reserved field 2) survives in the
+   * message's unknown fields; when the column has no typed field, parse the legacy value into one
+   * so pre-migration rows — notably output columns of views created directly via gRPC, which have
+   * no upstream to re-reconcile from — stay usable. Complex legacy values were flat container tags,
+   * so they upgrade to the legacy non-parameterised form (no nested tree).
+   *
+   * <p>Returns the column unchanged when it already has a type, has no legacy value, or the legacy
+   * value does not parse (readers then degrade as for any untyped column).
+   */
+  public static ai.floedb.floecat.query.rpc.SchemaColumn upgradeLegacyColumn(
```
mrouvroy-floe, line 318: ### The legacy-recovery path is safe, but unobservable and permanent
The mechanism works and is safe but:

a. It's silent. All three failure modes — no typed field, no legacy value, unparseable legacy value — return the column unchanged with no log and no counter. The consequence surfaces much later as MissingColumnLogicalTypeException from the planner's hint resolver at query time, reading "Missing logical type for column x", which gives no indication that the real cause was a failed migration of a pre-upgrade view. 
A warn log and a counter at the recovery site would help.

b. It's permanent. The upgrade runs on read and never on write, so the legacy bytes stay in the blob forever and this code can never be deleted.  A one-time backfill (load + re-save views) could have been better and we could remove this code later on ?

c. The test proves the wrong half. 
LogicalTypeTreeTest:302-331 constructs the unknown field by hand via the legacyColumn(...) helper. That tests the decoder, not that bytes written by the pre-change code actually arrive in that shape. One test that writes field 2 as a known field (a small test-only proto, or raw CodedOutputStream) and then parses those bytes as the new SchemaColumn would close the gap that actually matters.

core/types/src/main/java/ai/floedb/floecat/types/LogicalType.java
```diff
@@ -61,6 +73,14 @@ public final class LogicalType {
   public final Integer intervalLeadingPrecision;
   public final Integer intervalFractionalPrecision;
 
+  // Nested type tree (null on non-parameterised complex types and all scalar kinds).
```
mrouvroy-floe, line 77: ### Make the Java model isomorphic to oneof shape

Not a blocker, but it's the a good time to change this I think.

The implementation has one wide class with six nullable fields and five runtime invariant checks. 
I'd suggest mirroring the proto:
```
sealed interface Shape permits ArrayShape, MapShape, StructShape {}
// LogicalType keeps kind + scalar params, and holds an optional Shape
```
The payoff is that the Java model and the wire model stop being two things kept in agreement by hand:

- The five constructor checks (LogicalType.java:113-130 — "element type is only allowed for ARRAY", "key, value, and valueNullable must be set together") become unrepresentable rather than thrown.

- The Boolean elementNullable tri-state (LogicalType.java:78) disappears. Right now the constructor guarantees element and elementNullable are set together, so the null state is unreachable — but toProto calls t.elementNullable() and auto-unboxes, which is an NPE the moment that invariant slips. Nullability belongs on the shape that owns it.
- toProto stops dispatching on field presence (if element != null … else if key != null … else if fields != null) and becomes an exhaustive switch. 
The current dispatch is correct, but only because of those constructor checks — it reads as if it's inferring the kind from the payload.
- TypeResolver caches on Map<LogicalType, Optional<TypeNode>>, so equals/hashCode are load-bearing. Records give correct deep equality for free; the hand-written class has to get it right for six new fields, forever.

core/connectors/common/src/main/java/ai/floedb/floecat/connector/common/resolver/IcebergTypeMappings.java
```diff
@@ -50,9 +51,22 @@ public static LogicalType toLogical(Type t) {
...
+        Types.MapType map = t.asMapType();
+        yield LogicalType.map(
+            toLogical(map.keyType()), toLogical(map.valueType()), map.isValueOptional());
+      }
+      case STRUCT -> {
+        List<LogicalField> fields =
+            t.asStructType().fields().stream()
+                .map(f -> new LogicalField(f.name(), f.isOptional(), toLogical(f.type())))
+                .toList();
+        yield fields.isEmpty() ? LogicalType.of(LogicalKind.STRUCT) : LogicalType.struct(fields);
```
mrouvroy-floe: [major] **Empty `STRUCT<>` collapsed to unknown-shape legacy tag instead of `LogicalType.struct(List.of())`**

An explicitly empty source `STRUCT` is mapped to `LogicalType.of(LogicalKind.STRUCT)` — the legacy unknown-shape container — rather than `LogicalType.struct(List.of())`. The new model and proto explicitly distinguish these two states, so information is silently lost while appearing to preserve the full recursive type tree.

The same pattern appears in the Delta kernel mapper, Delta JSON fallback, Delta UC mapper, and Iceberg REST view mapper, meaning `STRUCT<>`, `ARRAY<STRUCT<>>`, and map values containing empty structs all degrade to an unknown legacy `STRUCT` shape.

Use `LogicalType.struct(List.of())` wherever the source schema explicitly provides an empty field list. Reserve the bare `LogicalType.of(STRUCT)` form only for genuinely legacy or unknown-shape inputs.

---

mrouvroy-floe (commented):
**[major] `information_schema.columns` polluted with synthetic nested-type rows** — `core/catalog/src/main/java/ai/floedb/floecat/systemcatalog/informationschema/ColumnsScanner.java:241`

`ColumnsScanner` iterates `ctx.graph().tableSchema()` with no filter, so every synthetic `SchemaColumn` emitted for nested nodes — list element (`arr[]`), map key (`m.key`), map value (`m{}`) — becomes a row in `information_schema.columns`. Before this PR, a table with `tags array<int>` produced one `tags` row; it now also produces a bogus `element` row of type `INT`.

This pollutes the standard `information_schema.columns` view and `ConstraintScanIndex.columnsById`, which reads the same list. Downstream tooling that relies on the columns view to discover user-defined columns will see spurious entries.

Filter out synthetic nested-field nodes (those whose `SchemaColumn` origin is an element, key, or value placeholder) before emitting rows in `ColumnsScanner`. The existing `ColumnsScannerTest` cases only use scalar columns and will not catch regressions here — add tests covering `ARRAY`, `MAP`, and nested `STRUCT` column types.

**[major] View create/update accepts untyped `SchemaColumn` output columns** — `service/src/main/java/ai/floedb/floecat/service/catalog/impl/ViewServiceImpl.java:180`

The view create path only checks that `output_columns` is non-empty, and the update path applies the same shallow check before persisting. With `SchemaColumn.type` now the sole authoritative type field, a direct gRPC caller can still store `SchemaColumn.newBuilder().setName("c").build()` — a column with no type — as a view output.

This leaves consumers with an untyped schema column: `ArrowSchemaUtil` throws on `!column.hasType()`, while catalog surfaces emit an empty type string. The migration helper (`upgradeLegacyColumn`) only runs when loading persisted views and only helps when the old `logical_type` field is present in unknown fields; it does not guard against new malformed writes.

Validate every supplied output column on create and update by requiring a decodable `type` field, optionally after calling `LogicalTypeProtoAdapter.upgradeLegacyColumn` to preserve backward compatibility with clients still sending the old field.

---

core/catalog/src/main/java/ai/floedb/floecat/systemcatalog/util/SchemaColumns.java
```diff
@@ -0,0 +1,59 @@
...
+  /**
+   * Drops the synthetic placeholder rows the nested-type traversal emits for container children
+   * that have no user-facing identity of their own: list elements (path {@code parent[]}), map
+   * values ({@code parent{}}), and map keys ({@code parent.key} where {@code parent} is a MAP-typed
+   * row — a real struct field named {@code key} keeps its row). These placeholders exist so schema
+   * paths and stats paths cover the same node set; user-facing catalog surfaces (e.g. {@code
+   * information_schema.columns}) must not present them as columns. Struct-field rows ({@code
+   * parent.child}) are real named fields and are kept.
+   */
+  public static List<SchemaColumn> withoutSyntheticNodes(List<SchemaColumn> columns) {
```
mrouvroy-floe, line 42: 1. A top-level row is never synthetic

GenericSchemaMapper.java:93,97 sets setName(name) and setPhysicalPath(name) from the same string, and the only name validation is a non-blank check (lines 70-73). 
A column named items[] therefore has physical_path == name == "items[]" and is silently dropped from information_schema.columns, DescribeInputs, and user-object bundles. 
Brackets are legal column-name characters in Iceberg, Delta, and Parquet.

I would replace: 
```java
  private static boolean isSyntheticContainerNode(SchemaColumn column, Set<String> mapPaths) {
    String path = column.getPhysicalPath();
    if (path.endsWith("[]") || path.endsWith("{}")) {
      return true;
    }
```
With
```java
  private static boolean isSyntheticContainerNode(SchemaColumn column, Set<String> mapPaths) {
    String path = column.getPhysicalPath();
    // Synthetic rows are always nested: the traversal derives their path from a parent. A
    // top-level row's path is its own name, so a user column that happens to be spelled
    // "items[]" or "m{}" is a real column, not a placeholder.
    if (path.equals(column.getName())) {
      return false;
    }
    if (path.endsWith("[]") || path.endsWith("{}")) {
      return true;
    }
```

2. Update the javadoc:
Append to the existing paragraph:
```
   * parent.child}) are real named fields and are kept, as is any top-level row whose path is its
   * own name.
 ```

File: core/catalog/src/main/java/ai/floedb/floecat/systemcatalog/util/SchemaColumns.java
Comment (mrouvroy-floe) [inline core/catalog/src/main/java/ai/floedb/floecat/systemcatalog/util/SchemaColumns.java:1]: I would add some tests for this class: 
```java
package ai.floedb.floecat.systemcatalog.util;

import static org.assertj.core.api.Assertions.assertThat;

import ai.floedb.floecat.query.rpc.SchemaColumn;
import ai.floedb.floecat.types.rpc.LogicalType;
import java.util.List;
import org.junit.jupiter.api.Test;

class SchemaColumnsTest {

  private static SchemaColumn col(String name, String path, LogicalType.Kind kind) {
    return SchemaColumn.newBuilder()
        .setName(name)
        .setPhysicalPath(path)
        .setType(LogicalType.newBuilder().setKind(kind))
        .build();
  }

  @Test
  void keepsTopLevelColumnsWhoseNamesLookLikeContainerPaths() {
    var columns =
        List.of(
            col("items[]", "items[]", LogicalType.Kind.TK_INT),
            col("m{}", "m{}", LogicalType.Kind.TK_STRING),
            col("plain", "plain", LogicalType.Kind.TK_INT));

    assertThat(SchemaColumns.withoutSyntheticNodes(columns))
        .extracting(SchemaColumn::getName)
        .containsExactly("items[]", "m{}", "plain");
  }

  @Test
  void dropsSyntheticNestedPlaceholdersButKeepsStructFields() {
    var columns =
        List.of(
            col("items", "items", LogicalType.Kind.TK_ARRAY),
            col("element", "items[]", LogicalType.Kind.TK_STRUCT),
            col("sku", "items[].sku", LogicalType.Kind.TK_STRING),
            col("m", "m", LogicalType.Kind.TK_MAP),
            col("key", "m.key", LogicalType.Kind.TK_STRING),
            col("value", "m{}", LogicalType.Kind.TK_INT));

    assertThat(SchemaColumns.withoutSyntheticNodes(columns))
        .extracting(SchemaColumn::getPhysicalPath)
        .containsExactly("items", "items[].sku", "m");
  }

  @Test
  void keepsStructFieldNamedKeyWhoseParentIsNotAMap() {
    var columns =
        List.of(
            col("s", "s", LogicalType.Kind.TK_STRUCT),
            col("key", "s.key", LogicalType.Kind.TK_STRING));

    assertThat(SchemaColumns.withoutSyntheticNodes(columns))
        .extracting(SchemaColumn::getPhysicalPath)
        .containsExactly("s", "s.key");
  }
}
```

service/src/main/java/ai/floedb/floecat/service/catalog/impl/ViewServiceImpl.java
```diff
@@ -577,6 +579,33 @@ private static FieldMask normalizeMask(FieldMask mask) {
...
+      } catch (IllegalArgumentException e) {
+        throw GrpcErrors.invalidArgument(
+            corr, VIEW_OUTPUT_COLUMN_TYPE_REQUIRED, Map.of("column", column.getName()));
+      }
+      typed.add(upgraded);
+    }
+    return typed;
+  }
+
   private static byte[] canonicalFingerprint(ViewSpec s) {
```
mrouvroy-floe, line 610: ### canonicalFingerprint — upgrade inside the method, so both call sites are covered

The fingerprint is computed from untyped columns, and there are two call sites that must agree — line 193 fingerprints the request, line 264 fingerprints specFromView(existing), whose stored columns are typed (they were upgraded before storage). Two consequences, both real:

With an explicit idempotency key, a legacy request's columns all render as the empty type token, so [a:INT] and [a:STRING] fingerprint identically and a replay after a type change returns the old result.
Without one, the untyped request fingerprint can never equal the typed stored fingerprint, so an identical legacy replay is reported as a conflict instead of a match.
Fixing it inside the method covers both sides and can't be forgotten by a third caller.

Replace
```java
        .list(
            "output_columns",
            s.getOutputColumnsList().stream()
                .map(
                    c ->
                        c.getName()
```
With
```java
        // Both callers must compare like with like: the request spec may still carry legacy
        // untyped columns while the stored spec is already typed, and an untyped column renders
        // as the empty type token — making distinct types fingerprint identically. Upgrade here
        // rather than at the call sites so a future caller cannot skip it.
        .list(
            "output_columns",
            s.getOutputColumnsList().stream()
                .map(ai.floedb.floecat.types.LogicalTypeProtoAdapter::upgradeLegacyColumn)
                .map(
                    c ->
                        c.getName()
```

### Known residual limit — add to the same comment
columnTypeString goes through LogicalTypeFormat.format, which renders ARRAY<INT> with no nullability, so two views differing only in element/field nullability still fingerprint the same. Column-level c.getNullable() is already in the token; it's the nested bits that are lost. Worth one sentence so it's a recorded limit rather than a latent surprise:
```
        // Nested nullability is not part of the token (LogicalTypeFormat.format omits it), so two
        // specs differing only in element/field nullability compare equal. Tracked separately.
```

### Test
Extend ViewMutationIT (which already pins the idempotent-replay and conflict-on-different-columns behavior): create a view whose output columns carry only the legacy field-2 value, then replay with the same idempotency key but a different legacy type, and assert a conflict rather than a replay.

core/types/src/main/java/ai/floedb/floecat/types/Shape.java
```diff
@@ -0,0 +1,59 @@
...
+ * limitations under the License.
+ */
+
+package ai.floedb.floecat.types;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Nested structure of a complex {@link LogicalType}, mirroring the wire model's {@code oneof shape}
```
mrouvroy-floe, line 23: The mirroring claim is only true structurally

Replace
```java
 * Nested structure of a complex {@link LogicalType}, mirroring the wire model's {@code oneof shape}
 * so the Java and proto representations cannot drift: which children a container carries is fixed
 * by its shape variant, nullability lives on the shape that owns it (primitive, no tri-state), and
 * adapters dispatch exhaustively instead of probing field presence.
 * 
```
With
```java
 * Nested structure of a complex {@link LogicalType}, mirroring the wire model's {@code oneof shape}
 * so the two cannot drift structurally: which children a container carries is fixed by its shape
 * variant, nullability lives on the shape that owns it (primitive, no tri-state), and adapters
 * dispatch exhaustively instead of probing field presence.
 *
 * <p>Polarity is deliberately <em>not</em> mirrored: this model says {@code nullable}, the wire says
 * {@code required}, so that an unset proto3 bool defaults to nullable — the direction that can only
 * cost performance, never correctness. The negation lives in {@link LogicalTypeProtoAdapter} and
 * nowhere else.
 * 
 ```

File: docs/types.md
Comment (mrouvroy-floe) [inline docs/types.md:1]: ### Three sections name a removed field and two deleted methods

Line 109 already states correctly that logical_type is reserved, so the document currently contradicts itself. I grepped for both cited methods: neither IcebergSchemaMapper.toCanonical nor DeltaSchemaMapper.deltaTypeToCanonical exists on this branch. The current entry points are IcebergTypeMappings.toLogical(Type) (IcebergTypeMappings.java:32), DeltaSchemaMapper.toLogicalType(DataType) (kernel, :179), DeltaSchemaMapper.fallbackLogicalType(JsonNode) (JSON fallback, :353), and DeltaTypeMapper.toLogical(DataType) (:88).

#### 4a. Line 188 — Arrow contract input
Replace:
```
- Input is `SchemaColumn.logical_type` and should be a Floecat canonical logical type string (or an
  accepted alias handled by `LogicalKind.fromName` semantics).
```
With:
```
- Input is `SchemaColumn.type`, a recursive `floecat.types.LogicalType` message; decode it with
  `LogicalTypeProtoAdapter.columnType`. Textual entry points (generic schema declarations) still
  accept canonical names and aliases via `LogicalKind.fromName` semantics.
```

#### 4b. Line 253-254 — data-flow diagram
Replace:
```
  → schema mapper emits canonical LogicalKind strings (e.g. "INT", "TIMESTAMPTZ", "ARRAY")
  → SchemaColumn.logical_type stores the canonical string
```
With:
```
  → schema mapper emits recursive LogicalType trees (e.g. INT, TIMESTAMPTZ, ARRAY<STRUCT<...>>)
  → SchemaColumn.type stores the typed tree; nested nodes also surface as child rows by path
```

#### 4c. Lines 271-279 — connector examples
Replace:
```
- **Iceberg schema parsing** – `IcebergSchemaMapper.toCanonical(Type)` converts Iceberg types to
  canonical strings (e.g. `TimestampType.withZone()` → `"TIMESTAMPTZ"`), storing them in
  `SchemaColumn.logical_type`. This avoids the historic ambiguity where `"timestamp"` meant
```
With:
```
- **Iceberg schema parsing** – `IcebergTypeMappings.toLogical(Type)` converts Iceberg types to
  `LogicalType`, recursing into LIST/MAP/STRUCT and preserving element/key/value types and
  nullability; `IcebergSchemaMapper` stores the result in `SchemaColumn.type`. This avoids the
  historic ambiguity where `"timestamp"` meant
```
And replace:
```
- **Delta schema parsing** – `DeltaSchemaMapper.deltaTypeToCanonical(JsonNode)` applies Delta-
  specific semantics: `"timestamp"` → `"TIMESTAMPTZ"` (UTC-stored), `"timestamp_ntz"` → `"TIMESTAMP"`
  (timezone-naive).
```
With:
```
- **Delta schema parsing** – `DeltaSchemaMapper.toLogicalType(DataType)` (kernel path) and
  `fallbackLogicalType(JsonNode)` (JSON fallback) apply Delta-specific semantics: `"timestamp"` →
  `TIMESTAMPTZ` (UTC-stored), `"timestamp_ntz"` → `TIMESTAMP` (timezone-naive). Both recurse into
  arrays/maps/structs, preserving `containsNull` / `valueContainsNull`.
```
#### 4d. Optional, one sentence worth adding
Near the wire description around line 106, so the tree/string split doesn't read as an oversight:
```
`ScalarStats.logical_type` remains a flat canonical string: stats rows exist only for leaves, and
leaf types are never parameterised containers.
```

**Comment by jincongho-floe** [comment]
Both fixed in 39f3c80cb (the review was against 6c3d02e06, one commit earlier):

- **Nested fields named like container syntax** (`s.items[]`): synthetic-node classification now requires a typed parent, not just a path suffix — a `[]`/`{}` suffix only marks a placeholder if an ARRAY/MAP-typed row exists at the would-be parent path. For `STRUCT<\"items[]\": INT>` there is no ARRAY row at `s.items`, so the field keeps its row. Composes with the existing top-level `path == name` guard. Pinned by `SchemaColumnsTest.keepsNestedFieldsWhoseNamesLookLikeContainerSyntax`.
- **Struct children in relation bundles**: `UserObjectBundleService` now builds relation payloads via `SchemaColumns.topLevelOnly` — no nested rows of any kind reach the engine, so per-parent ordinals/attnums are dense and collision-free. Nested typing travels in the per-column type tree. Pinned by `SchemaColumnsTest.topLevelOnlyDropsAllNestedRowsIncludingStructChildren`.

One deliberate exception to confirm: `DescribeInputs` still includes struct children (filtered of placeholders only). It's a logical schema listing with qualified names and no attnum semantics, and dropping struct children there would change pre-migration client-visible behavior — happy to make it strictly top-level too if you'd rather have one rule everywhere.

File: core/proto/src/main/proto/floecat/types/types.proto (hunk 3 above)
Comment (jincongho-floe) [reply core/proto/src/main/proto/floecat/types/types.proto:67]: Fixed in 2b261f5dd, taking your preferred option: the wire encodes `element_required` / `value_required` / `LogicalField.required`, so an unset proto3 bool decodes as nullable — the direction that can only cost performance. The Java model keeps `nullable` vocabulary; the negation lives only in `LogicalTypeProtoAdapter`, and `Shape`'s javadoc documents the deliberate polarity split. `wireEncodesRequiredSoUnsetDefaultsToNullable` pins both directions plus the unset-defaults-to-nullable case. (MapShape stays key-nullability-free.)

File: core/types/src/main/java/ai/floedb/floecat/types/LogicalType.java (hunk 4 above)
Comment (jincongho-floe) [reply core/types/src/main/java/ai/floedb/floecat/types/LogicalType.java]: Gone in 2b261f5dd via the sealed-`Shape` refactor you suggested in the other thread: nullability is now a primitive `boolean` on the shape record that owns it, so the unreachable null state and the auto-unboxing NPE hazard no longer exist.

File: core/types/src/main/java/ai/floedb/floecat/types/LogicalTypeFormat.java (hunk 5 above)
Comment (jincongho-floe) [reply core/types/src/main/java/ai/floedb/floecat/types/LogicalTypeFormat.java]: Fixed in 2b261f5dd: `LogicalTypeFormat.MAX_NESTING_DEPTH` is now public and is the single constant — the proto adapter and the gateway's write-time validation both reference it; the gateway's own `= 64` is deleted. Kept hardcoded per your note.

File: core/types/src/main/java/ai/floedb/floecat/types/LogicalTypeProtoAdapter.java (hunk 6 above)
Comment (jincongho-floe) [reply core/types/src/main/java/ai/floedb/floecat/types/LogicalTypeProtoAdapter.java:318]: (a) and (c) fixed in 2b261f5dd. Observable: recovery now warn-logs unrecoverable columns (view id + column + consequence) and counts both outcomes via `floecat.view.legacy_column_upgrade{outcome=recovered|unrecovered}`. Test: `upgradeLegacyColumnRecoversBytesWrittenByPreChangeSchema` writes field 2 with raw `CodedOutputStream` exactly as the pre-change proto did and parses those bytes as the new `SchemaColumn` — proving the real byte shape, not the hand-built fixture. (b) is partially mitigated rather than solved: create/update now upgrade-and-persist typed columns (so legacy bytes stop accumulating and old clients' writes come out typed), but dormant never-rewritten views keep their legacy bytes and the shim can't yet be deleted. I'd take the one-time backfill (load + re-save views) as a follow-up PR — it wants its own rollout consideration. Agree?

File: core/types/src/main/java/ai/floedb/floecat/types/LogicalType.java (hunk 7 above)
Comment (jincongho-floe) [reply core/types/src/main/java/ai/floedb/floecat/types/LogicalType.java:77]: Taken, in 2b261f5dd: `sealed interface Shape permits Array, Map, Struct` (records), `LogicalType` holds one shape, the five constructor invariants are now unrepresentable (one kind/shape consistency check remains), `toProto` is an exhaustive shape switch, and equals/hashCode collapse to `Objects.equals(shape, ...)` — records carry the deep equality the `TypeResolver` cache needs. Convenience accessors kept as projections so mappers/floecat-runtime needed no changes.

File: core/connectors/common/src/main/java/ai/floedb/floecat/connector/common/resolver/IcebergTypeMappings.java (hunk 8 above)
Comment (jincongho-floe) [reply core/connectors/common/src/main/java/ai/floedb/floecat/connector/common/resolver/IcebergTypeMappings.java]: Fixed in 2b261f5dd across all five mappers (IcebergTypeMappings, Delta kernel + JSON fallback, Delta UC, gateway): an explicitly empty source field list now maps to `LogicalType.struct(List.of())`; the bare legacy tag remains only when the source provides no field list at all (Delta/gateway JSON without a `fields` key). These were stale guards from before known-empty structs became first-class — thanks for catching that the model change never propagated to the producers.

File: core/catalog/src/main/java/ai/floedb/floecat/systemcatalog/util/SchemaColumns.java (hunk 9 above)
Comment (jincongho-floe) [reply core/catalog/src/main/java/ai/floedb/floecat/systemcatalog/util/SchemaColumns.java:42]: Applied as suggested in 71c3ad933 (guard + javadoc verbatim), and extended in 39f3c80cb for the nested variant of the same trap: a NESTED field literally named `items[]` (path `s.items[]`) passes the top-level guard's scope, so `[]`/`{}` suffixes now classify by typed parent — synthetic only if an ARRAY/MAP-typed row exists at the would-be parent path. Both compose with the existing `.key` map-parent check.

File: core/catalog/src/main/java/ai/floedb/floecat/systemcatalog/util/SchemaColumns.java
Comment (jincongho-floe) [reply core/catalog/src/main/java/ai/floedb/floecat/systemcatalog/util/SchemaColumns.java:1]: Added verbatim in 71c3ad933, plus two more cases in 39f3c80cb: `keepsNestedFieldsWhoseNamesLookLikeContainerSyntax` (the `s.items[]` nested variant) and `topLevelOnlyDropsAllNestedRowsIncludingStructChildren` (for the new bundle filter).

File: service/src/main/java/ai/floedb/floecat/service/catalog/impl/ViewServiceImpl.java (hunk 10 above)
Comment (jincongho-floe) [reply service/src/main/java/ai/floedb/floecat/service/catalog/impl/ViewServiceImpl.java:610]: Fixed in 4075acde7: the upgrade happens inside `canonicalFingerprint` so both call sites (and any future one) compare like with like. The token also went further than `columnTypeString`: it Base64s the serialized typed tree, which carries nested requiredness — so the residual limit you flagged (two specs differing only in element/field nullability fingerprinting the same) no longer applies; the comment in the method notes this.

File: core/types/src/main/java/ai/floedb/floecat/types/Shape.java (hunk 11 above)
Comment (jincongho-floe) [reply core/types/src/main/java/ai/floedb/floecat/types/Shape.java:23]: Applied verbatim in 6c3d02e06 — the javadoc now scopes the mirroring claim to structure and documents the deliberate polarity inversion with the negation confined to `LogicalTypeProtoAdapter`.

File: docs/types.md
Comment (jincongho-floe) [reply docs/types.md:1]: Fixed in 6c3d02e06: all three sections updated to the current entry points (`IcebergTypeMappings.toLogical`, `DeltaSchemaMapper.toLogicalType`/`fallbackLogicalType`, `DeltaTypeMapper.toLogical`) and to `SchemaColumn.type` + `LogicalTypeProtoAdapter.columnType`. Remaining `logical_type` mentions are the intentionally-correct ones: the reserved-field note and `ScalarStats.logical_type`, which really does stay a flat string.

File: service/src/main/java/ai/floedb/floecat/service/catalog/impl/ViewServiceImpl.java [hunk 12]
```diff
@@ -587,6 +617,24 @@ private static byte[] canonicalFingerprint(ViewSpec s) {
...
+        // legacy type field while stored views are upgraded on write. Normalizing here keeps
+        // idempotent replays stable and prevents distinct legacy types from sharing a fingerprint.
+        // The serialized typed tree also includes nested requiredness.
+        .list(
+            "output_columns",
+            s.getOutputColumnsList().stream()
+                .map(ai.floedb.floecat.types.LogicalTypeProtoAdapter::upgradeLegacyColumn)
+                .map(
+                    c ->
+                        c.getName()
```
mrouvroy-floe, line 630: [major] **View fingerprints omit schema semantics**

The new `output_columns` fingerprint token includes only `name`, `type`, and `nullable`, although `CreateView` persists the rest of each `SchemaColumn`. In particular, `ordinal` is planner-visible through `ColumnInfo`, while `field_id`, `physical_path`, `partition_key`, and `leaf` are also stored but absent from the fingerprint.

Reusing an idempotency key with the same name/type/nullability but a changed ordinal therefore produces the same fingerprint, so the idempotency guard can replay the first view instead of rejecting the conflicting request. The caller receives success while the requested schema was not applied.

Fingerprint every persisted semantic field (or serialize a normalized `SchemaColumn` after clearing only explicitly server-owned fields). Add a test that reuses an idempotency key with a changed `ordinal` or `field_id` and expects a conflict.

---

mrouvroy-floe: This is not really related to ARRAY / STRUCT work so feel free to just create an issue and assign it to me. 
I can fix this up later.

jincongho-floe: Created: https://github.com/eng-floe/floecat/issues/422

mrouvroy-floe (approved):

PR #424, opened 2026-07-30
feat(integration): add catalog integration API (stack 5/8)

mcusack-floe:
## Summary

Adds the catalog integration resource and gRPC API, including persistence, idempotent CRUD behavior, authorization, credential storage and cleanup, garbage-collection hooks, account-deletion participation, and service/storage documentation.

Catalog overlays and CLI commands are intentionally excluded into subsequent PRs.

## Type of Change

- [ ] fix (bug fix)

## Testing

- Focused catalog integration API/service suite: 117 tests passed.

- [x] `make fmt`
- [ ] `make verify`

- [x] No breaking changes

## Checklist

## Additional Notes

Stack position: **5/8**. Base: `mc-account-deletion-fence`. Review the integration resource independently of overlays, CLI, and connector-independent access.


service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
```diff
@@ -0,0 +1,501 @@
...
+              var meta = overlays.metaFor(id);
+              MutationOps.BaseServiceChecks.enforcePreconditions(
+                  corr, meta, request.getPrecondition());
+              if (row.getOwnsCatalog()
+                  && !markerStore.advanceCatalogMarker(row.getCatalogId(), catalogMarkerVersion)) {
+                throw GrpcErrors.preconditionFailed(
+                    corr, null, Map.of("reason", "catalog children changed"));
+              }
+              if (row.getOwnsCatalog()) {
+                boolean deleted = catalogs.delete(row.getCatalogId());
```
mrouvroy-floe: [major] **`deleteCatalogOverlay` deletes the owned catalog before the CAS delete of the overlay, leaving state permanently split on failure**

The owned-catalog branch deletes the `Catalog` (and invalidates the metadata graph) at lines 364-371 *before* `overlays.deleteWithPrecondition(id, meta.getPointerVersion())` at line 372. `meta` is read at line 356, so any concurrent write to the overlay in that window advances the pointer version; `deleteWithPrecondition` then returns `false` and the method throws `FAILED_PRECONDITION`.

`runWithRetry` only retries `AbortRetryableException` / `StorageAbortRetryableException`; `GrpcErrors.preconditionFailed` is a plain `StatusRuntimeException`, so the operation is not retried. The caller receives an error, but the catalog is already gone while the overlay record survives with `owns_catalog=true` and a `catalog_id` that no longer resolves — a permanently split durable state.

Fix: transition the overlay to a `COS_DELETING` tombstone state with a CAS *first* (making concurrent updates reject that state), then delete the catalog, then remove the overlay. Any failure after the tombstone CAS is idempotent and can be retried or reconciled; no failure path should leave only one half of the mapping committed.

---

🤖 _Generated with Claude Opus 5 and GPT-5.6 Sol._

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
```diff
@@ -0,0 +1,545 @@
...
+  private static void validatePersistedAuthConfig(
+      ai.floedb.floecat.connector.rpc.AuthConfig auth, String corr) {
+    PersistedSecretPropertyValidator.validateNoSecretKeys(
+        auth.getPropertiesMap(), corr, "auth.properties");
+    PersistedSecretPropertyValidator.validateNoSecretKeys(
+        auth.getHeaderHintsMap(), corr, "auth.header_hints");
+  }
+
+  private String credentialRef(String integrationId) {
+    return integrationId + ":" + randomUuid();
```
mrouvroy-floe: [major] **`credentialRef` contains `:`, which is percent-encoded to `%3A` and rejected by AWS Secrets Manager**

`credentialRef` is built as `integrationId + ":" + randomUuid()`. This string flows through `AuthConfigSecretManager.externalize` → `DefaultCredentialResolver.store` → `SecretsManager.buildSecretKey` → `Keys.encodeSegment`, which percent-encodes everything outside `[A-Za-z0-9-._~]`. The resulting secret name becomes `accounts/{acct}/connectors/{uuid}%3A{uuid}`.

`ProdSecretsManager` passes that name unchanged to `CreateSecretRequest.name(...)`. AWS Secrets Manager only permits ASCII letters, digits, and `/_+=.@-` in secret names — `%` is not in that set — so every `CreateCatalogIntegration` request carrying credentials fails with `InvalidParameterException` under the `prod` profile. `NotProdSecretsManager` does not enforce that character set, so the bug is invisible in CI.

Fix: use `-` as the separator (e.g. `integrationId + "-" + randomUuid()`), or sanitize the ref before passing it to the secret-name builder.

---

🤖 _Generated with Claude Opus 5 and GPT-5.6 Sol._

service/src/main/java/ai/floedb/floecat/service/catalog/impl/CatalogServiceImpl.java
```diff
@@ -353,6 +363,11 @@ public Uni<DeleteCatalogResponse> deleteCatalog(DeleteCatalogRequest request) {
                     return DeleteCatalogResponse.newBuilder().setMeta(safe).build();
                   }
 
+                  if (catalogOverlayRepo.getByCatalog(id.getAccountId(), id.getId()).isPresent()) {
```
mrouvroy-floe: [major] **Catalog rename and delete guards race with concurrent `CreateCatalogOverlay`**

The overlay guards added to `UpdateCatalog` and `DeleteCatalog` are one-time reads that are not mutually exclusive with `CreateCatalogOverlay`. Two interleavings break invariants:

1. **Rename race**: `UpdateCatalog` observes no overlay and proceeds to rename; concurrently, `CreateCatalogOverlay` validates against the old name and commits its overlay; then the catalog rename CAS succeeds because the attachment did not change the catalog pointer. The overlay's embedded catalog name now diverges from the catalog record.

2. **Delete race**: `DeleteCatalog` advances the child-count/marker, then an attachment creates an overlay and bumps that marker; the delete does not revalidate the marker or overlay index before committing, so it removes a catalog that has just gained an overlay, orphaning it.

Fix: Coordinate attachment and catalog mutation through the same exclusive lifecycle state (e.g. a `CCS_LOCKED` / `CCS_DELETING` tombstone on the catalog) so that attachment and rename/delete are serialized, or re-validate overlay absence inside the same CAS that commits the rename/delete.

---

🤖 _Generated with Claude Opus 5 and GPT-5.6 Sol._

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
```diff
@@ -0,0 +1,545 @@
...
+              AuthConfigSecretManager.requireExternalized(desired.getAuth());
+              try {
+                if (!integrations.update(desired, meta.getPointerVersion())) {
+                  throw GrpcErrors.preconditionFailed(corr, null, Map.of());
+                }
+              } catch (RuntimeException e) {
+                if (incomingSecret) secrets.delete(pc.getAccountId(), nextCredentialRef);
+                throw e;
+              }
+              if (authTouched && !priorCredentialRef.isBlank()) {
```
mrouvroy-floe: [major] **Credential secret deletion occurs after the integration record is mutated, permanently orphaning secrets on failure**

After an auth update commits the new `credential_ref`, the prior ref exists only in the local stack frame. If the subsequent `CredentialResolver.delete(oldRef)` throws or the process dies, the old secret is permanently unrecoverable: the integration already points to the new ref, no retry can reconstruct the old name (it includes a random suffix), and `CredentialResolver` exposes no listing API.

The same ordering hazard appears in `deleteCatalogIntegration`: the integration record is removed before its secret is deleted, and account-level cleanup follows the same pattern. Any failure leaves live credential payloads stranded in Secrets Manager while the caller has already received an error after the resource mutation committed.

Fix: Persist a durable pending-deletion tombstone (or a two-phase "old ref" field on the integration record) *before* committing the mutation, and drive the actual secret delete idempotently from that tombstone — either in a background reconciler or eagerly on next read. Alternatively, adopt a soft-delete with a TTL so that orphaned secrets are eventually reaped even without an explicit tombstone.

---

🤖 _Generated with Claude Opus 5 and GPT-5.6 Sol._

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
```diff
@@ -0,0 +1,501 @@
...
+              var now = nowTs();
+              var overlayBuilder =
+                  CatalogOverlay.newBuilder()
+                      .setResourceId(overlayId)
+                      .setDisplayName(name)
+                      .setIntegrationId(integrationId)
+                      .setCatalogId(catalogId)
+                      .addAllIncludeNamespaces(includes)
+                      .addAllExcludeNamespaces(excludes)
+                      .setRefreshPolicy(spec.getRefreshPolicy())
```
mrouvroy-floe: [minor] **`refresh_policy` is persisted without validation, allowing semantically invalid `ReconcilePolicy` records**

`createCatalogOverlay` (line 200) and the `refresh_policy` case in `patch` (line 399) copy `spec.getRefreshPolicy()` into the stored overlay without any checks. The legacy connector path validates the same `ReconcilePolicy` message: `ConnectorsImpl.validateReconcilePolicy` rejects `RSS_LATEST_N` when `latest_n <= 0` on create, update, and validate.

Without the equivalent check, an overlay can be stored with `scope=RSS_LATEST_N, latest_n=0`. That invalid record is silently accepted today and only causes a runtime failure in the follow-up PR that drives reconciliation off overlays — at which point the bad record already exists in durable storage.

Fix: extract or reuse `ConnectorsImpl.validateReconcilePolicy`, call it from both the create path and the `refresh_policy` branch of `patch`, and return `INVALID_ARGUMENT` on violation.

---

🤖 _Generated with Claude Opus 5 and GPT-5.6 Sol._

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
```diff
@@ -0,0 +1,545 @@
...
+  private String credentialRef(String integrationId) {
+    return integrationId + ":" + randomUuid();
+  }
+
+  private void restoreIntegrationState(
+      CatalogIntegration current, CatalogIntegrationState priorState, long expectedVersion) {
+    if (priorState == CatalogIntegrationState.CIS_DELETING) {
+      return;
+    }
+    integrations.update(
```
mrouvroy-floe: [minor] **`restoreIntegrationState` silently ignores a failed CAS, leaving integrations permanently stuck in `CIS_DELETING`**

`deleteCatalogIntegration` flips the integration to `CIS_DELETING` before checking for dependent overlays, then calls `restoreIntegrationState` on every abort path. That helper calls `integrations.update(...)` but discards the returned boolean — if the CAS loses (concurrent writer, or the `expectedVersion` is already stale), the rollback silently no-ops.

The integration is then permanently `CIS_DELETING`: `updateCatalogIntegration` rejects it (lines 246-248: "integration is deleting"), and `createCatalogOverlay` rejects it (lines 151-153). Re-creating the integration requires supplying credentials again.

This is reachable on the ordinary "delete an integration that still has overlays" path (returns `CONFLICT`), not only exotic races.

Fix: check the return value of `integrations.update(...)` in `restoreIntegrationState` and, on failure, either retry with a fresh read-modify-write or throw an internal error so the caller is not silently left with broken state.

---

🤖 _Generated with Claude Opus 5 and GPT-5.6 Sol._

core/proto/src/main/proto/floecat/common/common.proto
```diff
@@ -42,6 +42,7 @@ enum ResourceKind {
   RK_AGGREGATE = 14;
   RK_TRANSACTION = 15;
   RK_STORAGE_AUTHORITY = 16;
+  RK_CATALOG_INTEGRATION = 17;
```
mrouvroy-floe, line 45: [minor] **`CatalogOverlay` reuses `RK_OVERLAY = 5` instead of a dedicated `ResourceKind`, colliding with view-related CLI code**

This PR introduces `RK_CATALOG_INTEGRATION = 17` for integrations but has `CatalogOverlaysImpl` reuse `RK_OVERLAY = 5` for overlay resource IDs (`randomResourceId` at line 161, `ensureKind` at line 447). `RK_OVERLAY` is not a free slot: `QueryCliSupport.viewRid` and `ViewCliSupport.resolveViewId` both construct **view** `ResourceId`s with `RK_OVERLAY`.

The kind discriminator is what `ensureKind` and every future kind-based dispatch key off, so two unrelated resource families now share the same value. Once overlay `ResourceId`s exist in durable storage the collision is baked in and migration becomes painful.

Fix: add `RK_CATALOG_OVERLAY = 18` (making the two new resources symmetric with `RK_CATALOG_INTEGRATION = 17`) and use it in `CatalogOverlaysImpl`. If reuse is intentional, document the decision explicitly and explain how the dispatch ambiguity is resolved.

---

🤖 _Generated with Claude Opus 5 and GPT-5.6 Sol._

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
```diff
@@ -0,0 +1,501 @@
...
+  }
+
+  @Override
+  public Uni<CreateCatalogOverlayResponse> createCatalogOverlay(
+      CreateCatalogOverlayRequest request) {
+    return mapFailures(
+        runWithRetry(
+            () -> {
+              var pc = principal.get();
+              authz.require(pc, "catalog.write");
```
mrouvroy-floe: [minor] **Overlay RPCs require only `catalog.*`, letting principals without `connector.manage` bind and block integrations**

All five overlay RPCs are guarded by `catalog.read` / `catalog.write` only, while every integration RPC requires `connector.manage`. Because `createCatalogOverlay` accepts an arbitrary `integration_id`, a principal holding `catalog.write` but not `connector.manage` can attach any integration in the account.

Two concrete consequences today: (1) `GetCatalogOverlay` exposes the bound `integration_id` to anyone with `catalog.read`; (2) the new dependent-overlay check in `deleteCatalogIntegration` means such a principal can indefinitely block deletion of an integration they have no rights over.

Once the follow-up PR wires overlay reconciliation to the upstream connector credentials, this also becomes an indirect grant of those credentials.

Fix: require `connector.manage` (in addition to `catalog.write`) on `createCatalogOverlay` and `deleteCatalogOverlay`. The read-only RPCs (`GetCatalogOverlay`, `ListCatalogOverlays`) can retain `catalog.read`, but consider whether surfacing `integration_id` to `catalog.read`-only callers is intentional.

---

🤖 _Generated with Claude Opus 5 and GPT-5.6 Sol._

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
mcusack-floe: Fixed. Overlay deletion now CAS-transitions the overlay to `COS_DELETING` before touching an owned catalog. Updates reject the tombstone state; catalog deletion is preconditioned; the overlay tombstone is removed last; and lifecycle CAS races raise `AbortRetryableException` so the operation can resume safely. Added a regression test that verifies the tombstone is persisted before catalog deletion.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
mcusack-floe: Fixed. Credential refs now use `integrationId + "-" + UUID`, so the generated Secrets Manager name contains only AWS-supported characters. The create test also asserts that generated refs contain no colon.

service/src/main/java/ai/floedb/floecat/service/catalog/impl/CatalogServiceImpl.java
mcusack-floe: Fixed atomically. `GenericResourceRepository` now supports required-absent pointers in the same `compareAndSetBatch` transaction as update/delete. Catalog rename and delete include the by-catalog overlay pointer as a `CasCheckAbsent`, so a concurrent attachment and catalog mutation cannot both commit. Added a repository regression test covering both rename and delete.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
mcusack-floe: Fixed with durable cleanup state. `CatalogIntegration` now retains superseded refs in `pending_credential_delete_refs` in the same CAS that changes auth. Cleanup is idempotently retried after update and eagerly on subsequent get/list calls, while all internal refs remain cleared from API responses. Integration deletion and account cleanup now delete active and pending secrets before removing the durable integration record. Tests cover a secret-manager outage and secret-before-record ordering.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
mcusack-floe: Fixed. Reconcile policy validation is now shared through `ReconcilePolicyValidator`. Overlay create and refresh-policy update both reject `RSS_LATEST_N` with `latest_n <= 0` as `INVALID_ARGUMENT`. Added an overlay regression test.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
mcusack-floe: Fixed by removing the rollback path. An ordinary dependent-overlay conflict is checked before entering `CIS_DELETING`, so it leaves the integration unchanged. Once the tombstone is committed, later races use retryable aborts and deletion resumes from the durable tombstone instead of attempting a best-effort rollback. Added a test verifying that a dependency conflict never tombstones the integration.

core/proto/src/main/proto/floecat/common/common.proto
mcusack-floe: Fixed. Added `RK_CATALOG_OVERLAY = 18` and switched catalog-overlay creation, validation, repository tests, and account-cleanup tests to the dedicated kind. `RK_OVERLAY = 5` remains limited to the existing view-related usage.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
mcusack-floe: Fixed. Overlay create, update, and delete now require both `catalog.write` and `connector.manage`. Read-only list/get retain `catalog.read`; exposing the account-scoped integration ID there is intentional metadata visibility and does not resolve or expose credential refs or payloads. The service documentation now records the write permission requirements.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
```diff
@@ -0,0 +1,533 @@
...
+  }
+
+  @Override
+  public Uni<CreateCatalogOverlayResponse> createCatalogOverlay(
+      CreateCatalogOverlayRequest request) {
+    return mapFailures(
+        runWithRetry(
+            () -> {
+              var pc = principal.get();
+              authz.require(pc, List.of("catalog.write", "connector.manage"));
```
mrouvroy-floe: [major] **Overlay writes still accept either permission**

`Authorizer.require(PrincipalContext, List<String>)` returns as soon as it finds any listed permission, so this call does not enforce the documented requirement for both `catalog.write` and `connector.manage`. The same pattern is used by `updateCatalogOverlay` and `deleteCatalogOverlay`.

A principal holding only `catalog.write` can therefore bind, mutate, or remove overlays for integrations it is not allowed to manage, including creating a dependency that blocks integration deletion.

Call the single-permission overload separately for each required permission in all three write RPCs, and add negative tests for principals that hold only one of the two grants.

---

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
```diff
@@ -0,0 +1,587 @@
...
+        secrets.delete(accountId, credentialRef);
+      }
+      var meta = integrations.metaFor(integration.getResourceId());
+      CatalogIntegration latest =
+          integrations.getById(integration.getResourceId()).orElse(integration);
+      if (latest.getPendingCredentialDeleteRefsCount() == 0) {
+        return latest;
+      }
+      CatalogIntegration cleared =
+          latest.toBuilder().clearPendingCredentialDeleteRefs().setUpdatedAt(nowTs()).build();
```
mrouvroy-floe: [major] **Concurrent cleanup drops undeleted credential refs**

`drainPendingCredentialDeletes` deletes only the refs from its original `integration` snapshot, but then reloads `latest` and clears its entire `pending_credential_delete_refs` list. A concurrent auth rotation can append a new pending ref between those two operations, causing this method to erase a ref for a secret it never deleted.

For example:

- Cleanup starts with `[R1]` and deletes `R1`.
- An auth update commits `[R1, R2]`.
- Cleanup reloads that row and clears both entries.
- The updater reloads the now-cleared row and has nothing left to drain.

`R2` is then permanently unreachable and account cleanup cannot remove it. Remove only the successfully processed refs from the latest row, retaining refs added concurrently, and CAS/retry that subtraction.

---

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
```diff
@@ -0,0 +1,533 @@
...
+                  throw GrpcErrors.invalidArgument(
+                      corr,
+                      null,
+                      Map.of("field", "display_name", "catalog", existing.getDisplayName()));
+                }
+                if (overlays.getByCatalog(accountId, catalogId.getId()).isPresent()) {
+                  throw GrpcErrors.alreadyExists(
+                      corr, null, Map.of("catalog_id", catalogId.getId()));
+                }
+                if (namespaces.count(accountId, catalogId.getId(), List.of()) > 0) {
```
mrouvroy-floe: [major] **Catalog attachment is not linearized with namespace creation**

The caller-owned catalog is checked for namespaces only once, before `persistNew` creates the overlay. `persistNew` later revalidates the integration and catalog name, but never rechecks namespace emptiness or coordinates this check with catalog child mutations.

A concurrent `CreateNamespace` can commit after this count and before the overlay is created, so the RPC succeeds while attaching to a catalog that was not empty at the attachment point. This violates the existing-empty-catalog contract and can mix caller-owned objects with future overlay reconciliation.

Serialize attachment with namespace creation through a catalog lifecycle/CAS guard, or recheck emptiness after committing the overlay and roll back/retry when the catalog child set changed. Add a race regression test around the initial count and `overlays.create`.

---

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
mcusack-floe: Superseded by the PR downscope. Credential refs, pending credential deletion, and cleanup code have been removed entirely, so this concurrent cleanup path no longer exists.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
mcusack-floe: Superseded by domain-specific permissions in the downscoped API. Reads require catalog-overlay.read; writes require catalog-overlay.write; create separately requires catalog-integration.use and catalog.write. connector.manage intentionally grants no authority over either new resource, because legacy Connectors are outside this contract.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
mcusack-floe: Superseded by the downscoped permission model. CreateCatalogOverlay uses separate require calls for catalog-overlay.write, catalog-integration.use, and catalog.write, so the grants are conjunctive. Update/delete require catalog-overlay.write. Tests also verify that legacy connector.manage alone cannot authorize the new APIs.

service/src/main/java/ai/floedb/floecat/service/catalog/impl/CatalogServiceImpl.java
mcusack-floe: Superseded by the PR downscope. All overlay-specific guards and mutations were removed from CatalogServiceImpl. An overlay is now a non-owning reference to an existing catalog; creation checks the catalog pointer version, but catalog rename/delete lifecycle enforcement is explicitly deferred to the later lifecycle design.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
mcusack-floe: Superseded by the PR downscope. Overlay ownership and catalog lifecycle management have been removed from the protobuf and service. DeleteCatalogOverlay now atomically deletes only the overlay pointer set; it never deletes or mutates the referenced catalog. Catalog lifecycle cleanup will be designed in a later PR.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
mcusack-floe: Superseded by the PR downscope. Authentication configuration, credential refs, the credential resolver integration, and secret persistence have been removed entirely from the new contract and implementation. Connectivity and credential handling will be introduced with independent connectivity-domain objects in a later PR.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
mcusack-floe: Superseded by the PR downscope. Credential fields, secret externalization, pending-delete state, secret cleanup, and account-cleanup coupling have all been removed. This PR persists no authentication or credential material.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
mcusack-floe: Superseded by the PR downscope. refresh_policy and ReconcilePolicy have been removed from CatalogOverlay and its service. Reconciliation policy belongs with the future connectivity/reconciliation objects, not this CRUD foundation.

core/proto/src/main/proto/floecat/common/common.proto
mcusack-floe: Retained and verified. Catalog overlays use the dedicated RK_CATALOG_OVERLAY = 18 resource kind throughout the service, repository, keys, and tests. RK_OVERLAY remains separate for its existing usage.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
mcusack-floe: Superseded by the PR downscope. CIS_DELETING and the tombstone/rollback flow have been removed. Integration deletion now checks dependent overlays and then atomically deletes the integration while conditioning on the dependency marker; concurrent overlay creation invalidates that transaction and is retried.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
mcusack-floe: Superseded by the PR downscope. There is no longer an empty-catalog attachment or catalog-ownership contract: an overlay is a non-owning reference to an existing catalog and may coexist with its contents. Overlay creation atomically conditions on the integration and catalog pointer versions, while catalog/namespace lifecycle coordination is explicitly deferred.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
```diff
@@ -0,0 +1,343 @@
...
+            () -> {
+              var pc = principal.get();
+              authz.require(pc, RolePermissions.CATALOG_INTEGRATION_WRITE);
+              String accountId = pc.getAccountId();
+              String corr = pc.getCorrelationId();
+              CatalogIntegrationSpec spec = request.getSpec();
+              validateCreateSpec(spec, corr);
+              String name =
+                  normalizeName(mustNonEmpty(spec.getDisplayName(), "display_name", corr));
+              String uri = mustNonEmpty(spec.getCatalogUri(), "catalog_uri", corr);
```
mrouvroy-floe: [major] **Reject secret-bearing catalog URIs**

`CreateCatalogIntegration` validates `spec.properties` for secret-looking keys, but `catalog_uri` is accepted as any non-empty string and persisted directly on the `CatalogIntegration`. The update path does the same.

URIs like `https://user:password@catalog.example` or `https://catalog.example?access_token=...` become durable resource data and are returned by read/list APIs. This breaks the PR's stated CRUD-only/no-secrets boundary and gives callers no protection against accidentally embedding credentials.

Parse and validate `catalog_uri` on create and update before storing it. At minimum, reject URI user-info components and query parameters whose names match secret patterns.

---

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
```diff
@@ -0,0 +1,343 @@
...
+              // Read the marker before the dependency index. An overlay created after this read
+              // advances the marker and makes the conditional delete fail; one created before it
+              // is visible to the following count.
+              long markerVersion = markerStore.catalogIntegrationOverlaysMarkerVersion(id);
+              int dependents = overlays.countByIntegration(pc.getAccountId(), id.getId());
+              if (dependents > 0) {
+                throw GrpcErrors.conflict(
+                    corr, null, Map.of("dependent_overlays", Integer.toString(dependents)));
+              }
+              if (!integrations.deleteWithPreconditionAndOverlayMarker(
```
mrouvroy-floe, line 513: [minor] **`overlays-marker` pointer is never deleted or GC-collected**

Each integration accretes an `/catalog-integrations/overlays-marker/{id}` pointer once its first overlay is created (via `createWhilePointersMatchAndBumpMarker`). When the integration is later deleted, `deleteWithPreconditionAndOverlayMarker` only *checks* the marker's version — it never removes it.

Neither `PointerGc` nor `CasBlobGc` scans the `overlays-marker/` prefix (they cover only `by-id`, `by-name`, `by-integration`, `by-catalog`). The result is one orphaned pointer left behind per integration that ever had an overlay — an unbounded slow leak that contradicts the PR's stated goal of including integration/overlay records in pointer and CAS GC.

There is no correctness impact (integration IDs are random, so a stale marker can never be re-associated), but the leak should be addressed. Either delete the marker in `deleteWithPreconditionAndOverlayMarker`, or add `overlays-marker/` to the GC prefix scan.

---

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
```diff
@@ -0,0 +1,450 @@
...
+      throw GrpcErrors.conflict(corr, null, Map.of("reason", "integration is not active"));
+    }
+    long integrationMarkerVersion =
+        markerStore.catalogIntegrationOverlaysMarkerVersion(integrationId);
+    Map<String, Long> requiredVersions = new LinkedHashMap<>();
+    Set<String> requiredAbsent = new LinkedHashSet<>();
+    requiredVersions.put(
+        Keys.catalogIntegrationPointerById(accountId, integrationId.getId()), integrationVersion);
+
+    authz.require(principal.get(), "catalog.write");
```
mrouvroy-floe: [minor] **Hidden `catalog.write` authz check uses a raw string literal and is inconsistently placed**

`createAndPersist` performs a third authorization check — `authz.require(principal.get(), "catalog.write")` — buried mid-helper, whereas the two entry-point checks (`CATALOG_OVERLAY_WRITE`, `CATALOG_INTEGRATION_USE`) are declared up front in `createCatalogOverlay` using `RolePermissions` constants.

This creates two problems. First, the bare string `"catalog.write"` is not discoverable alongside the other required permissions; a rename of the catalog-write permission would silently miss this site. Second, requiring catalog *write* to create an overlay is a non-obvious coupling given the PR explicitly states overlay CRUD "never creates, mutates, locks, or deletes catalogs".

Replace the raw literal with a named `RolePermissions` constant, move or document the check alongside the other up-front authz declarations, and add a comment explaining why catalog-write access is required for overlay creation.

---

core/proto/src/main/proto/floecat/integration/catalog_integration.proto
```diff
@@ -0,0 +1,108 @@
...
+  CIS_PAUSED = 2;
+}
+
+message CatalogIntegration {
+  ai.floedb.floecat.common.ResourceId resource_id = 1;
+  string display_name = 2;
+  optional string description = 3;
+  CatalogIntegrationType type = 4;
+  string catalog_uri = 5;
+  CatalogIntegrationState state = 7;
```
mrouvroy-floe: [nit] **Field-number gaps in new messages without `reserved` declarations**

`CatalogIntegration` jumps from field 5 to `state = 7` (skipping 6), `CatalogIntegrationSpec` skips 5, and `CatalogOverlay`/`CatalogOverlaySpec` similarly skip fields before `state`. These are brand-new messages so there is no wire-compatibility concern today.

However, unexplained gaps invite a future author to reuse a skipped number for an unrelated field, which would silently break wire compatibility with any client that stored data using the original layout.

If the gaps are intentional placeholders for deferred connectivity fields, add `reserved` declarations (e.g. `reserved 6;`) or inline comments documenting that intent. Otherwise close the gaps.

---

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
mcusack-floe: This is deliberate. GC will be wired in in a later PR. The overarching plan is in https://github.com/eng-floe/floecat/pull/439

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
mcusack-floe: Addressed in ca9311283. Catalog URIs are now parsed and must be absolute; user-info and secret-bearing query keys are rejected on both create and update. The validation also catches percent-encoded secret keys and signature parameters without including the URI in error details. Focused regression tests cover the accepted and rejected cases.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
mcusack-floe: Addressed in 8917d4cb5, superseding the earlier defer-to-GC response. A successful integration deletion now conditionally removes the overlays marker in the same atomic pointer transaction. An absent marker must remain absent, while a present marker must match its observed version and is deleted with the integration; concurrent overlay creation therefore still prevents deletion. Regression tests cover both paths.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
mcusack-floe: Addressed in 793221ec2. catalog.write now has a named RolePermissions.CATALOG_WRITE constant and is required at the CreateCatalogOverlay RPC boundary alongside catalog-overlay.write and catalog-integration.use, before either referenced resource is read. The inline rationale explains that binding changes what is SQL-visible through the target catalog even though the overlay does not mutate or own the Catalog resource. A negative regression test verifies missing catalog.write is denied before integration or catalog lookup.

core/proto/src/main/proto/floecat/integration/catalog_integration.proto
mcusack-floe: Addressed in ea3001bcb. The accidental gaps left by the downscope are closed: Integration and Overlay resource/spec fields are now numbered contiguously through their state and timestamp fields. properties remains at 99 as the deliberate extension field. Future authentication and policy contracts can choose their final field numbers when implemented rather than preserving legacy Connector-derived slots.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
```diff
@@ -0,0 +1,452 @@
...
+  }
+
+  @Override
+  public Uni<UpdateCatalogOverlayResponse> updateCatalogOverlay(
+      UpdateCatalogOverlayRequest request) {
+    return mapFailures(
+        runWithRetry(
+            () -> {
+              var pc = principal.get();
+              authz.require(pc, RolePermissions.CATALOG_OVERLAY_WRITE);
```
mrouvroy-floe: [major] **Missing `catalog.write` on overlay mutations**

`updateCatalogOverlay` and `deleteCatalogOverlay` only require `catalog-overlay.write`, but `createCatalogOverlay` correctly requires `RolePermissions.CATALOG_WRITE`.

These mutation paths can still alter a catalog's effective SQL-visible surface: updating `include_namespaces`, `exclude_namespaces`, or `state` changes what is visible, and deleting an overlay removes the binding entirely. A principal granted only `catalog-overlay.write` can therefore mutate a catalog's visible surface without the `catalog.write` permission that create correctly enforces.

Require `RolePermissions.CATALOG_WRITE` on overlay update and delete as well, and add regression tests mirroring the existing create authorization test.

---

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
```diff
@@ -0,0 +1,392 @@
...
+    try {
+      uri = URI.create(candidate);
+    } catch (IllegalArgumentException e) {
+      throw GrpcErrors.invalidArgument(corr, null, Map.of("field", field));
+    }
+    if (!uri.isAbsolute() || uri.getRawUserInfo() != null) {
+      throw GrpcErrors.invalidArgument(corr, null, Map.of("field", field));
+    }
+
+    String rawQuery = uri.getRawQuery();
```
mrouvroy-floe: [major] **Secret-bearing URI fragments bypass validation**

`validateCatalogUri` only inspects `rawQuery` and returns early when the query string is absent, so a URI like `https://catalog.example#access_token=redacted` or `https://catalog.example#client_secret=redacted` passes validation and is persisted.

The PR contract states that secret-bearing catalog URIs must be rejected. URI fragments are still part of the stored `catalog_uri`, so token-shaped fragment content carries the same disclosure risk as token-shaped query content.

Reject URIs that contain a fragment outright, or run the same secret-key scan over `uri.getRawFragment()` before accepting the candidate URI.

---

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
```diff
@@ -0,0 +1,452 @@
...
+  private static byte[] canonicalFingerprint(CatalogOverlaySpec spec) {
+    var canonical =
+        new Canonicalizer()
+            .scalar("display_name", normalizeName(spec.getDisplayName()))
+            .scalar("description", spec.getDescription())
+            .scalar("integration_id", spec.getIntegrationId().getId())
+            .scalar("catalog_id", spec.getCatalogId().getId())
+            .scalar("state", spec.getState())
+            .map("properties", spec.getPropertiesMap());
+    for (NamespacePath path : spec.getIncludeNamespacesList()) {
```
mrouvroy-floe: [minor] **Overlay idempotency fingerprint loses namespace-path boundaries**

`canonicalFingerprint` flattens every include/exclude path by calling `canonical.list("include_namespaces", path.getSegmentsList())` once per path. `Canonicalizer.list` appends one `include_namespaces[]=<seg>` line per segment with no per-path delimiter, so the grouping of segments into paths is erased.

Structurally different specs can therefore produce identical fingerprints — e.g. `include=[["a","b"]]` and `include=[["a"],["b"]]` hash the same. When a client reuses an idempotency key for one of these against the other, `IdempotencyGuard` treats them as the same request and returns the previously-created overlay instead of reporting a mismatch, silently ignoring a genuinely different create.

Encode path boundaries so grouping is preserved — for example join each path's segments with an intra-path separator (e.g. `/`) and separate paths with a distinct delimiter (e.g. `|`) before appending to the canonical form.

---

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
mcusack-floe: Addressed in `daaaf1277`. `UpdateCatalogOverlay` and `DeleteCatalogOverlay` now require `RolePermissions.CATALOG_WRITE` alongside `CATALOG_OVERLAY_WRITE`, before reading the overlay. Added negative regression tests for both operations and updated the authorization documentation.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
mcusack-floe: Addressed in `daaaf1277`. Catalog URIs containing any fragment—including an empty fragment—are now rejected on create and update before persistence. Tests cover token-bearing fragments, client-secret fragments, and empty fragments. Proto and service documentation now state this restriction.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
mcusack-floe: Addressed in `daaaf1277`. Each complete `NamespacePath` protobuf is now encoded as one canonical fingerprint value, preserving both segment and path boundaries without delimiter ambiguity. Added a regression test proving `[["a","b"]]` and `[["a"],["b"]]` produce different fingerprints.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
```diff
@@ -0,0 +1,392 @@
...
+                                spec.getState() == CatalogIntegrationState.CIS_UNSPECIFIED
+                                    ? CatalogIntegrationState.CIS_ACTIVE
+                                    : spec.getState())
+                            .setCreatedAt(now)
+                            .setUpdatedAt(now)
+                            .putAllProperties(spec.getPropertiesMap());
+                    if (spec.hasDescription()) builder.setDescription(spec.getDescription());
+                    CatalogIntegration created = builder.build();
+                    try {
+                      integrations.create(created);
```
mrouvroy-floe: [major] **Broken integration idempotency recovery on `NameConflictException`**

`createCatalogIntegration` treats every `NameConflictException` from the repository as `ALREADY_EXISTS`, even when the caller supplied an idempotency key and the conflicting row belongs to an earlier attempt with the same key.

In the standard partial-commit failure scenario, `IdempotencyGuard` writes a pending record, `integrations.create(created)` commits the resource pointers, and then the process or store fails before `finalizeSuccess` persists the succeeded idempotency record. A later replay with the same key can lose or outlive the pending record, attempt the create again with a new random id, hit the unique `by-name` pointer, and return `ALREADY_EXISTS` instead of the original `CreateCatalogIntegrationResponse`. The client observes a failed retry for an operation that actually committed.

In the idempotent branch, catch `NameConflictException`, look up the existing integration by name, compare it to the normalized request spec, and return the existing integration if it matches — mirroring the pattern already used for account/catalog creation.

---

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
```diff
@@ -0,0 +1,459 @@
...
+                    ? CatalogOverlayState.COS_ACTIVE
+                    : spec.getState())
+            .setCreatedAt(now)
+            .setUpdatedAt(now)
+            .putAllProperties(spec.getPropertiesMap());
+    if (spec.hasDescription()) builder.setDescription(spec.getDescription());
+    CatalogOverlay overlay = builder.build();
+    try {
+      boolean committed =
+          overlays.createAttached(
```
mrouvroy-floe: [major] **Broken overlay idempotency recovery on `NameConflictException`**

`createCatalogOverlay` has the same idempotency recovery gap as the integration creation path: after `createAttached` loses to an existing `by-name` or `by-catalog` pointer, the service reports only `ALREADY_EXISTS` without attempting recovery.

If the first idempotent attempt commits the overlay pointers and marker bump but fails before `IdempotencyGuard.finalizeSuccess`, a replay with the same idempotency key generates a new random overlay id, collides with the already-published overlay, and surfaces `ALREADY_EXISTS` to the client. The idempotency guard then deletes the pending record because the conflict is treated as non-retryable, permanently losing the ability to recover the original response.

In the idempotent branch, catch the conflict, look up the existing overlay by catalog and/or name, compare it to the normalized request spec, and return the existing overlay if it matches — consistent with how integration idempotency should also be fixed.

---

service/src/main/java/ai/floedb/floecat/service/security/RolePermissions.java
```diff
@@ -33,20 +33,37 @@ public final class RolePermissions {
...
+  public static final String CATALOG_OVERLAY_WRITE = "catalog-overlay.write";
   private static final List<String> READ_PERMS =
-      List.of("account.read", "catalog.read", "namespace.read", "table.read", "view.read");
+      List.of(
+          "account.read",
+          "catalog.read",
+          "namespace.read",
+          "table.read",
+          "view.read",
+          CATALOG_OVERLAY_READ);
```
mrouvroy-floe: [minor] **`default` read-only role grants `catalog-overlay.read` but not `catalog-integration.read`**

The baseline `default` role receives `CATALOG_OVERLAY_READ` but not `CATALOG_INTEGRATION_READ`, whereas every other role that can read one of these resources can read both.

The practical effect is that a baseline reader can `List`/`Get` overlays — which expose `integration_id` references — but receives `PERMISSION_DENIED` when resolving the referenced integration. Both are new read-only resources introduced together, so the asymmetry is easy to introduce by oversight.

If integrations are intentionally more sensitive than overlays for baseline readers, confirm it is deliberate and document the rationale in `docs/fixed-roles.md`. Otherwise add `CATALOG_INTEGRATION_READ` to `READ_PERMS` alongside the existing overlay read grant.

---

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
```diff
@@ -0,0 +1,459 @@
...
+
+  private CatalogOverlay createAndPersist(
+      CatalogOverlaySpec spec,
+      String name,
+      List<NamespacePath> includes,
+      List<NamespacePath> excludes,
+      String accountId,
+      String corr) {
+    ResourceId integrationId = scopedIntegrationId(accountId, spec.getIntegrationId());
+    integrations
```
mrouvroy-floe: [nit] **Redundant `getById` call in `createAndPersist`**

`createAndPersist` reads the integration three times consecutively: `getById` for the not-found check (line 197), `metaFor` to pin the version (line 203), then `getById` again to read state (line 204).

The first `getById` is fully redundant — the version must be pinned before the state read (as the comment notes), and that pin-then-read pair already produces the same not-found error. This is three storage round-trips where `metaFor` followed by one `getById` would suffice.

Drop lines 197–200 and let the existing pin-then-read pair handle both existence verification and state retrieval.

---

core/proto/src/main/proto/floecat/integration/catalog_overlay.proto
```diff
@@ -0,0 +1,119 @@
...
+  COS_PAUSED = 2;
+}
+
+message CatalogOverlay {
+  ai.floedb.floecat.common.ResourceId resource_id = 1;
+  string display_name = 2;
+  optional string description = 3;
+  ai.floedb.floecat.common.ResourceId integration_id = 4;
+  // A non-owning reference to an existing catalog. Overlay CRUD never manages catalog lifecycle.
+  ai.floedb.floecat.common.ResourceId catalog_id = 5;
```
arivarola-floe: Is `catalog_id` intended to be the target for the external namespaces? In the SQL specification examples, the catalog overlay itself becomes the top-level object in SQL paths:
```
CREATE CATALOG OVERLAY my_overlay
     WITH INTEGRATION = my_integration,
     INCLUDE_NAMESPACES = ('foo');

SELECT * FROM my_overlay.foo.table;
```

Based on that, I'd expect the namespaces in include_namespaces to be resolved relative to the overlay, rather than directly to the underlying catalog. Could you clarify the intended relationship here?

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
mcusack-floe: Correction to my earlier reply: the final contract supersedes that intermediate authorization change. A CatalogOverlay is itself the top-level SQL catalog and no longer owns or references a separate Catalog resource or public catalog_id. Create, update, and delete therefore use catalog-overlay.write consistently; create additionally requires catalog-integration.use because it establishes the integration binding. Requiring catalog.write would incorrectly recouple this API to the separate legacy Catalog model. The service and fixed-role documentation now state this explicitly.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
mcusack-floe: Verified on the final head. validateCatalogUri rejects any URI whose raw fragment is non-null, including an empty trailing #, before create or update can persist it. The regression coverage includes access_token fragments, client_secret fragments, and empty fragments, and the protobuf/service documentation records the restriction.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
mcusack-floe: Verified on the final head. Each normalized NamespacePath is serialized as its complete protobuf value and Base64-encoded as one canonical scalar, so both segment boundaries and path boundaries are preserved without delimiter ambiguity. The regression test proves [[a,b]] and [[a],[b]] produce different fingerprints.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
mcusack-floe: Fixed with a stronger atomic recovery design. An idempotent integration create first stores a durable PENDING record containing the reserved ResourceId. The integration canonical/name pointers and the immutable SUCCEEDED receipt, including the exact response body and MutationMeta, are then published in the same compareAndSetBatch transaction. There is no longer a commit-before-finalize window or recovery by mutable name. A replay returns the stored receipt; an ordinary non-idempotent duplicate still returns ALREADY_EXISTS. Repository-level failure-injection tests cover both an aborted batch, where neither publishes, and a committed batch whose acknowledgement is lost, where both remain published.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
mcusack-floe: Fixed with the same atomic receipt design for overlays. The idempotency record durably reserves the overlay ResourceId before creation. The overlay canonical/name/integration pointers, integration dependency-marker advance, parent and deletion-fence conditions, and immutable SUCCEEDED receipt are committed in one compareAndSetBatch transaction. Replays therefore return the exact stored response rather than recovering through a mutable name or catalog lookup. Repository-level failure-injection tests cover both pre-commit failure and lost acknowledgement after commit.

service/src/main/java/ai/floedb/floecat/service/security/RolePermissions.java
mcusack-floe: Fixed. CATALOG_INTEGRATION_READ is now included in READ_PERMS alongside CATALOG_OVERLAY_READ, so the default read-only role can resolve the integration referenced by any readable overlay. RolePermissionsTest and docs/fixed-roles.md were updated to match.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
mcusack-floe: Fixed in the final create path. Overlay creation now resolves the integration once with getByIdWithMeta inside createOrReplace. That single coherent read supplies both the integration body and the exact pointer version used as the atomic parent condition; the earlier getById, metaFor, getById sequence is gone.

core/proto/src/main/proto/floecat/integration/catalog_overlay.proto
mcusack-floe: Yes, that interpretation is now the contract. CatalogOverlay is the top-level SQL catalog, and include/exclude namespace paths are resolved in the external integration namespace space beneath that overlay. The public catalog_id field and all separate Catalog ownership/lifecycle coupling were removed from the protobuf and service. The resulting SQL path is overlay_name.namespace.table, matching the specification examples.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
```diff
@@ -0,0 +1,485 @@
...
+          boolean reservedIdentity,
+          IdempotencyGuard.SuccessCommitter<CatalogIntegration> completion,
+          com.google.protobuf.Timestamp now,
+          String corr) {
+    var existing = integrations.getByNameWithMeta(newId.getAccountId(), name);
+    if (existing.isPresent()) {
+      var current = existing.get();
+      if (mode == CreateMode.CM_RETURN_EXISTING) return current;
+      if (mode != CreateMode.CM_REPLACE)
+        throw GrpcErrors.alreadyExists(corr, null, Map.of("display_name", name));
```
mrouvroy-floe, line 304: [major] **Idempotent create can return `ALREADY_EXISTS` instead of replaying the committed resource**

In `createOrReplace`, when `getByNameWithMeta(name)` resolves and the mode is `CM_ERROR_IF_EXISTS`, the method throws `ALREADY_EXISTS` without first checking whether the existing resource is the one this idempotent attempt reserved (`newId`). The recovery logic in the `NameConflictException` catch block (lines 291-302) handles this correctly, but the early by-name return bypasses it entirely. The same gap exists in `CatalogOverlaysImpl.createOrReplace` (line 282).

This breaks idempotency for the exact concurrent-retry scenario the key is designed to protect: request A wins `createPending` with reserved id `idA`; request B (same key, same request body) loses and correctly adopts `idA` from A's pending record; B's creator calls `getByIdWithMeta(idA)` → empty (A not yet committed); B falls through to `getByNameWithMeta` → finds A's now-committed record; B throws `ALREADY_EXISTS` instead of returning A's result.

Fix: before throwing `ALREADY_EXISTS` on the by-name hit, compare the found resource's id against `newId`. If they match, treat it as a successful idempotent replay and return the existing resource. Apply the same fix to `CatalogOverlaysImpl`.

---

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
```diff
@@ -0,0 +1,485 @@
...
+                if (hasMeaningfulPrecondition(request.getPrecondition()))
+                  throw GrpcErrors.notFound(corr, null, Map.of("id", id.getId()));
+                return DeleteCatalogIntegrationResponse.newBuilder()
+                    .setMeta(integrations.metaForSafe(id))
+                    .build();
+              }
+              var meta = current.get().meta();
+              MutationOps.BaseServiceChecks.enforcePreconditions(
+                  corr, meta, request.getPrecondition());
+              if (request.getCascade()) {
```
mrouvroy-floe: [major] **Cascade delete removes dependent overlays without overlay write authorization**

`DeleteCatalogIntegrationRequest.cascade` allows a caller holding only `catalog-integration.write` to delete every dependent `CatalogOverlay` via `cascadeDeleteIntegration`, entirely bypassing the `deleteCatalogOverlay` RPC and its `catalog-overlay.write` authorization boundary.

This violates the PR contract (integration deletion should be rejected while dependent overlays exist) and creates a privilege-escalation path: an integration admin can silently remove catalog surfaces they are not authorized to administer directly. Overlays are independently-governed SQL-visible catalog resources, so their deletion must be gated on the overlay permission regardless of how it is triggered.

Fix: either drop the `cascade` branch from this PR and preserve the existing dependent-overlay conflict behavior, or enforce the same `catalog-overlay.write` check inside the cascade path and update the protobuf contract and documentation accordingly.

---

service/src/main/java/ai/floedb/floecat/service/repo/impl/IdempotencyRepositoryImpl.java
```diff
@@ -138,29 +161,117 @@ public void finalizeSuccess(
 
     String uri = Keys.idempotencyBlobUri(accountId, key, "success-" + UUID.randomUUID());
     blobs.put(uri, rec.toByteArray(), "application/x-protobuf");
```
mrouvroy-floe, line 163: [minor] **Atomic idempotency completion leaks the superseded `pending-` blob and the `success-` blob on CAS failure**

The atomic receipt path publishes the `SUCCEEDED` record as a `CasUpsert` returned from `prepareSuccess`, committed inside `GenericResourceRepository.createInternal`'s pointer batch. Unlike the previous inline `finalizeSuccess` — which explicitly deleted the superseded blob via `blobs.delete(previous.getBlobUri())` — nothing ever deletes the `pending-{uuid}` blob written by `createPending` (line 109) once the atomic batch commits: `createInternal` only calls `healCanonicalBlobIfMissing` and returns.

Additionally, when the atomic batch *fails* after `prepareSuccess` has already written the `success-{uuid}` blob (line 163), the caller has no reference to that blob and cannot delete it. Both blob types become permanently unreachable: `PointerGc.shouldSkipPointer` explicitly skips `SEG_IDEMPOTENCY` pointers, and `CasBlobGc` never scans the idempotency segment.

Fix: after a successful atomic commit, explicitly delete the `pending-{uuid}` blob (mirroring the old `finalizeSuccess` cleanup). On atomic batch failure, delete the `success-{uuid}` blob before propagating the error, or record it somewhere the GC can reach.

---

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
```diff
@@ -0,0 +1,485 @@
...
+              }
+              if (integrations.cascadeDeletionFenceVersion(id) > 0L)
+                throw GrpcErrors.conflict(
+                    corr, null, Map.of("reason", "cascade deletion in progress"));
+              long markerVersion = markerStore.catalogIntegrationOverlaysMarkerVersion(id);
+              int dependents = overlays.countByIntegration(pc.getAccountId(), id.getId());
+              if (dependents > 0)
+                throw GrpcErrors.conflict(
+                    corr, null, Map.of("dependent_overlays", Integer.toString(dependents)));
+              if (!integrations.deleteWithPreconditionAndOverlayMarker(
```
mrouvroy-floe, line 513: [minor] **Non-cascade delete racing a cascade can orphan the integration deletion-fence pointer**

The non-cascade delete pre-checks `cascadeDeletionFenceVersion(id) > 0` (line 376), but the subsequent `deleteWithPreconditionAndOverlayMarker` (line 384) conditions only on the canonical pointer version and the overlays-marker — it never re-checks or atomically deletes the `catalog-integrations/deleting/{iid}` fence.

Race: (B) non-cascade passes the `fence == 0` check; (A) cascade then installs the fence via `beginCascadeDeletion` and deletes all dependent overlays; (B) now observes `dependents == 0` and wins the final canonical CAS-delete, removing the integration, its `by-name` pointer, and the overlays-marker. A's terminal `deleteWithPreconditionAndOverlayMarkerAndFence` then fails on the missing canonical pointer, A's retry returns success (integration already gone) — but the fence pointer A installed is never cleaned up and leaks indefinitely.

Fix: include the fence pointer deletion in B's final CAS batch (conditioned on `fenceVersion == 0` to fail safely if A has since installed it), or make the fence check and the canonical delete a single atomic operation.

---

mcusack-floe: Fixed in 5dfd412ad. The non-cascade repository delete now includes the integration deletion fence in its required-absent CAS conditions, alongside the dependency-marker condition. If a cascade installs the fence after the service pre-check, the final non-cascade batch fails atomically and cannot orphan the fence. Added repository regression coverage.

service/src/main/java/ai/floedb/floecat/service/repo/impl/IdempotencyRepositoryImpl.java
mcusack-floe: Confirmed. Cleanup of superseded pending blobs and uncommitted success blobs is intentionally deferred to a follow-on stacked GC PR. That PR has not been opened yet. The deferred work is now called out explicitly in the PR summary so this change does not imply that the cleanup is implemented here.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
mcusack-floe: Fixed in 5dfd412ad. Cascade remains an explicit API operation, but now requires catalog-overlay.write in addition to catalog-integration.write before reading the integration or any dependent overlay. Added negative and positive authorization coverage and updated the protobuf, service docs, fixed-role docs, and PR summary.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
mcusack-floe: Fixed in 5dfd412ad. Both createOrReplace paths now compare an early by-name owner with the idempotency reservation before returning ALREADY_EXISTS. When the IDs match, the creator adopts the committed resource and IdempotencyGuard returns the immutable succeeded receipt. Added concurrent same-key replay regression tests for integrations and overlays.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
```diff
@@ -0,0 +1,814 @@
...
+                          name,
+                          spec.getType(),
+                          uri,
+                          preparedAuthentication.authentication(),
+                          mode,
+                          false,
+                          null,
+                          now,
+                          corr);
+                  credentialStore.deleteQuietly(created.replacedIntegration());
```
mrouvroy-floe: [major] **Unused prepared credential leak**

`credentialStore.store(...)` runs before `createOrReplace(...)`, but the success path only deletes `created.replacedIntegration()`. When `CM_RETURN_EXISTING` loses the initial existence-check race, `createOrReplace(...)` returns the concurrently created row with an empty `replacedIntegration`, so the credential prepared for the freshly allocated `integrationId` is never deleted.

That leaves secret material in Secrets Manager under an internal integration id that was never published. Account or integration cleanup cannot find it later because no `CatalogIntegration` row points at that id/generation.

Make the create outcome distinguish `created`, `replaced`, and `returnedExisting`, or otherwise track whether the prepared credential was actually published. Delete the prepared credential on any successful non-publish outcome, especially `CM_RETURN_EXISTING`.

---

🤖 _Generated with GPT-5.5._

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
```diff
@@ -0,0 +1,559 @@
...
+      CreateCatalogOverlayRequest request) {
+    return mapFailures(
+        runWithRetry(
+            () -> {
+              var pc = principal.get();
+              authz.require(pc, RolePermissions.CATALOG_OVERLAY_WRITE);
+              authz.require(pc, RolePermissions.CATALOG_INTEGRATION_USE);
+              String corr = pc.getCorrelationId();
+              CatalogOverlaySpec spec = request.getSpec();
+              String name = mustNonEmpty(spec.getDisplayName(), "display_name", corr);
```
mrouvroy-floe: [major] **Overlay names bypass catalog normalization**

`CatalogOverlay` uses the raw `display_name` for existing lookup, catalog collision checks, idempotency fingerprinting, persistence, and the shared `sql-catalogs/by-name` secondary pointer. Regular `Catalog` creation normalizes display names with `normalizeName(...)` before reserving the same SQL-visible namespace, so an overlay named ` sales ` can coexist with a catalog named `sales` even though the PR is trying to enforce one account-level top-level catalog namespace.

This breaks the shared reservation guarantee and makes overlay lookup/idempotency semantics disagree with the existing catalog API's name rules.

Normalize the overlay display name once on create/replace and on `display_name` updates using the same `BaseServiceImpl.normalizeName(...)`, then use that normalized value for lookups, collision checks, fingerprinting, and persistence.

---

🤖 _Generated with GPT-5.5._

service/src/main/java/ai/floedb/floecat/service/gc/PointerGc.java
```diff
@@ -196,6 +196,65 @@ public Result runForAccount(String accountId, long deadlineMs) {
...
+              return key != null && (key.contains("/by-name/") || key.contains("/by-integration/"));
+            },
+            nowMs,
+            minAgeMs);
+    scanned += overlaySecondaryPointers.scanned;
+    deleted += overlaySecondaryPointers.deleted;
+    missingBlobs += overlaySecondaryPointers.missingBlobs;
+    staleSecondaries += overlaySecondaryPointers.staleSecondaries;
+
     Result catalogsByName =
```
mrouvroy-floe, line 234: [major] **SQL catalog reservation pointers are not GCed**

`Schemas.CATALOG` and `Schemas.CATALOG_OVERLAY` both create `/accounts/{account}/sql-catalogs/by-name/{name}` as a secondary reservation pointer, but `PointerGc` only scans catalog by-name and overlay by-name/by-integration prefixes. It never scans `Keys.sqlCatalogPointerByNamePrefix(accountId)`.

If that shared reservation pointer becomes stale or its blob is missing, pointer GC will leave it behind. Future `Catalog` or `CatalogOverlay` creates with that name will keep failing on the secondary-pointer conflict, so the SQL catalog name can remain permanently blocked after the canonical resource is gone.

Add a scan for `Keys.sqlCatalogPointerByNamePrefix(accountId)` and a regression test for an orphaned `/sql-catalogs/by-name/...` pointer. `canonicalPointerForBlobUri(...)` already knows how to map catalog and overlay blob URIs back to their canonical pointers.

---

🤖 _Generated with GPT-5.5._

service/src/main/java/ai/floedb/floecat/service/integration/CatalogOverlaysImpl.java
mcusack-floe: Fixed in 691fe5a02. Overlay display names are normalized once on create and update, and display-name lookup applies the same normalization. The normalized value is now used for existing lookups, catalog collision checks, idempotency fingerprints, persistence, and shared SQL-catalog reservations. Added create, get, update, and collision regression tests.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
mcusack-floe: Fixed in 691fe5a02. CreateOutcome now distinguishes whether the newly allocated integration identity was published. A CM_RETURN_EXISTING race cleans the prepared credential generation for the unpublished identity. Credential cleanup is also durable: a global cleanup marker is written before replacement, rotation, integration deletion, or account cleanup, and PointerGc retries failed secret deletion. Added regression coverage for the return-existing race, definite create failures, cleanup retry, rotation, deletion, and account cleanup.

service/src/main/java/ai/floedb/floecat/service/gc/PointerGc.java
mcusack-floe: Fixed in 691fe5a02. PointerGc now scans Keys.sqlCatalogPointerByNamePrefix(accountId) alongside catalog and overlay secondary indexes. Added a regression test proving an orphaned SQL catalog name reservation is removed.

mcusack-floe:
Superseded by #452, which preserves this PR’s integration-API scope and uses `mc-account-deletion-fence` as the correct stack base. This PR remains closed to preserve its discussion and review history.

PR #427, opened 2026-07-31
Rescan PR1: Define scan reuse contracts and remove stale stats fallback

mcusack-floe:
## Summary

Defines the contracts needed to identify and reuse previously captured scan artifacts. This adds connector content identities, capture fingerprints, reconciler proto contracts, durable artifact references and bundle descriptors, related domain models, and focused tests.

This PR also removes Floecat's stale planner-stats fallback as one atomic contract retirement. Raw snapshot IDs are identities rather than a portable ordering contract (notably for Iceberg), and planners only observe finalized snapshots whose exact stats should already be available. The protobuf fields/statuses, Java result models, resolver rung, repository indexes/scans, and focused stale-path tests are removed together so this stack level remains buildable.

## Type of Change

- [ ] fix (bug fix)

## Testing

- Prior scan-contract tip: focused Maven build/tests and `make test` passed.
- Updated tip: `make fmt` passed; two merge-adjacent blank lines were removed and committed.
- Updated tip: static diff/parity checks passed.
- `make test` has not been rerun after the stale-fallback removal, per current instruction.

- [x] `make fmt`
- [ ] `make verify`

- [ ] No breaking changes

Removes the `stale_ok`, `stale_targets`, `HIT_STALE`, and stale-degradation protobuf contract. Removed field numbers and names are reserved. Query-time capture remains disabled by default; a finalized-snapshot stats miss is surfaced rather than satisfied from a numerically ordered prior snapshot.

## Checklist

## Additional Notes

- Base: `main`
- Dependency: none
- Next: [#428 — Plan and execute remote scan artifact reuse](https://github.com/eng-floe/floecat/pull/428)
- The stale cleanup spans service/repository files because removing the public contracts without their implementations and call sites leaves this stack level uncompilable.
- The gated legacy manifest/reference fallback for tables created before reuse manifests remains unchanged.


reconciler/src/main/java/ai/floedb/floecat/reconciler/jobs/SnapshotPlanManifestIds.java
```diff
@@ -19,13 +19,26 @@
...
+import java.util.Comparator;
 import java.util.List;
+import java.util.Map;
 
 public final class SnapshotPlanManifestIds {
   private SnapshotPlanManifestIds() {}
 
   public static String manifestHash(List<ReconcileFileGroupTask> fileGroups) {
-    return hashValue(String.join("\n", canonicalSnapshotFileGroups(fileGroups)));
+    ManifestDigest digest = new ManifestDigest("snapshot-plan-v2");
```
mrouvroy-floe, line 30: [major] **`snapshot-plan-v2` rollout breaks in-flight reconcile plans**

Changing `manifestHash()` to produce `snapshot-plan-v2` filenames renames every snapshot-plan blob. `DurableReconcileJobStore.validateSnapshotPlanManifestHash()` recomputes only the current algorithm, so any v1 plan submitted by an older remote planner—or persisted just before the service is upgraded—is rejected with `snapshot plan manifest URI hash mismatch`.

This makes an otherwise normal rolling deployment interrupt in-flight reconcile planning. The new JSON fields default cleanly when an old plan is read; it is the filename validation alone that causes the breakage.

Keep writing v2 filenames, but retain the v1 hash implementation and update the validator to accept either suffix until all v1 workers have drained.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

core/proto/src/main/proto/floecat/reconciler/reconciler.proto
```diff
@@ -320,8 +337,44 @@ message StatsObjectDescriptor {
...
+  string file_path = 1;
+  string source_fingerprint = 2;
+  string stats_capture_signature = 3;
+  repeated string realized_stats_selectors = 4;
+}
+
+message ReusableIndexArtifactMetadata {
+  string file_path = 1;
+  string source_fingerprint = 2;
+  string index_capture_signature = 3;
```
mrouvroy-floe, line 351: [major] **Missing `realized_index_selectors` allows wrong-column index reuse**

`ReusableIndexArtifactMetadata` records only file identity and policy signature, not the selectors actually present in the sidecar. For default policies such as `FIRST_N`, `indexCaptureSignature()` is unchanged when the execution schema is reordered, and `indexSourceFingerprint()` deliberately excludes the schema.

An unchanged data file can therefore match an older sidecar containing a different `indexed_columns` set. The planner cannot reject that candidate without reading the bundle, so reuse can mark the current index request complete while serving indexes for the wrong columns. Stats metadata already avoids the analogous ambiguity via `realized_stats_selectors`.

Add manifest-resident `realized_index_selectors` and require them to cover the newly resolved request, or include the resolved selector set in the capture signature.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

core/connectors/spi/src/main/java/ai/floedb/floecat/connector/spi/FloecatConnector.java
```diff
@@ -642,56 +642,8 @@ record SnapshotFileEntry(
...
-          partitionDataJson,
-          partitionSpecId,
-          equalityFieldIds,
-          sequenceNumber,
-          deletionVector,
-          List.of());
-    }
-
+      List<SnapshotIcebergDeleteFile> icebergDeleteFiles,
+      String contentIdentity) {
```
mrouvroy-floe, line 646: [major] **Connector SPI constructor break: `SnapshotFileEntry` and `SnapshotIcebergDeleteFile` lose old arities**

Adding `contentIdentity` removes all previous `SnapshotFileEntry` constructor descriptors (the 9-, 10-, and 11-argument forms), and `SnapshotIcebergDeleteFile` changes from five arguments to six. Existing connector sources stop compiling, and already-built connector plugins fail at runtime with `NoSuchMethodError`. This is a separate public SPI break from the stale-protobuf retirement documented by the PR.

The contract already defines an empty identity as "reuse forbidden," so compatibility does not require guessing an identity.

Preserve the old constructors as overloads that delegate to the new canonical constructors with `contentIdentity = ""` (and do the same for the old `SnapshotIcebergDeleteFile` form), allowing existing connectors to keep working safely until they opt into reuse.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

reconciler/src/test/java/ai/floedb/floecat/reconciler/impl/SnapshotPlanBlobStoreTest.java
```diff
@@ -113,6 +118,10 @@ void persistPlanRoundTripsScopedFileGroupJobs() {
         SnapshotPlanManifestIds.manifestBlobUri("acct", "job-1", List.of(group)),
         persistedTask.fileGroupPlanBlobUri());
     assertNotNull(blobStore.bytesByUri.get(persistedTask.fileGroupPlanBlobUri()));
+    String storedJson =
+        new String(
```
mrouvroy-floe, line 123: [nit] **Double-negative assertion obscures intent**

`assertFalse(!storedJson.contains("fileExecutionPlans"))` is a double negative that reduces to asserting the JSON *contains* `fileExecutionPlans`. It reads awkwardly and is easy to misparse during maintenance, especially next to nearby `assertFalse(json.contains(...))` calls that assert *absence*.

Replace with the direct form:
```java
assertTrue(storedJson.contains("fileExecutionPlans"));
```

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

reconciler/src/main/java/ai/floedb/floecat/reconciler/jobs/SnapshotPlanManifestIds.java
mcusack-floe: Good catch: this would reject v1 in-flight manifests during a mixed-version rollout. I am leaving that behavior intentional. Backward compatibility and rolling mixed-version deployment are not a focus at this stage; planner workers and the service are expected to move together, and an in-flight plan can be restarted. Accepting v1 would retain legacy fallback code that we do not want to carry yet, so the v2 hash remains the sole accepted contract.

core/proto/src/main/proto/floecat/reconciler/reconciler.proto
mcusack-floe: Fixed and propagated through the full five-PR stack. PR 427 now adds manifest-resident realized_index_selectors and makes default index capture signatures execution-schema-sensitive. PR 428 publishes the sorted persisted indexed_columns set into that metadata; its planner rejects missing or empty coverage, requires every explicit selector to be present, and rejects default-policy reuse after a schema change. PRs 429-431 contain the same fix. Regression tests cover both an explicit selector mismatch and a reordered default schema; the focused final-stack suite passed all 50 tests.

core/connectors/spi/src/main/java/ai/floedb/floecat/connector/spi/FloecatConnector.java
mcusack-floe: The source and binary compatibility observation is accurate, but the break is intentional. Backward compatibility is not a focus at this stage: the connector SPI and the in-tree connectors are versioned and rebuilt together. Retaining the old arities would add legacy fallback API surface that we do not want to preserve yet. An empty content identity remains the explicit no-reuse contract for new callers; it is not being used as a compatibility shim for already-compiled plugins.

reconciler/src/test/java/ai/floedb/floecat/reconciler/impl/SnapshotPlanBlobStoreTest.java
mcusack-floe: Agreed that the direct assertion is clearer. This is a readability-only nit with no behavioral effect; it was not changed as part of the requested index-reuse correctness fix, so the current push leaves it unchanged.

reconciler/src/main/java/ai/floedb/floecat/reconciler/jobs/SnapshotPlanManifestIds.java
mrouvroy-floe, line 30: [major] **`manifestHash` v1→v2 change breaks resumption of in-flight snapshot plans**

The hash scheme changed from a string-join SHA-256 (`hashValue(String.join("\n", ...))`) to a structured `ManifestDigest("snapshot-plan-v2")`. For the same logical plan these produce different hashes, and the hash is embedded in the manifest blob URI via `manifestBlobUri`.

`DurableReconcileJobStore.validateSnapshotPlanManifestHash` recomputes the hash from persisted file groups and requires the stored `manifestUri` to end with the expected hash suffix. Any `PLAN_SNAPSHOT`/`FILE_GROUPS` job whose manifest URI was persisted by the old build (v1) and then adopted after deploying this build (v2) will recompute a v2 hash, mismatch the stored v1 suffix, and throw `IllegalArgumentException("snapshot plan manifest URI hash mismatch ...")` — failing adoption of in-flight jobs across the deploy boundary.

Add a migration path or fallback: either accept both hash suffixes during a transition window, re-write stored URIs on adoption, or gate the new scheme on a feature flag until all in-flight jobs have drained.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/FileArtifactReuse.java
```diff
@@ -0,0 +1,383 @@
...
+    digest.add("properties");
+    digest.add(properties.size());
+    properties.forEach(
+        entry -> {
+          digest.add(entry.getKey());
+          digest.add(entry.getValue());
+        });
+    if (!stats
+        && effective.requestsIndexes()
+        && effective.selectorsForIndex().isEmpty()
```
mrouvroy-floe: [major] **`indexCaptureSignature()` omits execution schema for explicit selectors, enabling unsafe sidecar reuse after schema evolution**

`indexCaptureSignature()` only binds `executionSchemaJson` into the signature when the selector set is empty. Explicit connector-native name/path selectors are schema-relative: Iceberg, for instance, resolves a textual column name to a field ID in the snapshot schema, while `indexSourceFingerprint()` deliberately excludes that schema.

After schema evolution renames a field or reuses a name for a different field ID, an unchanged data file with the same textual selector produces an identical source fingerprint and capture signature. The realized-selector metadata can still contain that name, so a subsequent plan can accept a sidecar built for the old physical column — corrupting index lookups silently.

Canonicalize explicit selectors to stable field IDs before signing and coverage checks, or unconditionally bind the execution schema into the index capture signature whenever explicit selectors are present.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/FileArtifactReuse.java
```diff
@@ -0,0 +1,383 @@
...
+      TargetStatsRecord record,
+      String sourceFingerprint,
+      String statsSignature,
+      List<String> realizedSelectors) {
+    return record.toBuilder()
+        .putProperties(SOURCE_FINGERPRINT_PROPERTY, sourceFingerprint)
+        .putProperties(STATS_SIGNATURE_PROPERTY, statsSignature)
+        .putProperties(
+            REALIZED_STATS_SELECTORS_PROPERTY,
+            String.join(",", realizedSelectors == null ? List.of() : realizedSelectors))
```
mrouvroy-floe: [major] **`REALIZED_STATS_SELECTORS_PROPERTY` encoding is lossy and round-trip-unsafe**

`stampStats()` serializes the selector list with `String.join(",", ...)`. Connector-native column names and paths are required to be preserved verbatim, but there is no escaping, so a single selector named `a,b` is indistinguishable from two selectors `a` and `b`.

A reader converting this property back into `ReusableStatsArtifactMetadata.realized_stats_selectors` can claim coverage for columns that were never captured, or fail to recognize a column that was, leading to unsafe reuse or unnecessary rescans.

Use a structured, round-trippable encoding (e.g. JSON array, percent-encoding, or a repeated protobuf field instead of a flat string map property) and add tests with selectors containing commas and other delimiter characters.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/FileArtifactReuse.java
```diff
@@ -0,0 +1,383 @@
...
+      String indexSignature) {
+    return record != null
+        && record.hasTarget()
+        && record.getTarget().hasFile()
+        && expectedFilePath.equals(record.getTarget().getFile().getFilePath())
+        && !record.getArtifactUri().isBlank()
+        && record.getState() == IndexArtifactState.IAS_READY
+        && sourceFingerprint.equals(record.getPropertiesMap().get(SOURCE_FINGERPRINT_PROPERTY))
+        && indexSignature.equals(record.getPropertiesMap().get(INDEX_SIGNATURE_PROPERTY))
+        && !record.getPropertiesMap().getOrDefault("indexed_columns", "").isBlank();
```
mrouvroy-floe: [nit] **`"indexed_columns"` magic string should be a named constant**

`compatibleIndex` gates on `!record.getPropertiesMap().getOrDefault("indexed_columns", "").isBlank()` using a bare string literal. Every other property key in this class is a named `*_PROPERTY` constant (`SOURCE_FINGERPRINT_PROPERTY`, `INDEX_SIGNATURE_PROPERTY`, etc.).

This key is part of the reuse-compatibility contract shared with whatever writes the index record. A divergent literal elsewhere would silently make records look incompatible at runtime with no compile-time signal.

Promote `"indexed_columns"` to a named constant so both the producer and this consumer reference the same symbol.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

mcusack-floe: Fixed in 112b996a0. INDEXED_COLUMNS_PROPERTY is now shared by the producer and consumers, together with the common selector codec.

reconciler/src/main/java/ai/floedb/floecat/reconciler/jobs/SnapshotPlanManifestIds.java
mcusack-floe: The mixed-version issue is understood, but intentionally remains out of scope. Per the project stage, no v1 hash fallback or migration path is being retained; planner workers and service move together and in-flight work can restart.

reconciler/src/test/java/ai/floedb/floecat/reconciler/impl/SnapshotPlanBlobStoreTest.java
mcusack-floe: Fixed in 112b996a0. The assertion now uses the direct assertTrue form.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/FileArtifactReuse.java
mcusack-floe: Fixed in 112b996a0. Index capture signatures now bind the execution schema for both default and explicit selectors, with regression coverage for explicit-selector schema changes.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/FileArtifactReuse.java
mcusack-floe: Fixed in 112b996a0. Realized stats selectors and indexed_columns now use a shared strict JSON string-array codec. Tests cover commas, quotes, newlines, paths, and field IDs; no legacy comma fallback was added.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/GrpcRemoteReconcileExecutorClient.java
```diff
@@ -1690,17 +1691,9 @@ var record = artifact == null ? null : artifact.record();
...
-    }
-    Set<String> selectors = new HashSet<>();
-    for (String token : encoded.split(",")) {
-      if (token != null && !token.isBlank()) {
-        selectors.add(token.trim());
-      }
-    }
-    return Set.copyOf(selectors);
+    return Set.copyOf(
+        FileArtifactReuse.decodeSelectors(
```
mrouvroy-floe, line 1695: [minor] **`persistedIndexSelectors` hard-fails on a malformed `indexed_columns` value**

This PR changes the persisted `indexed_columns` property from a comma-joined string to a JSON array (`FileGroupIndexArtifactStager` now calls `FileArtifactReuse.encodeSelectors(...)`). `persistedIndexSelectors` reads it back via `FileArtifactReuse.decodeSelectors(...)`, which throws `IllegalArgumentException` on any value that is not a JSON array.

Today the writer and reader run in the same reconciler binary, so the value is always well-formed JSON and this never fires. But it is an inconsistent posture: the sibling read path `FileArtifactReuse.compatibleIndex` deliberately swallows a bad value via `hasEncodedSelectors` (catch → `false`), whereas `persistedIndexSelectors` propagates the exception and aborts `submitSuccess`/`validateIndexArtifactCoverage` for the whole file group. Any legacy or malformed `indexed_columns` value (e.g. a record produced by an older build) would turn a recoverable "treat as no selectors" case into a hard failure.

Consider decoding defensively here for symmetry — wrap the `decodeSelectors` call and fall back to `Set.of()` on `IllegalArgumentException`, matching `hasEncodedSelectors`.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

mrouvroy-floe (approved):

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/GrpcRemoteReconcileExecutorClient.java
mcusack-floe: The asymmetry is intentional. compatibleIndex is a candidate filter, so malformed persisted metadata makes that candidate ineligible. persistedIndexSelectors validates the current execution result before success publication; malformed metadata there violates the writer/worker contract and must fail the file group rather than be normalized into an empty set. Supporting legacy comma values is backward-compatibility fallback behavior and remains out of scope at this stage.

PR #428, opened 2026-07-31
Rescan PR2: Plan and execute remote scan artifact reuse

mcusack-floe:
## Summary

Adds remote scan-artifact reuse planning and execution on top of the contracts in #427. This includes prior-manifest matching, file-group bundle selection, Java worker and reconciler changes, Rust-facing remote reconciler contract updates, and the service-side compact-bundle commit/read path required to accept worker results.

Reuse now follows the current snapshot explicit `parent_snapshot_id` and loads only that predecessor immutable reuse manifest. If the snapshot, predecessor, or manifest is unavailable, planning regenerates the artifacts. The former fallback through `listFileStats` and `listFileIndexArtifacts` is removed because it synchronously loads unbounded historical artifacts. Delta version zero is retained as a valid explicit predecessor.

## Type of Change

- [x] fix (bug fix)

## Testing

- `make fmt` passed with no files reformatted.
- Focused `ReconcilerServiceTest` regression tests passed (2 tests, 0 failures).
- `make test` passed across the full 34-module reactor, including 363 reconciler tests and 1,653 service unit tests.

- [x] `make fmt`
- [ ] `make verify`

- [ ] No breaking changes

Tables without a finalized reuse manifest now regenerate stats and indexes rather than listing and loading historical artifacts. The bounded migration path for legacy records embedded inside an existing reuse manifest remains.

## Checklist

## Additional Notes

- Base: `mc-scan-reuse-contracts`
- Dependency: [#427 — Define scan reuse contracts and content identities](https://github.com/eng-floe/floecat/pull/427)
- Next: [#429 — Finalize and publish reusable scan artifacts](https://github.com/eng-floe/floecat/pull/429)
- The remote snapshot finalizer client, executor, and focused test are included here because the shared gRPC executor client implements both remote execution interfaces; separating them leaves this stack level uncompilable.
- Compact-bundle service commit mapping, pointer staging, shared-bundle decoding, and their focused tests are included here because execution emits compact bundles.


reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteFileArtifactReusePlanner.java
```diff
@@ -0,0 +1,180 @@
...
+      Map<String, BundleSelectionBuilder> selected = new java.util.LinkedHashMap<>();
+      boolean reusable = !fullRescan && !plan.contentIdentity().isBlank();
+      if (reusable && requestsStats) {
+        statsByPath.getOrDefault(plan.filePath(), List.of()).stream()
+            .filter(
+                candidate -> sourceFingerprint.equals(candidate.metadata().getSourceFingerprint()))
+            .filter(
+                candidate -> statsSignature.equals(candidate.metadata().getStatsCaptureSignature()))
+            .findFirst()
+            .ifPresent(candidate -> selectStats(selected, candidate.bundle(), plan.filePath()));
```
mrouvroy-floe: [major] **Make stats reuse atomic across auxiliary files**

The planner selects the main file's stats independently from its on-disk deletion vector and Iceberg delete-file stats. A compatible main record can therefore set `plan.reusesFileStats()` even when one auxiliary candidate is missing or incompatible.

`StandaloneJavaFileGroupExecutionRunner` uses only `plan.reusesFileStats()` to decide that no stats scan is needed. It then publishes the partial selection, while finalization requires exact coverage of the main file plus every auxiliary target and rejects the group as incomplete.

Collect the main and auxiliary candidates first, and add stats selections only when the complete set is reusable. Alternatively, make execution explicitly capture whichever auxiliary targets were not selected.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotPlanningReconcileExecutor.java
```diff
@@ -433,6 +442,283 @@ private PlannedSnapshotCapture planSnapshotCapture(
...
+    }
+    Snapshot snapshot = backend.fetchSnapshot(context, tableId, snapshotId).orElse(null);
+    if (snapshot == null) {
+      return Optional.empty();
+    }
+    String uri = snapshot.getSummaryOrDefault(SnapshotReuseManifestMetadata.URI, "").trim();
+    if (uri.isBlank()) {
+      return Optional.empty();
+    }
+    byte[] bytes = blobStore.get(uri);
```
mrouvroy-floe: [major] **Regenerate when the manifest object is missing**

A predecessor with a manifest pointer does not follow the advertised regeneration fallback when the referenced object is unavailable. `S3BlobStore.get()` throws `StorageNotFoundException`, while `InMemoryBlobStore.get()` returns `null` and the subsequent `bytes.length` dereference fails; either path aborts planning.

This leaves snapshots with a dangling or temporarily unavailable manifest pointer unable to rescan, even though the contract says unavailable manifests should regenerate artifacts.

Treat a not-found or `null` manifest read as `Optional.empty()` so enrichment continues with fingerprints but no reuse selections. Integrity and parse mismatches for an object that does exist can still fail closed.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
```diff
@@ -555,6 +563,34 @@ private StagedArtifactReferences prepareArtifactReferences(
...
+  private static ArtifactMappings artifactMappings(
+      FileGroupArtifactBundleDescriptor artifactBundle,
+      ReconcileFileGroupResultDescriptor descriptor) {
+    if (artifactBundle == null || !artifactBundle.hasArtifact()) {
+      throw new IllegalArgumentException("file-group artifact bundle is required");
+    }
+    StatsObjectDescriptor artifact = artifactBundle.getArtifact();
+    if (artifact.getPayloadUri().isBlank()
+        || !artifact.getPayloadUri().startsWith(descriptor.statsObjectPrefix())
+        || !artifact.getPayloadUri().contains("/reuse-bundles/")
```
mrouvroy-floe: [major] **Bind bundle URIs to the declared digest**

The compact-bundle commit accepts any URI under the stats prefix that merely contains `/reuse-bundles/`; it never requires the filename to match `payload_sha256`. This is weaker than the former per-target validation even though the client already writes bundles as `reuse-bundles/<sha256>.pb`.

The protection/pointer path subsequently stores only the URI and byte count, so the declared digest is no longer available when stats or index records are read. A faulty or compromised remote worker can therefore commit a mutable bundle URI and later change the records behind every target pointer that shares it.

Require the artifact URI to equal the expected bundle prefix plus the lowercase hex `payload_sha256` and `.pb`, and retain the same check in the stats/index repository backstops.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteFileArtifactReusePlanner.java
mcusack-floe: Fixed in bca2b74c3. Stats reuse is now selected atomically across the main file and every auxiliary deletion-vector/Iceberg-delete target. If any auxiliary is missing or incompatible, the planner selects no stats and regenerates the complete set. Added regression coverage for partial and complete auxiliary coverage.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotPlanningReconcileExecutor.java
mcusack-floe: Fixed in bca2b74c3. Manifest reads now treat StorageNotFoundException and null blob reads as no reusable manifest, while existing corrupt or digest-mismatched manifests still fail closed. Added coverage for both storage behaviors.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
mcusack-floe: Fixed in bca2b74c3. Compact bundle URIs must exactly match the stats prefix plus reuse-bundles/<lowercase payload_sha256>.pb, and stats/index repository backstops enforce the same digest binding. Repository reads also recompute the payload digest, with mutation regressions for both paths.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotFinalizeReconcileExecutor.java
```diff
@@ -494,6 +503,40 @@ private ValidatedFileGroupArtifacts loadValidatedArtifacts(
...
       throw new IllegalArgumentException("snapshot file-group stats do not cover successful files");
     }
+    ai.floedb.floecat.reconciler.rpc.ReusableArtifactBundleReference reuseBundle =
+        payload.hasReusableArtifactBundle() ? payload.getReusableArtifactBundle() : null;
+    if (reuseBundle == null
+        || !reuseBundle.hasArtifact()
+        || reuseBundle.getArtifact().getPayloadUri().isBlank()
+        || reuseBundle.getArtifact().getPayloadBytes() <= 0
+        || reuseBundle.getArtifact().getPayloadSha256().size() != 32
+        || !reuseBundle.getArtifact().getPayloadUri().startsWith(descriptor.statsObjectPrefix())) {
```
mrouvroy-floe, line 519: [major] **Reusable artifact bundle not validated against committed descriptors**

`loadValidatedArtifacts` binds `file_stats` and `index_artifacts` descriptors to the durable result via `artifact_references_sha256`, but it accepts `payload.reusable_artifact_bundle.artifact` based solely on valid shape and prefix — it never verifies that the artifact's URI, byte count, and SHA-256 match the compact bundle those committed descriptors reference.

A faulty or out-of-date worker can therefore commit bundle A to the service while placing bundle B in the packed result. Finalization then publishes B in the predecessor manifest, so the next rescan selects an uncommitted or missing bundle and fails downstream (e.g. in `readReference`) rather than reusing the bundle that was actually accepted.

Require the reusable bundle artifact to match every compact descriptor field (URI, byte count, SHA-256), ignoring only fields like `target_storage` that are legitimately allowed to diverge at finalization time.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteFileArtifactReusePlanner.java
```diff
@@ -0,0 +1,199 @@
...
+      ReconcileCapturePolicy capturePolicy) {
+    Set<String> realizedSelectors =
+        metadata.getRealizedIndexSelectorsList().stream()
+            .map(String::trim)
+            .filter(selector -> !selector.isBlank())
+            .collect(Collectors.toSet());
+    if (realizedSelectors.isEmpty()) {
+      return false;
+    }
+    return realizedSelectors.containsAll(capturePolicy.selectorsForIndex());
```
mrouvroy-floe: [major] **Index selector namespace mismatch causes spurious reuse cache misses**

`reusableIndexCoversPolicy` compares manifest `realized_index_selectors` directly against raw policy selectors. The Java path populates `indexed_columns` from `ParquetPageIndexEntry.columnName`, which is the physical Parquet column path, while capture policies support stable `#<field-id>` selectors and logical aliases.

Consequently, a policy requesting `#1` cannot reuse a Java-produced index whose realized metadata records `customer_id`, even when the source fingerprint, capture signature, schema, and indexed column all match. Reuse is silently skipped and the index is recomputed unnecessarily.

The planner tests fabricate `#1` as realized metadata and do not exercise the production encoding path, so this divergence is invisible in the test suite.

Resolve requested selectors to physical paths before comparing, or persist the requested stable selector and all known aliases alongside the physical name in the realized metadata. Add a test that exercises the real encoding path end-to-end.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/StandaloneJavaFileGroupExecutionRunner.java
```diff
@@ -60,54 +69,349 @@ public CaptureEngineResult execute(
...
+          List.of(),
+          IndexArtifactRecord.getDefaultInstance());
+    }
+  }
+
+  private byte[] readReference(String uri, long expectedBytes, byte[] expectedSha256) {
+    if (blobStore == null) {
+      throw new IllegalStateException("Blob store is unavailable");
+    }
+    byte[] bytes = blobStore.get(uri);
```
mrouvroy-floe, line 358: [minor] **`readReference` dereferences a possibly-null blob read without a null check**

`readReference` calls `blobStore.get(uri)` and immediately accesses `bytes.length` without checking for `null`. Elsewhere in this PR — `RemoteSnapshotPlanningReconcileExecutor.loadReuseManifest` (line 681) and `StatsRepository.TargetStatsStorage.loadBundle` (line 2115) — a `null` return from `blobStore.get` is treated as a valid "absent" signal. `BlobStore.get` is not contractually non-null, so at least one implementation can return `null`.

A `null` here produces a raw `NullPointerException`, which `loadReuseBundles` wraps as `"Failed to load reusable artifact bundle ..."` with an NPE cause — an opaque diagnostic for what is really a missing-object condition.

Add an explicit null check and throw a descriptive exception (e.g. `BlobNotFoundException` or an `IllegalStateException` with the URI) so that a vanished reuse bundle produces a clear, actionable error rather than a confusing NPE.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotFinalizeReconcileExecutor.java
mcusack-floe: Fixed in f8bbfd8ba. Finalization now requires the reusable bundle artifact URI, byte count, and SHA-256 to match every committed file-stats/index descriptor, ignoring only target storage identity. Added regressions for URI, size, and digest mismatches. Propagated through PR429–PR431.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteFileArtifactReusePlanner.java
mcusack-floe: Fixed in f8bbfd8ba. Java capture now resolves Delta/Iceberg stable IDs and logical names against footer field IDs/physical paths, persists stable, logical, and physical aliases, and validates coverage end to end. Added production-path tests for stable selectors, renames, column mapping, schema evolution, and default selection. Propagated through PR429–PR431.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/StandaloneJavaFileGroupExecutionRunner.java
mcusack-floe: Fixed in f8bbfd8ba. readReference now checks for a null blob read and throws a descriptive missing reusable bundle error containing the URI instead of dereferencing null. Added regression coverage. Propagated through PR429–PR431.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotPlanningReconcileExecutor.java
```diff
@@ -433,6 +443,291 @@ private PlannedSnapshotCapture planSnapshotCapture(
...
+      return Optional.empty();
+    }
+    if (bytes == null) {
+      return Optional.empty();
+    }
+    String declaredBytes =
+        snapshot.getSummaryOrDefault(SnapshotReuseManifestMetadata.BYTES, "").trim();
+    String declaredSha256 =
+        snapshot.getSummaryOrDefault(SnapshotReuseManifestMetadata.SHA256, "").trim();
+    if (declaredBytes.isBlank()
```
mrouvroy-floe: [major] **`loadReuseManifest` throws on incomplete manifest metadata instead of treating it as absent**

A predecessor whose manifest URI is nonempty but whose `BYTES` or `SHA256` metadata is missing causes `loadReuseManifest` to throw `IllegalStateException` rather than returning `Optional.empty()`. A nonnumeric `BYTES` value escapes as an unhandled `NumberFormatException`. Both cases describe a not-yet-finalized or unavailable manifest — the same semantic as a missing object — but crash `PLAN_SNAPSHOT` rescans for that predecessor instead of falling through to the documented regeneration path.

This is especially disruptive during partial rollout or recovery from an interrupted finalization: every subsequent rescan of the affected predecessor fails hard, blocking progress even though the PR contract says tables without a finalized manifest should regenerate their artifacts.

Treat absent or unparseable completeness metadata (missing `BYTES`/`SHA256`, or a `BYTES` value that does not parse as a number) as `Optional.empty()` before reading or comparing the declared values, so the regeneration path is taken instead of throwing.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedPlannerWorkerService.java
```diff
@@ -576,8 +578,14 @@ public boolean persistPlanSnapshotSuccess(
             : 0L;
     List<SubmitLeasedPlanSnapshotResultRequest.Chunk> stagedChunks =
         loadStagedPlanSnapshotChunks(principalContext, jobId, leaseEpoch, chunkCount);
+    List<PlannedFileGroupJob> referencedJobs =
```
mrouvroy-floe: [major] **`referencedJobs` direct-plan path enqueues jobs without validating their content**

The new direct-plan path loads a URI supplied by the remote planner and validates only `referencedJobs.size()`. Every nonempty job is then enqueued verbatim: its `scope`, `tableId`, `snapshotId`, group identity, and file set are never checked against `durableSnapshotTask` or the leased parent. The submitted URI is also not restricted to the expected account/job prefix, and is not revalidated against `SnapshotPlanManifestIds.manifestHash`.

A stale, corrupted, or compromised worker-written plan that carries the right group count can therefore enqueue `EXEC_FILE_GROUP` work for a different table or snapshot, or for an unexpected capture policy. Because `LeasedFileGroupExecutionService.resolvePlannedTask` trusts a nonempty child task directly, those files are scanned and artifacts are staged before the finalizer can detect the mismatch — at which point the damage is already done.

Each loaded job must be validated against the owning `durableSnapshotTask` (table ID, snapshot ID, scope, group identity) and the plan URI must be scoped to the expected account/job prefix before any job is enqueued.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

service/src/main/java/ai/floedb/floecat/service/repo/util/BaseResourceRepository.java
```diff
@@ -175,7 +175,7 @@ protected Optional<T> read(String key) {
     Optional<T> loaded =
         blobCacheable()
             ? blobCache.get(blobUri, this::loadAndParseBlob)
```
mrouvroy-floe: [minor] **Cached-blob branch still calls the bundle-unaware `loadAndParseBlob` instead of the bundle-aware variant**

This PR switched both the non-cacheable single-read branch and the batch path to `loadAndParseReferencedBlob`/`parseReferencedBlob`, which select a per-pointer record from a shared reuse bundle. The `blobCacheable()` branch on line 177 was not updated and still calls `this::loadAndParseBlob`, which invokes `parser.parse` on the whole blob without extracting the correct per-pointer record.

The impact is dormant today — every `blobCacheable` `GenericResourceRepository` (tables, accounts, connectors, …) resolves plain blob URIs, not bundle URIs — but it is a real footgun: if a bundle-backed reader is ever made cacheable, or if a bundle URI is ever surfaced to a cacheable reader, the cached entry will contain the wrong decoded value silently.

Update the `blobCacheable()` branch to use the same bundle-aware loading method (`loadAndParseReferencedBlob` or its functional equivalent) so all three read paths are consistent.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/GrpcRemoteReconcileExecutorClient.java
```diff
@@ -1664,21 +1729,27 @@ var record = artifact == null ? null : artifact.record();
             "index artifact targets do not match the planned file group");
       }
       Set<String> persistedSelectors = persistedIndexSelectors(record);
+      if (!requiredSelectors.isEmpty() && !persistedSelectors.containsAll(requiredSelectors)) {
```
mrouvroy-floe: [minor] **Explicit-selector coverage check compares raw selector strings instead of normalized identities**

The explicit-selector coverage check calls `persistedSelectors.containsAll(requiredSelectors)` using raw strings. The adjacent default-selection branch and the `FIRST_N` limit check in the same method were deliberately changed to normalize via `FileArtifactReuse.selectorIdentities(...)` so that a column-name form and its `#fieldId` alias are treated as equivalent.

If the configured selector (`policy.selectorsForIndex()`) and the persisted alias for the same column ever use different textual representations, this check will over-reject a fully valid (including reusable) index artifact with `"index artifact does not cover the explicitly requested selectors"`. It can only over-reject, never silently accept wrong reuse, but it would block valid reuse unexpectedly.

Apply `FileArtifactReuse.selectorIdentities(...)` normalization to both `persistedSelectors` and `requiredSelectors` before the `containsAll` comparison, consistent with the rest of the method.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

service/src/main/java/ai/floedb/floecat/service/repo/impl/StatsRepository.java
```diff
@@ -2060,14 +2071,71 @@ private boolean sameReference(PrewrittenStatsWrite other) {
...
+    @Override
+    protected TargetStatsRecord parseReferencedBlob(String pointerKey, String blobUri, byte[] bytes)
+        throws Exception {
+      if (!ReusableArtifactBundleUris.isBundleUri(blobUri)) {
+        return super.parseReferencedBlob(pointerKey, blobUri, bytes);
+      }
+      if (!ReusableArtifactBundleUris.matchesPayload(blobUri, bytes)) {
+        throw new CorruptionException("reusable artifact bundle digest mismatch: " + blobUri);
+      }
+      return targetFromBundle(pointerKey, ReusableArtifactBundlePayload.parseFrom(bytes));
```
mrouvroy-floe: [minor] **Bundle decoding does not enforce `format_version == 1`, creating worker/service disagreement**

The direct decode in `StatsRepository` and the `loadBundle` helper both parse `ReusableArtifactBundlePayload` and immediately select a target record without checking `format_version`. `IndexArtifactRepository.loadReusableArtifactBundle` has the same gap. By contrast, the worker-side reader explicitly rejects every version other than `1`.

A digest-valid version-`0` or future-version bundle can therefore be exposed through activated stats or index pointers on the service side while the worker would have rejected it outright. This makes rolling upgrades and legacy data recovery behave unpredictably depending on which component reads the bundle — the same on-disk payload is valid in one path and invalid in another.

Centralize compact-bundle decoding and require `getFormatVersion() == 1` in all read paths (cached, single-read, and batch) before selecting any target record, bringing the service in line with the worker's existing validation.

---

🤖 _Generated with Claude Opus 4.8 and GPT-5.6 Sol._

mrouvroy-floe (commented):
**[major] Reuse manifest is never published to the successor snapshot** — `service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java:171`

The finalization path validates the reuse manifest and publishes stats/index generations, but never writes the `SnapshotReuseManifestMetadata.URI/BYTES/SHA256` summary keys to the snapshot that the next planning run will inspect.

As a result, even after a compact bundle is successfully finalized and committed, the next non-full rescan treats the predecessor as having no finalized reuse manifest and regenerates stats and indexes from scratch. The reuse planner is effectively read-only in this PR: it can discover reuse from a predecessor but cannot propagate its own output forward.

Publish the manifest summary keys (URI, BYTES, SHA256) to the current snapshot during finalization, mirroring whatever test harness currently writes them, so that successive planning runs can discover the bundle.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteFileArtifactReusePlanner.java
mrouvroy-floe: [minor] **Index reuse coverage check uses raw `containsAll` instead of normalized selector identities**

`reusableIndexCoversPolicy` compares selectors with a plain `realizedSelectors.containsAll(capturePolicy.selectorsForIndex())` over trimmed strings, skipping the normalization applied everywhere else.

`GrpcRemoteReconcileExecutorClient.coversExplicitSelectors` (lines 1764–1774) normalizes via `FileArtifactReuse.selectorIdentities(...)` when selectors are field-id identities (`#`-prefixed), and `realizedIndexColumnCount` does the same. When one side expresses selectors as field-id identities and the other as plain names or `name#id` forms, the raw `containsAll` returns `false` even though the bundle genuinely covers the policy, causing the planner to skip valid reuse opportunities.

Replace the raw `containsAll` with the same `selectorIdentities`-based normalization used in the sibling comparison paths to keep the planner and local-execution verifier consistent.

---

mcusack-floe: Addressed in 67c9a44b4. I centralized explicit selector coverage so both RemoteFileArtifactReusePlanner and GrpcRemoteReconcileExecutorClient now call FileArtifactReuse.coversExplicitSelectors. The shared rule trims/deduplicates selector syntax and requires every explicit selector exactly, failing closed. We deliberately do not apply selectorIdentities to the required-selector collection: for a mixed policy such as #1 plus customer_email, that projection drops customer_email and could accept an artifact that never indexed it. Identity projection remains limited to realized-output consistency comparisons and column counting. Added regressions for syntax normalization and for preserving named requirements even when stable identity projections match. The 57 focused tests passed, and make fmt passed.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/StandaloneJavaFileGroupExecutionRunner.java
```diff
@@ -60,54 +69,352 @@ public CaptureEngineResult execute(
...
+                  List<ai.floedb.floecat.connector.spi.FloecatConnector.ParquetPageIndexEntry>
+                      completedPageIndexEntries =
+                          pageIndexEntries == null ? List.of() : pageIndexEntries;
+                  throwIfCancellationRequested(stop);
+                  completedStats.addAll(completedFileStats);
+                  completedIndexes.addAll(completedPageIndexEntries);
+                  throwIfCancellationRequested(stop);
+                });
+        throwIfCancellationRequested(stop);
+        for (TargetStatsRecord record : completedStats) {
```
mrouvroy-floe: [major] **Index-only captures publish spurious file stats, causing snapshot finalization mismatch**

`StandaloneJavaFileGroupExecutionRunner` publishes every `completedStats` record from each capture subcall, even for batches where `captureStats` is `false` and only a fresh page index is requested.

This is not a no-op for Java connectors: Delta and Iceberg treat an empty requested stats-kind set as `FILE` stats, so the index-only subcall returns file stats for staging. Those records are then included in `publishedFileStatsRecords`, `publishReuseBundle(...)`, and the file-group payload. However, finalization computes `statsRequested` from `capturePolicy.requestsStats()` and expects `payload.getFileStatsCount()` to be zero for a `PARQUET_PAGE_INDEX`-only policy. The result is a valid index-only scan failing snapshot finalization with a `snapshot file-group stats count mismatch` error.

Fix: gate the stats publishing loop on `captureStats` being `true`, or filter out file-stats records before assembling the payload when the capture policy is index-only.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotPlanningReconcileExecutor.java
```diff
@@ -433,6 +443,298 @@ private PlannedSnapshotCapture planSnapshotCapture(
...
+    if (bytes == null) {
+      return Optional.empty();
+    }
+    if (expectedBytes != bytes.length
+        || !MessageDigest.isEqual(
+            declaredSha256.getBytes(java.nio.charset.StandardCharsets.US_ASCII),
+            Base64.getEncoder()
+                .encodeToString(sha256(bytes))
+                .getBytes(java.nio.charset.StandardCharsets.US_ASCII))) {
+      throw new IllegalStateException("snapshot reuse manifest metadata mismatch: " + uri);
```
mrouvroy-floe: [major] **Predecessor manifest integrity mismatch causes infinite retry loop in planning**

`loadReuseManifest` throws a plain `IllegalStateException` when the predecessor snapshot's declared reuse-manifest metadata does not match the stored blob (byte-length/SHA mismatch at line 703, identity/format checks at lines 710/714). That exception propagates out of `loadPredecessorReuseManifest` → `enrichFileGroupTasks` (lines 468-469, no `try`/`catch`) into `execute`, where a non-`ReconcileFailureException` is classified as `FailureKind.INTERNAL` with the default `RetryDisposition.RETRYABLE`.

Because the predecessor snapshot's summary metadata and its immutable manifest blob never change, every retry re-hits the same mismatch. Planning for the successor snapshot retries indefinitely and never completes. This is inconsistent with every other unavailability path (not-found, network errors) which either surface a terminal failure or trigger manifest regeneration.

Fix: catch the `IllegalStateException` (or a purpose-built subclass) in `enrichFileGroupTasks` or `execute` and either (a) map it to a non-retryable `ReconcileFailureException`, or (b) treat the mismatch as a signal to discard and regenerate the predecessor manifest rather than aborting planning.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
```diff
@@ -312,8 +316,7 @@ public boolean persistSuccess(
               leaseEpoch,
               existing.fileGroupTask,
```
mrouvroy-floe: [major] **Replay path skips task resolution, leaving empty file paths and stalling finalization**

The terminal replay branch handles already-succeeded jobs by passing `existing.fileGroupTask` directly to `prepareArtifactReferences(...)`. For durable file-group jobs that task is only a compact reference: children are enqueued with `fileGroupTask().asReference()`, and `getLeaseView()` reconstructs just plan/group/table/snapshot/fileCount with empty `filePaths` and no execution plans.

The normal path calls `resolvePlannedTask(lease)` before staging; the replay path skips it. If the first commit completes the job but the process dies before `markPreparedFileGroup(...)`, the worker's idempotent retry enters this branch and cannot restage the bundle pointers because every stats/index target looks outside an empty planned group. Snapshot finalization then keeps waiting on the prepared marker indefinitely.

Fix: call `resolvePlannedTask(lease)` (or an equivalent lookup) in the replay branch before invoking `prepareArtifactReferences`, mirroring the normal path's task enrichment.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/StandaloneJavaFileGroupExecutionRunner.java
mcusack-floe: Fixed in 24987756b. Index-only execution now gates publication of connector-returned stats on captureStats, so Java connectors can still return incidental FILE stats without leaking them into the reuse bundle or file-group result. Added a regression that exercises an index-only capture and verifies zero published stats plus one staged page index.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotPlanningReconcileExecutor.java
mcusack-floe: Fixed in 24987756b. Integrity, parse, format, account, connector, table, and snapshot mismatches now raise a purpose-built terminal ReconcileFailureException (INTERNAL/NONE) instead of entering the generic retryable path. Missing, incomplete, or explicitly unmarked manifests still regenerate. Added terminal-classification and identity regressions.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedFileGroupExecutionService.java
mcusack-floe: Fixed in 24987756b. Terminal-success replay now resolves the compact child reference through the durable parent snapshot plan before preparing artifact references, matching the normal execution path. The regression uses an asReference() child task and verifies replay restages the expected file target and prepared marker.

mcusack-floe:
Implemented and pushed the final review fixes in `24987756b`.

Changes include:
- suppressing incidental connector stats for index-only captures;
- classifying corrupt reuse-manifest integrity/identity failures as terminal while retaining regeneration for unavailable or unmarked manifests;
- resolving compact child tasks during terminal-success replay;
- requiring stats reuse to cover the requested selectors;
- treating corrupt/missing reuse bundles as terminal and honoring cancellation during reuse-only work;
- HEAD-validating reused sidecars, including nonempty content and ETag consistency;
- selectively backporting the PR 431 complete-manifest contract: producers mark complete manifests and readers reject unmarked manifests for reuse.

The cumulative append-only manifest construction remains in PR 431. The separate review note about publishing the successor snapshot manifest is intentionally implemented in stacked PR 429 (`mc-scan-reuse-finalization`, currently `d0b390e9a`), where activation/finalization and snapshot-summary publication are introduced; it does not belong in this execution-only PR.

Validation:
- focused reconciler tests: 80 passed;
- focused service tests: 17 passed;
- full 36-module Maven unit-test reactor: BUILD SUCCESS;
- `make fmt`: passed;
- `git diff --check`: clean.

The three newly addressed inline review threads have replies and are resolved.

service/src/main/java/ai/floedb/floecat/service/repo/util/BaseResourceRepository.java
```diff
@@ -449,7 +477,7 @@ protected List<KeyedValue<T>> listByPrefixWithKeys(
             }
 
             try {
-              T parsed = parser.parse(bytes);
+              T parsed = parseReferencedBlob(row.getKey(), blobUri, bytes);
               if (blobCacheable()) {
                 blobCache.put(blobUri, parsed);
```
mrouvroy-floe, line 482: [minor] **Batch list path caches bundle projections under URI-only key**

The single-read path was made bundle-aware (line 177): for bundle URIs it routes through `blobCache.getProjection(blobUri, key, ...)` so each pointer's per-record projection gets a distinct cache key. The batch path in `listByPrefixWithKeys` was not given the same treatment — `getAllPresent(uris)` (line 453) and `blobCache.put(blobUri, parsed)` (line 482) both key by URI only, even though `parseReferencedBlob(row.getKey(), ...)` returns a per-pointer projection of a shared bundle.

Within a single call results are correct because each row is re-parsed with its own key. Across calls, however, the URI-only cache entry holds whichever pointer's projection was written last, so a later `getAllPresent([bundleUri])` returns that stale projection for every reader of that bundle URI — causing incorrect data to be served from cache.

Apply the same bundle-aware keying used in the single-read path: use a composite `(blobUri, rowKey)` cache key for bundle URIs in both `getAllPresent` and `blobCache.put` within `listByPrefixWithKeys`.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/StandaloneJavaFileGroupExecutionRunner.java
```diff
@@ -60,54 +69,373 @@ public CaptureEngineResult execute(
...
+          }
+        }
+        if (captureIndex) {
+          stagedIndexArtifacts.addAll(
+              stampCapturedIndexes(
+                  FileGroupIndexArtifactStager.stage(
+                      payload.tableId(),
+                      payload.snapshotId(),
+                      paths,
+                      completedStats,
```
mrouvroy-floe, line 187: [minor] **Index-only staging path undercounts rows when stats are reused but index is not**

The new split capture path can invoke `FileGroupIndexArtifactStager.stage(...)` with `captureIndex == true` and `captureStats == false`. In this case `completedStats` is empty even though the file's stats were already reused earlier in the same execution.

When the stager has no `FileTargetStats`, it synthesizes file stats from page-index entries and uses the maximum page `liveRowCount` as the file row count. For multi-page or multi-row-group files this under-reports `IndexCoverage.rows_indexed` / `live_rows_indexed` and writes incorrect sidecar metadata for the freshly generated artifact.

Fix by passing the reused file stats to index staging when stats reuse succeeds but index reuse misses, or by changing the stager's fallback to sum row coverage across all pages or row groups rather than taking the maximum of any single page.

---

mrouvroy-floe (approved):

PR #429, opened 2026-07-31
Rescan PR3: Finalize and publish reusable scan artifacts

mcusack-floe:
## Summary

Completes scan reuse by publishing, finalizing, selecting, and consuming reusable artifacts under one fail-closed end-to-end contract. This adds snapshot finalization, immutable result-manifest validation, repository activation and identity mappings, migration behavior, and focused tests.

Reuse selection is bounded by the table root: `GetLatestFinalizedSnapshot` selects the newest non-excluded root-published finalized candidate without scanning snapshot history. Once selected, that basis cannot silently fall through to an older snapshot or cold planning. Missing or temporarily invisible root, snapshot, manifest, or bundle objects are retryable; malformed identities, references, manifests, and compatibility metadata are terminal.

Publisher and planner share structural reuse-manifest validation. Finalization validates exact deduplicated planned stats/index coverage, including Delta deletion vectors and Iceberg delete files; binds every content-addressed bundle descriptor to the service-accepted file-group artifact digest; and verifies each metadata entry belongs to its durable group. This remains metadata-only and performs no per-bundle payload GETs in the control plane. The consuming worker verifies bundle byte length, SHA-256, and record compatibility before reuse.

This update also fixes the nondeterministic `BoundedFanoutTest.refillsBeforeDeliveringCompletionToTheCaller` CI failure. A task could complete during initial submission, opening a permit immediately before ordered delivery blocked the caller. `BoundedFanout` now refills that opening before delivery.

## Type of Change

- [x] fix (bug fix)
- [x] docs
- [x] test

## Testing

- `make fmt` passed.
- Protobuf generation and compilation passed: 1,110 generated sources compiled.
- Focused reconciler suite: 62 tests passed.
- Focused service suite: 89 tests passed.
- Full `BoundedFanoutTest`: 36 tests passed.
- Previously flaky refill test: 20 consecutive isolated runs passed.
- `git diff --check` passed.
- No integration or smoke tests were run.

This intentionally removes hidden reuse fallbacks. Implementations must provide finalized reuse selection. Tables without a root-published finalized reuse candidate regenerate artifacts; after a candidate is selected, loss or invalidity is surfaced rather than converted to stale or cold planning.

Compatibility metadata is trusted executor publication metadata, cryptographically bound to service-accepted staged mappings. Planner/finalizer executors and the control plane must be cut over together after existing leases drain. Rolling or split deployments across these plan revisions are unsupported.

## Additional Notes

- Base: `main`
- Dependency: #428 (merged)
- Follow-ons: #430 and #431
- Review hardening: `6b39c3037`
- Backported from the later stack: retryable bundle-read classification and terminal handling for blank reuse-manifest URIs.
- The append-only work can build on this bounded finalized-snapshot API without replacing the contract.


service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
```diff
@@ -786,6 +921,80 @@ private static byte[] sha256(byte[] bytes) {
...
+              .addAllRealizedIndexSelectors(
+                  FileArtifactReuse.decodeSelectors(
+                          record.getPropertiesOrDefault(
+                              FileArtifactReuse.INDEXED_COLUMNS_PROPERTY, ""))
+                      .stream()
+                      .sorted()
+                      .toList()));
+      indexDescriptors.add(artifact.toBuilder().setTargetStorageId("file:" + filePath).build());
+    }
+    if (!expected.build().equals(submitted)
```
mrouvroy-floe: [blocker] **`validateReusableArtifactBundle` equality check compares full payload against deduplicated manifest bundle**

`expected` is reconstructed from every record in the bundle payload blob, then required to be exactly equal to `submitted`. However, `submitted` is the bundle reference as it appears in `SnapshotCaptureManifest`, which has already been rewritten by `RemoteSnapshotFinalizeReconcileExecutor.deduplicateSnapshotArtifacts` via `bundle.toBuilder().clearFileStats()` — re-adding only entries this bundle *owns*.

Whenever the same stats target appears in two bundles (e.g. on-disk deletion-vector paths or Iceberg delete-file paths added by `expectedFileStatsTargets`/`expectedFileStatsPaths`), the manifest copy is a strict subset of the full payload and the equality can never hold. This sharing is reachable in normal operation, not a hypothetical edge case, so the validation will spuriously fail for any snapshot where artifact deduplication has split stats across bundles.

Fix: compare only the entries the manifest declares this bundle owns (i.e. compare `submitted` against the filtered/deduplicated view of `expected`, not the raw payload union), or restructure the check to verify each declared entry is *present* in the payload rather than requiring set equality.

---

🤖 _Generated with Claude Opus 5 and GPT-5.5._

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
```diff
@@ -577,6 +695,22 @@ private void publishCaptureArtifacts(
...
+          }
+          matchedPrefix = statsPrefix;
+        }
+      }
+      if (matchedPrefix == null) {
+        throw new IllegalArgumentException(
+            "snapshot reusable bundle does not match a staged file group");
+      }
+      String expectedArtifactDigest = stagedArtifactDigests.remove(matchedPrefix);
+      validateReusableArtifactBundle(bundle, expectedArtifactDigest);
```
mrouvroy-floe: [major] **One serial blob GET per file group executed inside the fenced finalize-commit window**

`validateReusableArtifactBundle` calls `loadRequiredPublishedObject` for every bundle, so `publishCaptureArtifacts` now issues one blob GET per file group. This work happens *after* `jobs.beginSnapshotFinalizeCommit`, i.e. inside the fenced commit phase, and runs fully serially.

For a snapshot planned into hundreds or thousands of file groups this turns a metadata-only commit into an O(file-groups) round-trip storm while the lease is held. The widened window materially increases the probability that a lease expiry or node loss forces a full replay of the finalize phase. It also contradicts the design note in `docs/proto.md` stating that the compact reuse index exists precisely so consumers do not need to read bundle payloads.

Fix: move bundle payload validation outside the commit fence (e.g. into `prepareReusableArtifacts`), or issue the blob GETs in parallel. If the integrity guarantee is essential inside the fence, consider a lightweight manifest-hash check rather than fetching full payloads.

---

🤖 _Generated with Claude Opus 5 and GPT-5.5._

service/src/main/java/ai/floedb/floecat/service/repo/impl/SnapshotManifests.java
```diff
@@ -384,6 +385,29 @@ public static Optional<SnapshotManifestEntry> latestQueryableCurrent(
     return Optional.ofNullable(best[0]);
   }
 
+  /** The newest finalized entry at or before the committed current that publishes reuse data. */
+  public static Optional<SnapshotManifestEntry> latestReusableCurrent(
```
mrouvroy-floe: [major] **`latestReusableCurrent` / `getLatestFinalizedSnapshotForReuse` performs an unbounded full-history walk**

`latestReusableCurrent` uses `forEachEntry`, which reads every page of the snapshot-manifest chain with no short-circuit and no use of the cached `manifestEntryIndex`. `getLatestFinalizedSnapshotForReuse` (SnapshotRepository.java:421) delegates to this, so every incremental planning job pays an O(table snapshot history) cost regardless of where the latest reusable entry sits.

This directly contradicts the stated contract: the `GetLatestFinalizedSnapshotRequest` proto comment, the `SnapshotService` row in `docs/proto.md`, and the PR description all promise a lookup that "selects a root-published reuse basis without scanning snapshot history." Tables with no reusable manifest after a long history pay a full walk before deciding to regenerate artifacts — the exact behavior this PR claims to remove from the fallback path.

Fix: publish a root-level pointer to the latest reusable finalized snapshot (analogous to how `manifestEntryIndex` fast-paths other lookups), or stop the `forEachEntry` walk as soon as an entry newer than the ordering threshold is found.

---

🤖 _Generated with Claude Opus 5 and GPT-5.5._

service/src/main/java/ai/floedb/floecat/service/repo/impl/SnapshotRepository.java
```diff
@@ -211,6 +214,49 @@ public Optional<Snapshot> getById(ResourceId tableId, long snapshotId) {
...
+        || statsGenerationManifestUri == null
+        || statsGenerationManifestUri.isBlank()) {
+      throw new IllegalArgumentException("complete reuse manifest metadata is required");
+    }
+    SnapshotKey key = new SnapshotKey(tableId.getAccountId(), tableId.getId(), snapshotId);
+    for (int attempt = 0; attempt < 8; attempt++) {
+      MutationMeta meta = repo.pointerMetaForSafe(key);
+      Snapshot current =
+          repo.getByKey(key)
+              .orElseThrow(() -> new IllegalStateException("snapshot not found: " + snapshotId));
```
mrouvroy-floe: [minor] **`recordReuseManifest` throws non-retryable `IllegalStateException` on a concurrently deleted snapshot**

Every other failure mode in this method surfaces as `StorageAbortRetryableException`, but a missing snapshot (deleted concurrently) raises `IllegalStateException`. The call site runs after `publishCaptureArtifacts` inside the finalize commit, so a `DeleteSnapshot` landing in that window turns into a non-retryable internal error rather than something `runWithRetry` or an operator can reason about.

Deletion is arguably not an error in this context: the artifacts were already published, and there is simply no longer a snapshot to attach a reuse ref to. This mismatch also inconsistent with how `TableRootWriter.commitSnapshotEntry` handles the same race one call later.

Fix: either return quietly (the reuse ref attachment is moot if the snapshot is gone) or throw `StorageAbortRetryableException` so the surrounding retry loop can observe the snapshot's absence and terminate cleanly.

---

🤖 _Generated with Claude Opus 5 and GPT-5.5._

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedPlannerWorkerService.java
```diff
@@ -727,9 +727,8 @@ static void validateReferencedPlan(
           invalidReferencedPlan("snapshot plan contains duplicate file execution plans");
         }
       }
-      if (!executionPlans.isEmpty()
-          && (executionPlans.size() != groupFilePaths.size()
-              || !executionFilePaths.equals(groupFilePaths))) {
+      if (executionPlans.size() != groupFilePaths.size()
```
mrouvroy-floe, line 730: [minor] **Execution plans silently became mandatory while the mixed-version operator guidance was deleted from docs**

The `!executionPlans.isEmpty() &&` guard is removed, so a submitted plan is now rejected unless it carries one execution plan per declared file path. The producer-side change in `RemoteSnapshotPlanningReconcileExecutor.partitionFiles` (execution plans built even when `executionSchemaJson` is blank) keeps this consistent within a single build, but a control plane on this revision will reject any plan produced by an older reconciler.

This is a hard executor/control-plane incompatibility — the same class as the new mandatory `ReconcilerBackend.latestReconciledSnapshotForReuse` field introduced in this PR. Simultaneously, the PR deletes the "Breaking remote-executor revision" section from `docs/proto.md`, which carried exactly the guidance operators need: "versions must be cut over together after existing leases drain."

Fix: restore (or replace) the operator cut-over guidance in `docs/proto.md`, and consider a transitional period where the control plane tolerates empty execution-plan lists from older producers before enforcing the new invariant.

---

🤖 _Generated with Claude Opus 5 and GPT-5.5._

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
```diff
@@ -786,6 +946,123 @@ private static byte[] sha256(byte[] bytes) {
     }
   }
 
+  ValidatedReusableArtifactBundle validateReusableArtifactBundle(
+      ReusableArtifactBundleReference submitted, String expectedArtifactDigest) {
+    StatsObjectDescriptor artifact = submitted.getArtifact();
+    byte[] bytes =
+        loadRequiredPublishedObject(
```
mrouvroy-floe: [major] **Bundle payload validation runs inside the point-of-no-return commit window**

`validateReusableArtifactBundle` loads and hashes reusable bundle blobs for every submitted bundle, but this is called only after `jobs.beginSnapshotFinalizeCommit(...)` has accepted the snapshot finalize commit.

That marker is documented in `ReconcileJobStore` as the point of no return for snapshot publication. Performing serial object-store reads per reusable file group inside that fenced window makes large snapshots spend the commit section on remote blob I/O, increasing lease-expiry and replay risk. It also contradicts the documented finalizer contract that reusable bundle references are carried without reading bundle payloads.

Move bundle payload validation before `beginSnapshotFinalizeCommit`, or prefetch it into a prepared validation result so the commit window contains only metadata writes.

---

service/src/main/java/ai/floedb/floecat/service/repo/impl/SnapshotRepository.java
```diff
@@ -353,6 +400,31 @@ public Optional<Snapshot> getCommittedCurrentSnapshot(ResourceId tableId) {
...
+        || lookup.root() == null
+        || !lookup.root().hasCurrentSnapshotId()
+        || !lookup.root().hasSnapshotManifestRef()) {
+      return Optional.empty();
+    }
+    for (SnapshotManifestEntry entry : lookup.root().getReusableSnapshotCandidatesList()) {
+      if (excludedSnapshotId != null && entry.getSnapshotId() == excludedSnapshotId) {
+        continue;
+      }
+      if (!entry.hasSnapshotRef() || entry.getSnapshotRef().getUri().isEmpty()) {
```
mrouvroy-floe: [minor] **`getLatestFinalizedSnapshotForReuse` cannot fall through to the second candidate on a load miss**

The loop returns on the first non-excluded candidate: it returns `Optional.empty()` when that candidate lacks a `snapshotRef` (lines 420–421), and returns a filtered-to-empty result when the loaded snapshot lacks a `reuseManifestRef` (line 423). In both cases it never advances to the retained second candidate.

The `table_root.proto` comment justifies keeping two entries because "a replay may exclude the newest snapshot and request its predecessor." That replay case is handled by the `continue` on `excludedSnapshotId`, but a newest candidate that is present-but-unusable (missing ref, or a snapshot blob that no longer carries `reuse_manifest_ref`) short-circuits the whole lookup to empty rather than falling through to the older, still-valid basis — so the second slot provides no resilience against a corrupt or incomplete newest entry.

Replace the early-return paths with `continue` so the loop advances to the next candidate before giving up.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
mcusack-floe: Fixed in 0a881b95f. The per-bundle blob GETs were removed from `publishCaptureArtifacts`; the fenced commit path now carries and structurally validates bundle references without reading their payloads. `LeasedSnapshotFinalizeExecutionServiceTest` explicitly verifies that the bundle URI is never GET.

service/src/main/java/ai/floedb/floecat/service/repo/impl/SnapshotRepository.java
mcusack-floe: Fixed in 0a881b95f. `recordReuseManifest` now returns `Optional<Snapshot>`: empty when concurrent deletion wins, or the exact persisted revision on success. The finalizer treats empty as a successful no-op and skips pointer advancement. Passing the returned revision directly also removes the follow-up reload and its deletion-to-NOT_FOUND race.

service/src/main/java/ai/floedb/floecat/service/repo/impl/SnapshotManifests.java
mcusack-floe: Fixed by the root-level bounded candidate index in 5d7ecaf9f and completed in 0a881b95f. `getLatestFinalizedSnapshotForReuse` reads `reusable_snapshot_candidates` directly and never walks snapshot-manifest history. Both root mutation and legacy-root synthesis publish the newest two eligible candidates; synthesis coverage now verifies ordering and the committed-current boundary.

service/src/main/java/ai/floedb/floecat/service/repo/impl/SnapshotRepository.java
mcusack-floe: Fixed in 0a881b95f. `getLatestFinalizedSnapshotForReuse` now continues to the retained predecessor when a candidate lacks a snapshot ref, cannot be loaded, or no longer carries a reuse manifest. `latestFinalizedSnapshotForReuseFallsBackWhenNewestCandidateIsUnusable` covers the stale-newest/valid-predecessor case.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
mcusack-floe: Fixed in 0a881b95f. Bundle payload validation and all per-bundle object-store GETs were removed rather than moved. The point-of-no-return section now performs only structural reference checks plus the existing prepared-file-group verification; the only payload GET in this flow is the single snapshot capture manifest read.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedPlannerWorkerService.java
mcusack-floe: Addressed in 5d7ecaf9f. The strict one-execution-plan-per-file invariant is intentional; no mixed-version fallback is retained. `docs/proto.md` now explicitly requires planner/finalizer executors and the control plane to be cut over together after existing leases drain, and states that rolling or split deployments across these plan revisions are unsupported.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
mcusack-floe: Fixed in 0a881b95f by removing payload reconstruction and equality checking entirely. The leased finalizer no longer reads or parses reusable bundle payloads and no longer builds payload-derived metadata maps. It validates the compact manifest descriptors structurally, including content-addressed URI/digest agreement and one-to-one ownership by a staged prepared file group.

mrouvroy-floe (commented):
**[major] Missing or invalid finalized reuse manifest silently falls back to cold planning** — `reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotPlanningReconcileExecutor.java:678`

`loadReuseManifest` returns `Optional.empty()` when the selected snapshot's reuse manifest blob is missing, returns `null`, or parses as a manifest with `reusable_artifact_bundles_complete == false`. Because the basis came from `latestReconciledSnapshotForReuse`, this is not the same as having no reusable basis — it is a root-published finalized snapshot whose immutable reuse ref cannot be validated.

Treating this as an empty result hides corruption or GC loss behind a silent cold rescan, and reintroduces the hidden empty fallback this PR states it intentionally removes.

Fail closed once a finalized reuse basis has been selected: a missing or temporarily invisible blob should surface as a retryable storage failure, while malformed or incomplete validated content should become a terminal/invalidation error, not a transparent fallback.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
```diff
@@ -344,7 +376,90 @@ private SnapshotCaptureManifest validateManifestObject(
...
+          || !bundledStatsPrefixes.add(statsPrefix)
+          || artifact.getPayloadBytes() <= 0L
+          || artifact.getPayloadSha256().size() != 32
+          || !ReusableArtifactBundleUris.matchesDigest(
+              artifact.getPayloadUri(), artifact.getPayloadSha256().toByteArray())
+          || !bundleUris.add(artifact.getPayloadUri())) {
+        throw new IllegalArgumentException(
+            "snapshot capture manifest reuse bundle identity mismatch");
+      }
+      for (var metadata : bundle.getFileStatsList()) {
```
mrouvroy-floe: [major] **Bundle contents are never verified against manifest metadata before root-publishing**

`validateReusableArtifactCoverage` checks bundle URI/digest shape and confirms that each manifest metadata entry has nonblank `file_path`, fingerprint, and capture-signature fields, but the service never reads the accepted `ReusableArtifactBundlePayload` before root-publishing the manifest.

A bad executor can therefore publish compatibility metadata that does not describe the records inside the content-addressed bundle. Future planning trusts this manifest metadata when selecting reuse; the next worker will eventually reject incompatible or missing records, but by then the service has already made a bad finalized reuse basis discoverable.

During file-group acceptance or snapshot finalization, load each bundle by its digest-bound URI and compare the manifest metadata and target IDs against the actual bundle contents before accepting the submission.

---

docs/reconciler.md
```diff
@@ -106,6 +107,16 @@ Internally, the worker poller exposes `pollEvery` via `@Scheduled` (default ever
 - **File-group execution**:
   - `EXEC_FILE_GROUP` resolves the parent snapshot plan, captures file-target stats, and records
     per-file execution results on the child job payload.
+  - Each file execution plan carries connector-provided physical `content_identity`, source
+    fingerprints, stats/index capture signatures, auxiliary-target fingerprints, and any selected
+    reusable bundle records. Empty physical content identity disables cross-snapshot reuse for that
+    file.
+  - Snapshot planning reads the compact capture manifest from the most recently reconciled snapshot
+    with finalized reusable artifacts, ordered by durable `ingested_at` rather than snapshot id or
```
mrouvroy-floe: [minor] **Reuse-basis ordering documented as `ingested_at` no longer matches implementation**

The docs say snapshot planning reads the manifest from the most recently reconciled snapshot ordered by durable `ingested_at`. That no longer matches this PR's contract or implementation: reuse selection goes through `GetLatestFinalizedSnapshotForReuse` and the root-published reusable candidate list, bounded to finalized candidates at or before the committed current snapshot.

Keeping the `ingested_at` wording can send implementers toward a different compatibility basis than the service actually exposes.

Update this paragraph to describe selecting the newest root-published finalized reusable candidate at or before the committed current snapshot, and remove `ingested_at` as the ordering key.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
mcusack-floe: Addressed in 817954f6c by making the intended trust boundary explicit in the PR description, docs/proto.md, and docs/reconciler.md. The proposed control-plane bundle GET is deliberately not added: compatibility metadata is executor-authored publication metadata, and the service validates its structure, leased ownership, counts, content-addressed descriptor, and staged target mappings. The consuming file-group worker loads the selected bundle, verifies byte length and SHA-256, and rejects absent or incompatible records before reuse. The same commit also makes a selected root-published reuse manifest fail closed rather than silently cold-planning when its manifest is unavailable or invalid.

docs/reconciler.md
mcusack-floe: Fixed in 817954f6c. docs/reconciler.md now describes selection as the newest root-published finalized reusable candidate at or before the committed current snapshot. It explicitly states that candidate selection is bounded by the table root and does not scan snapshot history; the stale ingested_at ordering language is removed.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/StandaloneJavaFileGroupExecutionRunner.java
```diff
@@ -305,6 +306,9 @@ private Map<String, LoadedReuseBundle> loadReuseBundles(
         throwIfCancellationRequested(stop);
       } catch (java.util.concurrent.CancellationException error) {
         throw error;
+      } catch (StorageAbortRetryableException error) {
+        throw retryableReuseBundleRead(
+            "Failed to load reusable artifact bundle " + selection.payloadUri(), error);
       } catch (ReconcileFailureException error) {
         throw error;
       } catch (Exception error) {
```
mrouvroy-floe, line 320: [major] **Missing reuse-bundle object classified TERMINAL instead of retryable**

`readReference` calls `blobStore.get(uri)`, which for the S3 backend throws `StorageNotFoundException` when the object is absent, and otherwise raises an `IllegalStateException` when the store returns `null`.

In `loadReuseBundles`, the catch chain only special-cases `StorageAbortRetryableException`. Because `StorageNotFoundException` is a *sibling* of `StorageAbortRetryableException` (both extend `StorageException`), a not-found bundle falls through to the generic `catch (Exception)` and is wrapped by `invalidReuseBundle(...)` — `RetryDisposition.TERMINAL`, `RetryClass.NONE`. This directly contradicts the planner's `loadReuseManifest` intent, which treats a missing bundle as a retryable condition, not a permanent failure.

Add an explicit catch for `StorageNotFoundException` before the generic handler and map it to a retryable disposition, consistent with how `loadReuseManifest` handles transient storage unavailability.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotPlanningReconcileExecutor.java
```diff
@@ -660,53 +659,41 @@ private List<ReconcileFileExecutionPlan> immutablePlans() {
...
       throw invalidReuseManifest("snapshot reuse manifest metadata mismatch: " + uri, null);
     }
     try {
       SnapshotCaptureManifest manifest = SnapshotCaptureManifest.parseFrom(bytes);
       validateReuseManifestIdentity(tableId, snapshotId, expectedConnectorId, manifest, uri);
       if (!manifest.getReusableArtifactBundlesComplete()) {
-        return Optional.empty();
+        throw invalidReuseManifest("snapshot reuse manifest is incomplete: " + uri, null);
       }
       return Optional.of(new HistoricalArtifacts(manifest.getReusableArtifactBundlesList()));
```
mrouvroy-floe, line 700: [major] **Selected reuse manifest not structurally validated before planning**

`loadReuseManifest` only validates manifest identity and `reusable_artifact_bundles_complete` before returning the selected bundle list. A parseable manifest can still contain malformed reusable bundle refs — missing `artifact`, blank `payload_uri`, invalid digest/byte metadata, or blank compatibility metadata.

This selected basis then flows into `RemoteFileArtifactReusePlanner`, which silently skips bundles without a usable artifact and treats missing metadata as no match. A broken root-published reuse basis therefore degrades into partial or cold planning instead of the fail-closed terminal behavior that is expected once a basis has been selected.

Validate the reusable bundle section in `loadReuseManifest` before returning it, mirroring the structural checks the finalizer applies to bundle descriptors and metadata.

---

service/src/main/java/ai/floedb/floecat/service/repo/impl/SnapshotRepository.java
```diff
@@ -353,6 +387,35 @@ public Optional<Snapshot> getCommittedCurrentSnapshot(ResourceId tableId) {
...
+        continue;
+      }
+      Optional<Snapshot> candidate =
+          getByBlobUri(entry.getSnapshotRef().getUri()).filter(Snapshot::hasReuseManifestRef);
```
mrouvroy-floe: [major] **Broken root reuse candidate silently falls through to cold plan**

`getLatestFinalizedSnapshotForReuse` iterates `reusable_snapshot_candidates` but silently skips a non-excluded root-published candidate when its `snapshot_ref` is missing/unreadable or the loaded snapshot lacks `reuse_manifest_ref`. It can then return an older snapshot or `Optional.empty()`.

A broken published candidate becomes indistinguishable from having no reusable candidate at all. Because the planner only fail-closes after a basis is selected, this service-side fallback can convert root-index corruption or an incomplete candidate entry into stale reuse or a cold plan — without any signal to the caller.

Once a non-excluded `reusable_snapshot_candidates` entry is found, surface the broken candidate rather than continuing iteration: a missing blob or unreadable ref should surface as a retryable error, and a candidate without a valid `reuse_manifest_ref` should be treated as a terminal structural failure.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
```diff
@@ -574,21 +693,26 @@ private void publishCaptureArtifacts(
...
       }
+      if (matchedPrefix == null) {
+        throw new IllegalArgumentException(
+            "snapshot reusable bundle does not match a staged file group");
+      }
+      stagedArtifactPrefixes.remove(matchedPrefix);
     }
     if (!storedFileGroups.isEmpty()
+        || !stagedArtifactPrefixes.isEmpty()
         || declaredFileStats < manifest.getFileStatsRecordCount()
```
mrouvroy-floe, line 806: [major] **Underreported file stats allow incomplete reuse manifest to pass finalization**

`publishCaptureArtifacts` only rejects a manifest when `declaredFileStats < manifest.getFileStatsRecordCount()`. A buggy executor can therefore lower `file_stats_record_count` to match omitted accepted file-stat targets in `reusable_artifact_bundles`, pass `validateReusableArtifactCoverage`, and still have the snapshot recorded with `reusable_artifact_bundles_complete = true`.

Future planners then encounter an incomplete reuse manifest and silently recapture the omitted files, defeating the reuse guarantee without any error signal at finalization time.

Require the finalized manifest to account for all staged file-stat mappings. If duplicates across file groups must be tolerated, compare against the deduplicated staged set rather than the raw declared count.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/StandaloneJavaFileGroupExecutionRunner.java
mcusack-floe: Fixed in 6b39c3037. `loadReuseBundles` now classifies both `StorageNotFoundException` and `StorageAbortRetryableException` as RETRYABLE / TRANSIENT_ERROR. The null-return path in `readReference` now raises `StorageNotFoundException` too, so an absent or not-yet-visible bundle is retryable for both supported BlobStore behaviors. Tests cover the thrown-not-found and null-return cases.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotPlanningReconcileExecutor.java
mcusack-floe: Fixed in 6b39c3037. Publisher and planner now share `ReusableArtifactManifest.validate`. Before planning can use the selected basis it verifies the complete marker, one-to-one file-group/bundle identity, content-addressed URI/digest/size, unique bundle ownership, required compatibility metadata and selectors, and exact declared coverage. A malformed selected bundle reference now produces a terminal invalid-reuse-manifest result instead of partial or cold planning.

service/src/main/java/ai/floedb/floecat/service/repo/impl/SnapshotRepository.java
mcusack-floe: Fixed in 6b39c3037. The first non-excluded root-published candidate is now the selected basis and cannot fall through to an older candidate or cold planning. A missing/unreadable root or selected snapshot blob is retryable; an incomplete candidate, identity mismatch, invalid reuse ref, or stats-generation mismatch is structural corruption. The gRPC backend preserves that distinction by making structural reuse-selection responses terminal while leaving ABORTED transient. Repository and backend tests cover both paths.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
mcusack-floe: Fixed in 6b39c3037. Finalization now loads the durable file-group plan before the commit fence and requires exact deduplicated stats/index path coverage, including on-disk Delta DVs and Iceberg delete files. It also requires each bundle metadata entry to belong to its durable group and recomputes `ArtifactReferenceDigest` from planned target IDs plus the submitted content-addressed bundle descriptor, matching it to the accepted file-group digest. Underreporting, swapping bundle metadata, or substituting another bundle descriptor is rejected while legitimate cross-group deduplication remains valid.

service/src/main/java/ai/floedb/floecat/service/repo/impl/SnapshotRepository.java
mcusack-floe: The earlier reply is superseded by the fail-closed contract in 6b39c3037. The retained predecessor is used only when the newest candidate is explicitly excluded by replay. Once a non-excluded candidate is selected, a missing blob is retryable and a malformed/incomplete candidate is terminal; the lookup never silently advances to an older basis or returns empty.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
mcusack-floe: Strengthened in 6b39c3037 without adding control-plane bundle GETs. Finalization now binds each submitted bundle descriptor to the service-accepted file-group mappings by recomputing the accepted artifact-reference digest from the durable plan, verifies per-group metadata ownership, and enforces exact aggregate coverage. The consuming worker remains responsible for loading the content-addressed payload and validating its bytes and records.

mrouvroy-floe (commented):
**[blocker] `listByPrefixWithKeys` caches bundle projections under bare URI key** — `service/src/main/java/ai/floedb/floecat/service/repo/util/BaseResourceRepository.java:482`

In `listByPrefixWithKeys`, the batch probe (`getAllPresent(uris)` and `cached.get(blobUri)`) keys cache hits by bare URI, and the write-back at line 482 does `blobCache.put(blobUri, parsed)` — where `parsed` is the pointer-specific projection returned by `parseReferencedBlob(row.getKey(), blobUri, bytes)`. Multiple pointers legitimately share one bundle blob URI, each resolving to a distinct record (e.g. `TargetStatsStorage.targetFromBundle` returns the single `TargetStatsRecord` matching the caller's `pointerKey`).

This is wrong in two ways: (1) after the first list call caches one projection under the shared URI, every subsequent pointer to that bundle receives the same stale record; (2) if a prior `read` call populated the cache via `blobCache.getProjection(blobUri, pointerKey, ...)` using a `(uri, pointerKey)` compound key, the list path writes back a bare-URI entry with the wrong type (`TargetStatsRecord` instead of `ReusableArtifactBundlePayload` or vice versa), corrupting future reads.

The list path must be updated to mirror the single-key read path: probe via `getProjection(blobUri, row.getKey(), ...)` and write back with the same compound key, not the bare URI.

**[minor] `getAll` silently bypasses projection entries for bundle URIs** — `service/src/main/java/ai/floedb/floecat/service/repo/cache/ImmutableBlobCache.java:216`

`getAll` builds lookup keys with `CacheKey::blob` (projection `null`), so it can never hit entries populated by `getProjection`, which stores under `(uri, pointerKey)` compound keys. For bundle-backed target stats the individual read path warms only projection entries, meaning any batch or prefetch consumer going through `getAll` will always miss and re-fetch from storage.

This is correctness-safe in isolation — a miss just reloads — but combined with the `listByPrefixWithKeys` write-back bug above it means the batch path is both semantically incorrect and entirely un-cached for the reuse path this PR introduces.

`getAll` (and any other bulk accessor) should be made projection-aware: accept an optional pointer-key parameter or a `Map<URI, pointerKey>` so it can resolve compound keys, rather than silently falling back to bare-URI lookups that will never hit.

---

mcusack-floe:
Addressed the open review findings and the ci-localstack flake in bb2f3dafb.

- Blocker — bundle-backed rows are now cached and bulk-probed by (blob URI, pointer key), so distinct records projected from one shared immutable bundle cannot collide. The list path still uses URI-only entries for ordinary blobs and deduplicates backend fetches for shared bundle misses.
- Minor — ImmutableBlobCache now exposes explicit projection-aware bulk probe and population methods. Regression coverage verifies that two projections of one URI stay distinct and do not leak into the bare-URI cache; the repository test exercises repeated list reads followed by individual reads.
- ReconcileJobGcTest flake — the deadline transition is now triggered after the first retained marker is actually consumed, rather than by a fragile wall-clock/call-count race. The resume half uses a fixed clock so a 1 ms slice cannot expire before the quarantine scan on slower LocalStack CI.

Validation:
- make fmt
- Focused unit suite: 106 tests, 0 failures/errors
- Broad unit reactor: green (service: 1,754 tests, 0 failures/errors, 1 skipped; remaining modules green)
- make test-localstack TEST=ReconcileJobGcTest: 35 tests, 0 failures/errors
- git diff --check

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/StandaloneJavaFileGroupExecutionRunner.java
```diff
@@ -305,6 +307,9 @@ private Map<String, LoadedReuseBundle> loadReuseBundles(
         throwIfCancellationRequested(stop);
       } catch (java.util.concurrent.CancellationException error) {
         throw error;
+      } catch (StorageAbortRetryableException | StorageNotFoundException error) {
```
mrouvroy-floe: [minor] **`loadReuseBundles` classifies transient storage errors as TERMINAL**

`loadReuseBundles` only treats `StorageAbortRetryableException` and `StorageNotFoundException` as retryable; every other exception falls through to `catch (Exception)` and is wrapped by `invalidReuseBundle(...)` as `RetryDisposition.TERMINAL` / `RetryClass.NONE`.

`S3BlobStore.get()` maps a non-404, non-5xx GET failure to a base `StorageException` (e.g. a 403 during IAM/credential propagation, a throttle surfaced as 4xx, or a 400), and maps 409/412 to `StorageConflictException`/`StoragePreconditionFailedException` — none of which are subtypes the runner catches as retryable. All of these therefore become a permanent terminal failure of an otherwise-valid file-group capture, contradicting the PR's stated contract that missing or temporarily invisible objects should be retried.

Expand the retryable catch set to include base `StorageException` subclasses that indicate transient conditions, mirroring the same split applied in `loadPlanJobs`.

---

mrouvroy-floe (commented):
**[major] Transient storage failures in `loadPlanJobs` are terminalized instead of retried** — `reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/SnapshotPlanBlobStore.java:183`

`loadPlanJobs` wraps every failure from `blobStore.get(effectiveSnapshotPlanUri)` and JSON decoding into `IllegalStateException`, including `StorageNotFoundException` and a temporarily-invisible `null` blob read.

`RemoteSnapshotFinalizeReconcileExecutor` catches `IllegalStateException` as a terminal snapshot-finalize failure, so a transiently missing immutable snapshot plan manifest can permanently fail finalization even though the same object may become visible on a later attempt.

Split the load path by failure class: missing or not-yet-visible plan blobs should propagate as a retryable reconcile/storage failure, while malformed JSON or structurally invalid plan content should remain terminal.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/StandaloneJavaFileGroupExecutionRunner.java
mcusack-floe: Fixed in 2bc9e6796. `loadReuseBundles` now classifies `StorageAbortRetryableException`, `StorageNotFoundException`, `StorageConflictException`, and `StoragePreconditionFailedException` as RETRYABLE / TRANSIENT_ERROR. `S3BlobStore` now also normalizes HTTP 408 and 429 (along with existing 5xx, SDK-client, and closed-pool failures) to `StorageAbortRetryableException`. Generic `StorageException` remains terminal deliberately so permanent 400/403 request or authorization failures are not retried indefinitely, consistent with `ReconcileFailureClassifier`. Regression coverage verifies both structured retryable read failures and the terminal generic-storage boundary.

mcusack-floe:
Fixed the latest snapshot-plan loading review finding in 2bc9e6796. `SnapshotPlanBlobStore.loadPlanJobs` now separates object retrieval from decoding: a missing/null blob, an existing retryable storage error, or a read conflict/precondition failure surfaces as `StorageAbortRetryableException`; malformed or structurally invalid plan content and generic permanent storage failures remain terminal. `RemoteSnapshotFinalizeReconcileExecutorTest` verifies that temporary plan unavailability reaches the finalizer as RETRYABLE / TRANSIENT_ERROR. Focused validation after `make fmt`: 43 tests, 0 failures/errors.

reconciler/src/main/java/ai/floedb/floecat/reconciler/jobs/ReusableArtifactManifest.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+    if (reusableFileStats.size() != manifest.getFileStatsRecordCount()
+        || reusableIndexArtifacts.size() != manifest.getIndexArtifactCount()) {
+      throw new IllegalArgumentException(
+          "snapshot capture manifest reuse bundle coverage mismatch");
+    }
+    return new Coverage(Set.copyOf(reusableFileStats), Set.copyOf(reusableIndexArtifacts));
+  }
+
+  private static boolean hasInvalidSelectors(List<String> selectors) {
```
mrouvroy-floe, line 116: [minor] **`hasInvalidSelectors` accepts empty selector lists, allowing degenerate index entries**

`hasInvalidSelectors` returns `false` (i.e., valid) when the selector list is empty. As a result, a reusable index-artifact entry with zero `realized_index_selectors` passes structural validation at line 101.

This shared validator is the trust boundary that both the planner (`RemoteSnapshotPlanningReconcileExecutor.loadReuseManifest`) and the finalizer (`LeasedSnapshotFinalizeExecutionService`) rely on to accept executor-authored compatibility metadata without reading bundle payloads. The documented contract (see `docs/rust-remote-capture-executor.md`) requires every reusable index entry to declare a non-empty, sorted, distinct selector set matching its wrapper, and default index selection must resolve to a non-empty set for non-empty snapshots.

An index entry claiming coverage for zero columns is a degenerate reuse record that should be rejected. Add an explicit empty-list check inside `hasInvalidSelectors` so that an empty selector list is treated as invalid.

---

mrouvroy-floe (approved):
**[major] Stats selector coverage not verified against policy requirements** — `service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java:620`

`validateRealizedStatsSelectors` checks for duplicates, blank selectors, and the default-column count limit, but never verifies that the explicit stats selectors requested by the leased `ReconcileCapturePolicy` are actually present in `manifest.getRealizedStatsSelectorsList()`.

The index finalization path does this correctly via `policy.selectorsForIndex()` and `realized.containsAll(required)`, but stats finalization accepts any non-empty realized set with a sufficient column count. This means a finalized, root-published reuse manifest can claim compatibility for entirely wrong column-stats capture — e.g., a request for `customer_id` can finalize with `realized_stats_selectors=["region"]`. `ReconcileSnapshotContentState.validateMaterializedStatsCoverage` only compares realized selector cardinality for explicit stats selectors, so the mismatch propagates silently.

Add a containment check in `validateRealizedStatsSelectors` mirroring the index path: collect required selectors from the policy, then assert `realized.containsAll(required)` before accepting the manifest.

**[minor] Generic repository depends on reuse-bundle URI semantics** — `service/src/main/java/ai/floedb/floecat/service/repo/util/BaseResourceRepository.java:179`

`BaseResourceRepository` imports `ReusableArtifactBundleUris` and branches on the reconciler-specific `reuse-bundles` URI convention when selecting cache keys for reads and list scans. This pushes artifact-reuse semantics into the generic repository/cache layer, meaning every subclass now implicitly carries knowledge of one feature's blob format.

This violates the abstraction boundary: the base class should not know whether a pointer resolves to a shared bundle projection. The logic belongs at the concrete repository boundary that has that context.

Introduce an overridable hook such as `cacheProjectionKey(pointerKey, blobUri)` or `usesPointerProjection(pointerKey, blobUri)` in the base class. `StatsRepository.TargetStatsStorage` can then opt into projection-key semantics, while ordinary repositories continue using URI-only cache keys without being affected.

---

mcusack-floe:
Addressed the remaining major stats-selector coverage finding in 2e23cd771.

`LeasedSnapshotFinalizeExecutionService.validateRealizedStatsSelectors` now derives the explicitly required set from `policy.selectorsForStats()` and rejects a non-empty capture unless the realized selectors contain every requested selector. This mirrors the existing index-selector validation and prevents a same-cardinality substitution such as requesting `customer_id` while reporting `region`.

The regression test now exercises that exact mismatch and verifies rejection before the finalize commit/publication boundary.

Validation:
- `make fmt` passed (0 files reformatted)
- `mvn -pl service -Dtest=LeasedSnapshotFinalizeExecutionServiceTest -Dsurefire.failIfNoSpecifiedTests=false test`: 25 tests, 0 failures/errors
- `git diff --check` passed
- No integration or smoke tests were run.

PR #430, opened 2026-08-01
Rescan PR4: bound CAS marking memory and resume GC passes

mcusack-floe:
## Summary

Fix CAS GC heap exhaustion by removing account-wide retained blob-key sets and clarifying blob-family ownership.

- Make `StatsRepository` the sole collector for generation-owned target-stats records, manifests, index wrappers, worker uploads, and finalizer outputs.
- Keep shared index sidecars under generic CAS GC, while discovering their references through exact generation identities rather than scanning the broad target-stats tree again.
- Replace the account-wide and settled re-mark `HashSet<String>` instances with configurable, fixed-size, per-epoch Bloom reference indexes. Capacity exhaustion fails closed.
- Process generic table data at table scope and add native common-prefix table discovery for pointerless/orphaned tables.
- Enforce the scheduler deadline across paginated mark, listing, sweep, generation cleanup, deferred-delete, and re-mark work, retaining one resumable local epoch.
- Preserve version-targeted deletes, pass-frozen minimum age, owner/pin rechecks, and exact post-delete corruption verification.

Before this change, mark memory was proportional to every referenced key in an account and the observed set retained 5.89 GB. With the default capacity and false-positive rate, each Bloom index uses approximately 8 MiB; the worst active phase uses approximately 32 MiB of filter storage plus bounded pages and small exact identity sets.

## Type of Change

- [x] fix (bug fix)

## Testing

Validated with:

```text
mvn -pl service -Dtest=BloomReferenceIndexTest,CasBlobGcTest,CasBlobGcSchedulerTest,StatsRepositoryTargetStorageTest,KeysTest -Dsurefire.failIfNoSpecifiedTests=false test
```

Result: 132 selected service tests and 10 storage-SPI tests passed with zero failures.

Smoke and integration tests were not run.

- [ ] `make fmt`
- [ ] `make verify`

- [ ] No breaking changes

The legacy non-generation target-stats fallback is intentionally removed. Existing blobs in that layout will no longer be collected and require a one-off cleanup if reclaiming that storage matters.

New optional configuration, with defaults:

- `FLOECAT_GC_CAS_REFERENCE_INDEX_EXPECTED_CAPACITY=2000000`
- `FLOECAT_GC_CAS_REFERENCE_INDEX_FALSE_POSITIVE_RATE=0.0000001`

## Checklist

## Additional Notes

- This PR is stacked on `mc-scan-reuse-finalization`.
- Bloom false positives only retain garbage; false negatives are not permitted. Saturation and estimated false-positive probability are reported separately.
- Losing local continuation state on restart abandons the incomplete epoch and starts a new mark before any further deletion.
- S3 bucket versioning and lifecycle expiration of noncurrent versions remain required.


service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
```diff
@@ -508,121 +943,349 @@ public Result runForAccount(String accountId, long deadlineMs) {
...
+    rescued += snapshots.rescued();
+
+    for (String prefix :
+        List.of(
+            Keys.tableRootBlobPrefix(accountId, tableId),
+            Keys.tableIndexSidecarBlobPrefix(accountId, tableId))) {
+      boolean more;
+      do {
+        if (continuation.deferredPage == null) {
+          var deferred = new ArrayList<DeferredCandidate>(pageSize);
```
mrouvroy-floe: [major] **Deferred ownerless candidates discarded when deadline splits listing**

`deferred` is a local variable only captured into `continuation.deferredPage` after `deleteUnreferenced` returns normally. However, `deleteUnreferenced` persists its resumable page cursor into `continuation.blobTokens` after every fully-scanned page, then throws `DeadlineReached` the moment the scheduler budget is hit.

When that deadline throw happens mid-listing, the exception unwinds past the `continuation.deferredPage = new DeferredPageState(...)` assignment, silently discarding every ownerless candidate accumulated so far — yet `blobTokens` has already advanced past all pages those candidates came from. On resume, `deleteUnreferenced` restarts from the stored token and never re-lists those pages, so the candidates are permanently lost and their blobs leak.

Persist accumulated `deferred` candidates into `continuation.deferredPage` before each `checkDeadline()` call, or catch `DeadlineReached` locally, flush `deferred` into the continuation, then rethrow.

---

service/src/main/java/ai/floedb/floecat/service/repo/impl/StatsRepository.java
```diff
@@ -1643,29 +1874,80 @@ public GenerationGcResult deleteUnreferencedGenerations(
...
+        // are not object-only crash residue and remain owned by their explicit publisher cleanup.
+        continuation.candidateIndex++;
         continue;
       }
-      // One HEAD answers both existence and age: absent means unpublished (an in-flight replace
-      // writes records before publishing) or already reclaimed.
+      // Prefer the manifest as the age fence. Legacy/crashed remote work can have upload objects
+      // but no manifest or lifecycle marker, so use an upload object as the repair-scan anchor.
       var header = blobStore.head(manifestUri).orElse(null);
+      if (header == null && orphanWithoutLifecycle) {
```
mrouvroy-floe: [major] **`WRITING` generations retain abandoned uploads on worker crash**

`LeasedFileGroupExecutionService.resolve` now calls `beginStatsGeneration` before returning the upload prefix, so a full-rescan worker that crashes before publishing a manifest leaves a `WRITING` lifecycle row plus objects under `worker-uploads/`.

The orphan cleanup path only uses an upload or finalizer object as the age anchor when `orphanWithoutLifecycle` is true, so the reserved-but-never-published generation reaches `header == null` and is skipped. Because generic CAS intentionally excludes `/target-stats/`, nothing else collects those generation-owned objects. The lifecycle reservation makes the crash residue discoverable, but the cleanup path then treats it as non-orphan and leaves it forever.

Treat stale `WRITING` generations with no manifest like the orphan repair case: after the manifest grace period expires, delete the generation row and any objects under its upload prefix.

---

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
```diff
@@ -508,121 +943,349 @@ public Result runForAccount(String accountId, long deadlineMs) {
...
+      }
+      if (!state.remarkComplete) {
+        if (!remarkTable(accountId, tableId, state.fresh, pageSize)) {
+          walkFailures[0]++;
+          return new DeleteResult(0, 0, 0, true);
+        }
+        state.remarkComplete = true;
+      }
+      var guarded =
+          reachabilityGuard.deleteIfUnchanged(
```
mrouvroy-floe: [minor] **Blob I/O runs inside reachability-guard lock, can stall publisher hot path**

`deleteIfUnchanged` executes the entire deletion closure while holding `proof.entry.lock`, and that closure performs remote I/O per candidate: `keepForLatePin` -> `rootLivePinChains` -> `rootTableRootChain` (root/manifest-page blob reads) plus `blobStore.delete`/`blobStore.head`. The loop is bounded only by `checkDeadline()`, so a single invocation can hold the table's entry lock for close to a full `max-tick-millis` slice (default 45s).

The same per-table `Entry.lock` is taken by every publisher for that table — `QueryContextStoreImpl.registerResolvingPinBlobs`, `IndexArtifactRepository`, and `TableRootCommitter.commit` all go through `reachabilityGuard.publishing(...)`. Pin registration is on the query resolution hot path, so a large deferred flush can stall commits and queries for the duration of the GC tick.

Perform blob reads outside the lock, capturing only the data needed for the reachability decision, then take the lock briefly to apply the deletion decision.

---

mrouvroy-floe (commented):
**[major] Account discovery runs outside tick-budget deadline, can exhaust `max-tick-millis`** — `service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGcScheduler.java:174`

`fetchAllAccounts` runs before any deadline check and follows every account page until the repository is exhausted. With a large account set or a slow account repository, this call can consume or exceed the entire `floecat.gc.cas.max-tick-millis` budget before the scheduler reaches the line 190 guard or any resumable CAS work.

This breaks the tick-budget guarantee for the scheduler that now resumes mark/list/sweep work under a deadline. It can delay shutdown, starve a retained continuation, and make `max-tick-millis` misleading on deployments with many accounts.

Carry the deadline and `stopping` flag into account discovery, processing account pages incrementally under the same tick budget. If discovery stops mid-list, retain enough account cursor state to resume fairly on the next tick instead of rebuilding from scratch.

---

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
mcusack-floe: Fixed in 75ed345d8. Deferred ownerless candidates are now transferred one listing page at a time into `continuation.deferredPage` before another deadline can be observed. The resumable blob token still advances after the page is fully scanned, but `deleteUnreferenced` returns immediately for a deferred page, so the page state is durable before control reaches another deadline check. This also bounds the retained candidate list by `pageSize`. Added `deferredPageIsRetainedBeforeDeadlineAdvancesPastIt` to cover resume after the token has advanced.

service/src/main/java/ai/floedb/floecat/service/repo/impl/StatsRepository.java
mcusack-floe: Confirmed this is handled by the supported full-rescan job lifecycle rather than generic age-based GC. Terminal failed/cancelled `PLAN_SNAPSHOT` jobs call `cleanupAbandonedFullRescanStatsGenerationIfTerminal`, which derives `full-rescan-<jobId>` and invokes `deleteUnpublishedStatsGeneration`; `runAbandonedFullRescanStatsCleanupMaintenanceOnce` retries incomplete cleanup from durable job state. That path safely distinguishes an abandoned generation from a legitimate long-running `WRITING` generation, so I have not added an age-only sweep.

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
mcusack-floe: Fixed in 75ed345d8. `keepForLatePin` and the other remote reachability reads now run outside the publication lock. The code captures a reachability proof/epoch and uses `deleteIfUnchanged` only around the single immutable-version delete; an overlapping publisher invalidates the proof and restarts the bounded re-mark. Added `deferredFlushLocksAroundOnlyOneVersionDeleteAtATime` to verify the critical section never spans a candidate page.

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
```diff
@@ -752,44 +1546,185 @@ private int collectPointers(
...
+        if (storageEstimate != null) {
+          storageEstimate.observe(pointer);
+        }
+        try {
+          byte[] bytes = blobStore.get(pointer.getBlobUri());
+          if (bytes == null) {
+            throw new IllegalStateException(
+                "missing index artifact wrapper " + pointer.getBlobUri());
+          }
+          if (pointer.getBlobUri().contains("/reuse-bundles/")) {
```
mrouvroy-floe: [minor] **Fragile `"/reuse-bundles/"` string coupling for payload-type dispatch**

The dispatch between parsing a blob as `ReusableArtifactBundlePayload` vs a single `IndexArtifactRecord` keys off the literal `pointer.getBlobUri().contains("/reuse-bundles/")`. The producing side independently builds these URIs with the raw literal `getStatsObjectPrefix() + "reuse-bundles/"`. There is no shared constant tying the two together, unlike the `Keys.SEG_*` segments used elsewhere.

If either side's path segment ever changes, GC will silently misclassify: a bundle blob parsed as a bare `IndexArtifactRecord` (or vice versa) throws `IllegalStateException` and poisons the account's sweep, or worse, quietly fails to expand shared-sidecar references.

Extract a single shared constant (e.g. in `Keys`) used by both the writer and this dispatch, so a rename is a one-line change with a compile-time guarantee rather than a silent runtime misclassification.

---

mrouvroy-floe (commented):
**[major] Account discovery runs outside the tick deadline and stopping check** — `service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGcScheduler.java:174`

`fetchAllAccounts` fully paginates the account repository before the tick's `deadline` (line 168) or the `stopping` flag is ever consulted. The per-account budget check only starts after this call returns, so with a large account set or a slow account repository, discovery alone can consume — or exceed — the entire `floecat.gc.cas.max-tick-millis` budget (default 45 s).

When that happens, the per-account loop breaks on its very first iteration because `System.currentTimeMillis() >= deadline` is already true, so the tick performs no mark/list/sweep/continuation work at all. This repeats every tick, starving GC and letting the backlog-age metric climb without cause. It also blocks a clean shutdown because `stopping` is never checked during discovery, undermining the PR's resumability guarantee.

Make account discovery itself deadline-aware and resumable: paginate accounts incrementally under the same deadline and `stopping` flag, and retain the continuation token across ticks so that a partial account traversal can be resumed on the next tick rather than restarted from the beginning.

---

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
```diff
@@ -752,44 +1546,185 @@ private int collectPointers(
...
+    // then query only each generation's index-artifact sub-prefix instead of paging millions of
+    // unrelated target-record pointers. Root walks may add a newly protected generation identity
+    // before a settled re-mark reaches this point.
+    var tableResourceId =
+        ResourceId.newBuilder()
+            .setAccountId(accountId)
+            .setId(tableId)
+            .setKind(ResourceKind.RK_TABLE)
+            .build();
+    for (Keys.GenerationKey generation : continuation.tableGenerationKeys) {
```
mrouvroy-floe, line 1773: [major] **Direct index-sidecar blobs are never added to the mark set**

`collectSharedIndexArtifactReferences` only expands wrapper pointers for generation IDs already present in `continuation.tableGenerationKeys`, which are populated from stats-generation discovery and table-root stats-generation refs.

`IndexArtifactRepository.putIndexArtifact` can publish a live `direct` index-artifact generation under `snapshotIndexArtifactGenerationPrefix(...)` without creating a stats lifecycle marker or table-root stats ref. A table whose sole active index generation is a direct one therefore has live wrapper pointers whose `IndexArtifactRecord.artifactUri` values are never expanded into the generic mark set. The table sweep subsequently visits `Keys.tableIndexSidecarBlobPrefix(...)`; because the sidecar was absent from both the original mark and the deferred re-mark passes, CAS GC can delete live sidecar blobs for those tables.

Include direct index-artifact generations in `continuation.tableGenerationKeys` (or add a dedicated scan path for them) so their sidecar blobs are protected by the mark phase.

---

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
```diff
@@ -508,121 +944,361 @@ public Result runForAccount(String accountId, long deadlineMs) {
...
+      long referenceCapacity,
+      double referenceFalsePositiveRate) {
+    DeferredPageState state = continuation.deferredPage;
+    while (true) {
+      rootLivePinChains(referenced, walkedPinRoots, walkFailures);
+      if (walkFailures[0] > 0) {
+        return new DeleteResult(0, 0, 0, true);
+      }
+      if (state.fresh == null) {
+        state.remarkProof = reachabilityGuard.beginProof(accountId, tableId);
```
mrouvroy-floe: [nit] **`beginProof` guard reference can be orphaned if Bloom allocation throws `Error`**

In `flushDeferredTableCandidates`, `state.remarkProof = reachabilityGuard.beginProof(...)` acquires a reference-counted `TableBlobReachabilityGuard.Entry` before `state.fresh = newReferenceIndex(...)` allocates the ~8 MiB Bloom filter.

If the allocation throws an `Error` (e.g. `OutOfMemoryError`), the `catch (RuntimeException)` handler in `runForAccount` does not fire, so the continuation is retained with `remarkProof` set but `fresh == null`. On the next resume the `if (state.fresh == null)` block overwrites `state.remarkProof` with a new `beginProof` call without closing the previous one, orphaning the first entry so its reference count never reaches zero and the entry lingers in the guard's map indefinitely.

The window is narrow (only an `Error` between the two lines triggers it), but the fix is straightforward: acquire the guard reference and allocate the filter atomically, or close any previously acquired proof in the `finally` / error path before overwriting `state.remarkProof`.

---

mrouvroy-floe (commented):
**[major] `fetchAllAccounts` fully paginates before tick deadline or stop flag is checked** — `service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGcScheduler.java:174`

`fetchAllAccounts(accountRepo, accountsPageSize)` runs to completion before the per-account loop begins, and neither the tick `deadline` (line 168) nor the `stopping` flag is consulted during discovery.

On deployments with many accounts or a slow account repository, discovery alone can consume or exceed `floecat.gc.cas.max-tick-millis` (default 45 s). When it does, the per-account loop breaks on its very first iteration because `System.currentTimeMillis() >= deadline` is already true, so the tick performs zero GC or continuation progress — and this repeats every tick while the backlog-age metric climbs. The `stopping` flag being ignored during pagination also delays clean shutdown.

Make account discovery itself deadline-aware and resumable: thread the deadline and stop signal into account pagination, persist the account continuation token across ticks, and process only the account page(s) that fit within the remaining budget per tick.

---

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
mcusack-floe: Fixed in 3501a4c75. `ReusableArtifactBundleUris.BUNDLE_DIRECTORY` now defines the path segment used by the remote worker producer, manifest/finalizer validation, and URI matching; CAS GC dispatch now calls `ReusableArtifactBundleUris.isBundleUri(...)` instead of carrying its own literal.

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
mcusack-floe: Fixed in 3501a4c75. Snapshot pointer scans now discover an active `direct` index generation and add its generation key before shared wrapper expansion, both in the initial table mark and the publication-guarded re-mark. Added `directIndexGenerationWithoutStatsLifecycleProtectsItsSharedSidecar`, which constructs a direct generation with no stats lifecycle row and verifies its sidecar survives GC.

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
mcusack-floe: Fixed in 3501a4c75. The Bloom reference index is now allocated before acquiring the retained reachability proof. Proof/index initialization uses locals and publishes both into continuation state only after both succeed; any failure after allocation closes both resources before rethrowing, so an allocation `Error` cannot leave or overwrite an orphaned proof.

service/src/main/java/ai/floedb/floecat/service/repo/impl/StatsRepository.java
```diff
@@ -1589,49 +1628,158 @@ public int deleteUnreferencedGenerations(
...
+
+    public List<Keys.GenerationKey> generations() {
+      return candidates;
+    }
+
+    private void discover(Keys.GenerationKey generation) {
+      if (discovered.contains(generation)) {
+        return;
+      }
+      if (discovered.size() >= maxCandidates) {
```
mrouvroy-floe, line 1702: [minor] **Per-table generation cap poisons the whole account's GC**

`GenerationGcContinuation.discover` throws `GenerationGcCapacityExceededException` once a single table exceeds `max-stats-generations-per-table` (default 100,000 lifecycle-bearing generations). The only caller catches this at the account level (`CasBlobGc.runForAccount`, `CasBlobGc.java:373`), converts it to a `poisonedResult`, and clears the epoch.

Because discovery re-runs from scratch on the next pass, the same oversized table throws again every tick — permanently stopping generation reclamation for the *entire* account even though only one table is oversized. That account's dead-generation blobs are never collected, and the blast radius is the full account rather than the offending table.

The threshold is extreme and the `CAS_OLDEST_SWEEP_AGE` gauge will surface the condition. Consider catching the exception at the table level instead, skipping or rate-limiting just the offending table while allowing the rest of the account's GC to proceed.

---

mrouvroy-floe (commented):
**[major] Account discovery runs outside the tick deadline and `stopping` check** — `service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGcScheduler.java:174`

`fetchAllAccounts(accountRepo, accountsPageSize)` fully paginates every account into an in-memory list before the per-account loop's `deadline`/`stopping` guard (line 190) is ever consulted.

On deployments with many accounts or a slow account store, discovery alone can consume or exceed `floecat.gc.cas.max-tick-millis`. When it does, the per-account loop breaks on its first iteration, the tick performs zero mark/list/sweep/continuation work, and the same full account traversal repeats on every subsequent tick while the backlog-age metric climbs. Clean shutdown is also delayed because `stopping` is not checked during pagination — directly undercutting the PR's stated deadline/resume guarantee.

Make account discovery deadline-aware and resumable: store an account continuation cursor in scheduler state, check `deadline` and `stopping` between account pages, and process only the accounts discovered within the remaining budget before yielding.

---

service/src/main/java/ai/floedb/floecat/service/repo/impl/StatsRepository.java
mcusack-floe: Fixed in 8c97ce118. `GenerationGcCapacityExceededException` is now isolated at the active table: GC marks the account poisoned/backlogged, closes the partial table state, advances the durable table cursor, and continues the same account epoch with later tables. Added `oversizedGenerationTableDoesNotStopLaterTablesInTheAccount` to verify the next table is still processed.

mcusack-floe:
Addressed the review-summary account-discovery issue in 8c97ce118. `CasBlobGcScheduler` no longer materializes every account before checking the tick budget. It retains the account page, page index, and backend cursor across ticks; checks the deadline/stopping state between page fetches and account runs; and prioritizes a retained account GC epoch via a bounded direct lookup. Full-cycle account/backlog cleanup now occurs only when the resumable cursor completes. Added `accountPageFetchedAtDeadlineIsRetainedForTheNextTick`, which forces page loading past the budget and verifies the next tick consumes the retained page without re-fetching it.

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
```diff
@@ -90,14 +100,162 @@ public class CasBlobGc {
...
+    }
+  }
+
+  private enum Phase {
+    ACCOUNT_MARK,
+    TABLES,
+    ACCOUNT_SWEEP
+  }
+
+  private static final class PassContinuation {
```
mrouvroy-floe, line 176: [major] **Monolithic GC state machine with implicit phase boundaries**

`CasBlobGc` now owns account marking, table discovery, table-scoped marking, stats-generation reclamation, shared index sidecar expansion, deferred ownerless deletion, publication-proof verification, chain-walk continuation, pointer-scan continuation, and blob-scan continuation in one 2k+ line class.

`PassContinuation` is the clearest symptom: it mixes global account state, per-table state, chain-walk tokens, deferred-delete proof state, generation cleanup cursors, metrics, and scheduling phase control. That makes this PR's central safety property hard to audit. A future change to one phase can easily invalidate another phase's checkpointing or proof invariants because the boundaries are all implicit in shared mutable fields like `pointerTokens`, `blobTokens`, `chainPageUris`, and `currentTable`.

Consider decomposing `PassContinuation` into per-phase continuation types and extracting each phase into a dedicated class with a narrow interface, so that invariants and checkpoint boundaries are explicit and independently testable.

---

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
```diff
@@ -380,20 +907,24 @@ public Result runForAccount(String accountId, long deadlineMs) {
...
-          tablesScanned,
+          referenceInsertions(pass),
+          saturationPpm(pass),
+          falsePositivePpb(pass),
+          pass.tablesScanned,
           true,
           false,
-          generationCleanupPending);
+          pass.generationCleanupPending);
     }
```
mrouvroy-floe, line 964: [minor] **Sweep counters lost on DeadlineReached during blob-family scan**

The account sweep copies `pass.blobsScanned`, `pass.blobsDeleted`, and `pass.blobsRescued` into local variables before calling `deleteUnreferenced` for each blob family. `deleteUnreferenced` can persist a resumed blob token after a page, execute version-targeted deletes, and then throw `DeadlineReached` before those local counters are copied back into `pass` or returned to the scheduler.

On the next tick the persisted token correctly resumes after the completed page — so no work is repeated — but the already-completed page's blob counts are missing from `blobs-scanned`, `blobs-deleted`, and `blobs-rescued`. This causes GC throughput telemetry to under-report exactly when the new deadline/resume path is active, making it hard to detect regressions or tune budgets.

Checkpoint the counters into `PassContinuation` after each family/page, or make `deleteUnreferenced` update a retained progress object so counters survive a thrown deadline.

---

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
```diff
@@ -295,57 +798,53 @@ public Result runForAccount(String accountId, long deadlineMs) {
...
+      // makes pin validation fail closed, while pin-first makes the protection predicate keep the
+      // generation and every shared sidecar its wrappers reference.
+      if (!statsRepository.discoverGenerationKeys(rid, deadlineMs, pass.generationGcContinuation)) {
+        checkDeadline();
+        throw DeadlineReached.INSTANCE;
+      }
+      for (Keys.GenerationKey generation : pass.generationGcContinuation.generations()) {
+        pass.addGenerationKey(generation);
       }
+      int generationDeleteBudget = remainingGenerationBlobDeletes;
```
mrouvroy-floe: [minor] **Reachability lock held across remote I/O during generation reclamation**

`deleteUnreferencedGenerations` is wrapped entirely in `reachabilityGuard.exclusive(rid, ...)`, so the table's `ReentrantLock` is held for an entire bounded delete slice — which performs a `head()` plus a version-targeted `delete()` per generation blob (up to the per-account delete budget, default 1000). Because the guard contract requires publishers and CAS GC to be co-located in one process, a concurrent `putIndexArtifact` or `registerPrewrittenIndexArtifactReferencesInGeneration` call on the same table will block on this lock for the full duration of that slice, including all network round-trips.

The deferred-sidecar flush path already applies better discipline: it serializes only the single irreversible `deleteIfUnchanged` call and keeps reachability reads outside the lock. Consider applying the same pattern here — perform the `head()` reachability checks outside the lock, then acquire it only for the irreversible delete steps — to avoid stalling publishers during large generation reclamation passes.

---

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
```diff
@@ -680,7 +1463,14 @@ private boolean rootTableRootChain(String rootBlobUri, Set<String> referenced) {
               rootBlobUri, pageRef.getUri());
           return false;
         }
```
mrouvroy-floe, line 1529: [nit] **Resumed chain page skips re-seeding the acyclic-chain cycle detector**

When `resumingPage` is true, the resumed page is intentionally not re-added to `walkedPages` (lines 1464-1465). In the rare interleaving where `continuation.chainVisited` was reset between ticks — reachable because `normalizedPinRoots()` returns an unordered `HashSet`, so a second pinned table-root sharing the `table:<tid>` scope can be walked between the interruption and the resume and trigger the reset near lines 1409-1419 — the resumed chain's earlier pages are absent from the now-empty `walkedPages`.

This only weakens the corrupt-data cycle detector: a genuinely cyclic manifest chain could then loop instead of being caught and poisoning the sweep. It cannot skip a live reference or cause data loss on well-formed (acyclic, content-addressed) chains.

Consider re-seeding `walkedPages` from `continuation.chainVisited` on resume, or adding a comment documenting why the omission is safe, so the invariant is not accidentally removed by a future refactor.

---

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
mcusack-floe: Partially addressed in `a2d28e4cf`. Pointer/blob/chain traversal cursors now live in `TraversalContinuation`, and the account-mark, table, and account-sweep stages have explicit `Phase` gates so completed stages are not replayed on resume. `PassContinuation` still retains cross-phase state and the phases are not yet extracted into dedicated classes. We assessed the full decomposition as too large and risky for this already broad safety PR, so I am leaving this thread open for a follow-up refactor.

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
mcusack-floe: Fixed in `a2d28e4cf`. `DeleteProgress` now checkpoints partial scan/delete/rescue progress when `DeadlineReached` unwinds `deleteUnreferenced`, and both table and account sweeps retain completed-family totals before propagating the deadline. Added `blobSweepCountersSurviveADeadlineAfterACompletedPage` to cover the resumed totals.

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
mcusack-floe: Fixed in `a2d28e4cf`. Generation discovery and blob `head`/delete I/O now occur outside the table reachability lock. The new claim guard serializes only the final protection recheck and lifecycle transition to `DELETING`; after that claim, the lifecycle prevents new publication while cleanup proceeds. Added `generationGcDoesNotHoldThePublicationGuardAcrossBlobIo`.

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
mcusack-floe: Fixed in `a2d28e4cf`. Pin roots are now traversed deterministically, and if traversal switches chains while another chain has a saved partial-page cursor, that cursor is discarded so the interrupted chain restarts from its root with a fresh cycle detector. Consequently, `resumingPage` is true only while the matching retained `chainVisited` detector is still active; the code now documents that invariant.

mrouvroy-floe (commented):
**[minor] Generation cursor wraps to the start table on full-pass completion, defeating budget round-robin** — `service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java:925`

After each completed table the durable generation cursor is set to `tableIds.get((tableIndex + 1) % tableIds.size())`. On the final table (`tableIndex == size - 1`) the modulo wraps back to `tableIds.get(0)`, which — because `tableIds` was rotated to `-indexOf(cursor)` at line 748 — is the same table the pass started at. Across fully-completing passes the cursor is therefore pinned to the same starting table.

Trace `[A, B, C]` starting at `A`: the pass ends with cursor at `A`, the next pass rotates by 0 and starts at `A` again, and so on. The shared per-account generation-delete budget (`remainingGenerationBlobDeletes`, default 1000, consumed in table order at lines 839/873) always favors the same early tables; a high-churn early table that saturates the budget every pass will starve later tables indefinitely.

Fix: after a fully-completed pass, advance the cursor to the table *after* the last one processed (i.e. use `(tableIndex + 1) % size` relative to the un-rotated list), or explicitly record a "pass-complete" sentinel that causes the next pass to start at the successor table.

---

service/src/main/java/ai/floedb/floecat/service/repo/model/Keys.java
```diff
@@ -1601,7 +1641,70 @@ public static GenerationKey generationFromTargetPointerKey(String pointerKey) {
...
+    int generationEnd = manifestUri.endsWith(".pb") ? manifestUri.length() - 3 : -1;
+    if (generationEnd <= generationStart || manifestUri.indexOf('/', generationStart) >= 0) {
+      return null;
+    }
+    return new GenerationKey(
+        snapshotId, percentDecode(manifestUri.substring(generationStart, generationEnd)));
+  }
+
+  /** Recovers a generation identity from a generation object-store common prefix. */
+  public static GenerationKey generationFromGenerationBlobPrefix(String generationPrefix) {
```
mrouvroy-floe: [minor] **`generationFromGenerationBlobPrefix` is dead code, leaving advertised generation-rediscovery unimplemented**

`generationFromGenerationBlobPrefix` has zero callers anywhere in the repo (production or test) — only its definition exists. Its sibling `generationFromManifestBlobUri` is correctly wired into `CasBlobGc`, `StatsRepository`, and `QueryContextStoreImpl`.

The PR description advertises "native common-prefix table discovery for pointerless/orphaned tables", and this method appears to be the object-store-side generation-discovery fallback intended to back that claim — yet nothing lists blob common-prefixes to feed it. A generation whose lifecycle pointer was lost but whose blobs survive would therefore never be rediscovered from the object store and would leak.

Either wire the method into the GC sweep (listing blob common-prefixes and feeding them here) to fulfill the advertised guarantee, or remove the dead method and update the PR description to remove the rediscovery claim.

---

mcusack-floe: Addressed in `6dcffb8f4`. Removed the unused `generationFromGenerationBlobPrefix` parser rather than introducing a second object-store generation-discovery fallback. Native common-prefix discovery remains implemented for pointerless/orphaned table IDs in `collectTableBlobIds`, which is the scope advertised by the PR summary.

mcusack-floe:
Addressed the review-summary generation-cursor fairness issue in `6dcffb8f4`. Mid-pass checkpoints still resume at the next unprocessed table, while completing a full circuit now advances the next pass to the successor of the previous starting table. The same rule is applied when the final table is skipped for generation-capacity overflow. Added `completedGenerationCleanupPassRotatesItsStartingTable`, which verifies complete passes rotate from A to B to C instead of repeatedly starting at A. ([review](https://github.com/eng-floe/floecat/pull/430#pullrequestreview-4901229830))

mrouvroy-floe (approved):

PR #431, opened 2026-08-02
Rescan PR5: reuse append-only snapshot bases

mcusack-floe:
## Summary

Adds an append-only snapshot fast path that reuses the latest finalized snapshot as an immutable capture base and plans/captures only newly appended files.

The change includes:

- A connector contract for calculating snapshot file deltas, with Iceberg and Delta implementations. Delta commit-log reads are partitioned across a bounded number of virtual-thread readers, with a non-retaining classification pass so rewrite/OPTIMIZE commits fail out before additions are materialized.
- Append-only base descriptors in the reconciler protocol and compact persisted snapshot plans that authenticate the inherited manifest, reusable artifact index, artifact counts, digest, and chain depth.
- A persistent reusable-artifact run index. Snapshot manifests contain only the current generation's inline file-group bundles; inherited coverage is read through the authenticated base index rather than copied cumulatively into every manifest.
- Local and remote finalizer support for validating exact current file-group identity across groups, rolling inherited aggregate records into the new snapshot, publishing inherited stats/index generation membership, and enforcing complete source-file coverage.
- Planner activation that selects the latest finalized reusable snapshot, verifies capture-policy and schema compatibility, requests a connector-proven append-only delta, enqueues only delta file groups, supports zero-delta snapshots, and bounds reuse chains with checkpoints. Snapshot IDs remain opaque; lineage proof never relies on numeric snapshot ordering.
- Repository overlay support for inherited stats and index generations, including lookup, listing, pagination, counting, rebinding, and GC reachability protection.
- Deferred snapshot-finalize publication backed by a durable accepted-result intent. Publication is independent of the originating worker lease, survives service restart/lease expiry, and uses epoch-guarded recovery without resurrecting stale intent payloads.
- Cached/single-flight STS credential vending keyed by authority, source credentials, and normalized scope, with freshness skew, transient retries, and `UNAVAILABLE` mapping after retry exhaustion.
- Planner observability for reuse levels plus exact encoded RPC request sizes and submission timing.

A finalized manifest without the new persistent artifact index is not eligible for reuse and safely triggers a full capture. Once an indexed contract is selected, malformed, cross-owner, or digest-invalid metadata fails closed.

This reduces planning, execution, manifest, and credential-vending overhead for large append-only tables while retaining exact validation and periodic checkpoint behavior.

## Type of Change

- [x] fix (bug fix)

## Testing

- `make proto` passed; 1,126 generated sources compiled.
- `make fmt` passed earlier in the branch validation.
- Focused snapshot planner suite: 34 tests passed.
- Focused snapshot finalizer suite: 31 tests passed.
- Review follow-up suites: `DeltaConnectorTest` 21, reconciler finalizer/index/job-store tests 81, `DurableReconcileJobStoreTest` 150, and GC/publication tests 94.
- `BoundedFanoutTest`: 36 tests passed as part of the broader changed-class reactor run.
- The remaining selected connector, reconciler, service, repository, GC, and storage-authority unit tests passed in the broader reactor run.
- `git diff --check` passed.
- No integration or smoke tests were run.

- [x] `make fmt`
- [ ] `make verify`

- [ ] No breaking changes
- [x] Breaking changes (described below)
- [x] Config/env var changes

Planner/finalizer executors and the control plane must be deployed together after existing leases drain. Split deployments across these protocol revisions are unsupported. Existing finalized manifests without a persistent reusable artifact index remain readable but are deliberately ineligible for append-only reuse and receive a full capture.

New publication controls are `floecat.reconciler.snapshot-finalize-publication.tick-every` (default `250ms`), `.page-size` (default `100`), and `.max-parallelism` (default `4`), with corresponding `FLOECAT_RECONCILER_SNAPSHOT_FINALIZE_PUBLICATION_*` environment variables. The general job-store lease default remains `120000`; reverting the temporary `600000` value is a consequence of accepted finalizer publication no longer depending on the worker lease lifetime.

## Checklist

## Additional Notes

- This is a stacked PR based on `mc-gc-oom-fix`.
- Inline `reusable_artifact_bundles` cover only current file groups; the persistent run index supplies inherited coverage.
- `ReusableArtifactIndexStore.append` now validates ownership of every inherited run/filter/manifest/block-pack object. This adds metadata reads proportional to inherited runs and their run manifests on the finalize path; it does not read or authenticate every pack payload, and the work is bounded by the persisted run/block structure.
- Successful `PLAN_SNAPSHOT` request-size/timing logging and planner reuse-level logging are included with planner activation because they share the activation hunks; the remaining submission instrumentation is isolated in the final commit.
- Rebased head: `4577dd814`.


reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotPlanningReconcileExecutor.java
```diff
@@ -661,45 +1041,151 @@ private List<ReconcileFileExecutionPlan> immutablePlans() {
...
+    try {
+      new ReusableArtifactIndexStore(blobStore)
+          .validateLookupReference(manifest.getReusableArtifactIndex());
+      if (manifest.getReusableArtifactIndex().getFileStatsRecordCount()
+              != manifest.getFileStatsRecordCount()
+          || manifest.getReusableArtifactIndex().getIndexArtifactCount()
+              != manifest.getIndexArtifactCount()) {
+        throw new IllegalArgumentException("reusable artifact index count mismatch");
+      }
+    } catch (IllegalArgumentException e) {
```
mrouvroy-floe: [major] **Invalid index fails open**

After selecting a finalized snapshot reuse basis, `loadReuseManifest` treats malformed reusable-index metadata as ineligible and returns `Optional.empty()`. That converts an indexed-but-invalid basis into a full capture at lines 1102-1113.

The PR contract says legacy manifests without the persistent index are not eligible, but once an indexed contract is selected, malformed or digest-invalid metadata must fail closed. Falling back here hides corruption or partial publication in the reusable index and lets the system proceed as though the snapshot simply lacked a reusable basis.

Only the `!manifest.hasReusableArtifactIndex()` legacy case should downgrade to a full capture. For selected indexed manifests, propagate `IllegalArgumentException` as a terminal failure and `StorageNotFoundException` as retryable/unavailable, consistent with the manifest/digest validation paths.

---

🤖 _Generated with GPT-5.5._

core/proto/src/main/proto/floecat/reconciler/reconciler.proto
```diff
@@ -374,6 +374,89 @@ message ReusableArtifactBundleSelection {
...
+
+// Size-bounded data block within a sorted run.
+message ReusableArtifactIndexBlock {
+  uint32 format_version = 1;
+  repeated ReusableArtifactIndexEntry entries = 2;
+}
+
+// Immutable predecessor used by the append-only fast path. The referenced manifest and persistent
+// reusable-artifact run index are verified before its aggregate generation is inherited.
+message AppendOnlySnapshotBase {
```
mrouvroy-floe, line 445: [major] **Missing reuse chain depth**

`AppendOnlySnapshotBase` authenticates the predecessor manifest and index counts, but it has no `chain_depth` field. The persisted mirror in `SnapshotPlanBlobStore.AppendOnlyBase` also lacks depth, and the planner builds a new base from the selected manifest without incrementing or enforcing any bound.

That misses the advertised checkpoint behavior. Long-lived append-only tables can build unbounded reuse chains, and neither the finalizer nor the persisted plan can authenticate how far the current snapshot is from a full checkpoint.

Add chain depth to the proto and persisted `AppendOnlyBase`, include it in `manifestIdentity()`, validate it in the finalizer against the base manifest, and make the planner force a full/checkpoint capture when the configured maximum depth is reached.

---

🤖 _Generated with GPT-5.5._

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReusableArtifactIndexStore.java
```diff
@@ -0,0 +1,1628 @@
...
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.PriorityQueue;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.function.Consumer;
+
+/** Persistent immutable sorted-run index for reusable file artifacts. */
+public final class ReusableArtifactIndexStore {
```
mrouvroy-floe, line 45: [major] **Oversized reusable-index module**

`ReusableArtifactIndexStore` is introduced as a 1,628-line class that owns too many independent invariants at once: sorted-run format validation, Bloom-filter encoding/parsing, object caching, pack IO/range reads, compaction, lookup, pagination, and GC reachability walking.

This code is now part of the trusted append-only reuse contract. Keeping all of those responsibilities in one class makes it very hard to review whether malformed metadata fails closed, whether lookup and pagination share the same ordering semantics, and whether GC walks authenticate the same objects that planning/finalization use.

Split the earned concepts into focused collaborators, for example a run/block codec, an object reader/cache, a run compactor, a lookup/pager, and a reachability walker. Keep `ReusableArtifactIndexStore` as the small facade that wires those pieces together and enforces the top-level contract.

---

🤖 _Generated with GPT-5.5._

mrouvroy-floe (commented):
**[major] Finalizer owns validation policy** — `reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotFinalizeReconcileExecutor.java:57`

`RemoteSnapshotFinalizeReconcileExecutor` has grown to 1,305 lines and now mixes executor control flow with append-only base authentication, aggregate inheritance, selector-policy validation, reusable-bundle validation, deduplication, and manifest preparation inputs.

Those rules are not incidental executor details; they are the safety boundary that prevents partial file-group output or stale inherited coverage from being published as a complete snapshot. Burying them inside the remote executor makes the happy path and the failure-closed checks hard to audit independently, and future changes to the executor loop can accidentally weaken validation.

Move the artifact/manifest validation and dedupe logic into focused helper classes with narrow APIs, then leave this executor responsible for orchestration: load input, validate through those helpers, prepare the publication request, and submit the terminal result.

---

🤖 _Generated with GPT-5.5._

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotPlanningReconcileExecutor.java
mcusack-floe: Intentional policy decision in 5565b02ee. This branch accepts only the format-1 manifest contract introduced by this PR. SnapshotReuseManifestRef now carries format_version, so pre-PR/version-zero references are rejected before any manifest blob is fetched or parsed. Missing, unavailable, malformed, digest-invalid, incomplete, or unsupported manifests and indexes make the basis ineligible and trigger a full capture; we deliberately do not parse legacy formats or fail reconciliation because optional reuse metadata is unusable. Regression coverage includes version-zero no-fetch behavior and invalid/missing metadata fallback, and the reconciler/proto docs now state this contract.

core/proto/src/main/proto/floecat/reconciler/reconciler.proto
mcusack-floe: Addressed in 5565b02ee. AppendOnlySnapshotBase and the persisted SnapshotPlanBlobStore.AppendOnlyBase now carry chain_depth; it participates in the persisted manifest identity, is propagated through the remote client, and is authenticated by the extracted AppendOnlySnapshotBaseLoader against the predecessor manifest. The planner forces a full capture at floecat.reconciler.snapshot-plan.max-append-only-chain-depth (default 16; zero disables chaining). Added plan round-trip, manifest-depth, and checkpoint regression tests.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReusableArtifactIndexStore.java
mcusack-floe: Addressed in 5565b02ee by extracting the independent stateful components: ReusableArtifactIndexBloomFilter owns filter encoding/parsing/membership, and ReusableArtifactIndexObjectCache owns authenticated object loading plus bounded process caching. ReusableArtifactIndexStore now delegates those concerns and deduplicates final run publication through finishRun. I kept lookup, paging, compaction, and reachability traversal together because they share the same validated run/block ordering contract; splitting those paths further would duplicate or obscure that invariant. Append-only predecessor loading/authentication was also extracted from the finalizer into AppendOnlySnapshotBaseLoader.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/AppendOnlySnapshotBaseLoader.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  Loaded load(
+      ReconcileJobStore.LeasedJob lease,
+      StandaloneSnapshotFinalizeExecutionPayload input,
+      SnapshotPlanBlobStore.AppendOnlyBase base) {
+    if (base == null) {
+      return null;
+    }
+    if (base.snapshotId() == input.snapshotId()
```
mrouvroy-floe, line 39: [major] **Future base accepted**

`AppendOnlySnapshotBaseLoader.load` rejects a base whose `snapshotId` is equal to the target, but it still accepts a base whose `snapshotId` is greater than `input.snapshotId()` as long as the source-file count is not larger.

That breaks the append-only inheritance contract this loader is supposed to authenticate. The finalizer then rewrites the base aggregate records onto the target snapshot and publishes the base index as inherited coverage, so a malformed plan can make snapshot `N` inherit artifacts from snapshot `N+1` instead of failing closed.

Change this guard to require a strictly earlier base, for example `base.snapshotId() >= input.snapshotId()`, and add a finalizer/plan-blob validation test that a future base is rejected before any publication is prepared.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReusableArtifactIndexStore.java
```diff
@@ -0,0 +1,1479 @@
...
```
mrouvroy-floe, line 45: [major] **Oversized reusable index module**

`ReusableArtifactIndexStore` is introduced as a 1,479-line class that owns the persistent index format, append/compaction writer, Bloom/filter/object authentication, lookup, pagination, full traversal, and GC reachability walking.

Those are separate safety boundaries. A mistake in this class can affect both read correctness and object-retention correctness, and the current shape makes it difficult to audit changes to one concern without revalidating the others. The concern is not the line count by itself; it is that authentication and reachability logic now live next to run-writing, cache priming, pagination cursors, block packing, and compaction state.

Split this into focused package-private components with narrow APIs, for example:

- `ReusableArtifactIndexValidator` for reference/object/block shape and digest checks
- `ReusableArtifactIndexWriter` for append, pack writing, and compaction
- `ReusableArtifactIndexReader` for lookup, paging, and block cursors
- `ReusableArtifactIndexReachabilityWalker` for GC traversal

Keep `ReusableArtifactIndexStore` as the small facade that wires those pieces together. That gives the finalizer, repository overlay, and GC code an auditable boundary instead of depending on one large all-purpose implementation.

---

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
```diff
@@ -824,56 +841,78 @@ private Result runPass(String accountId, long deadlineMs) {
...
+            pass.blobsDeleted += generationGc.blobsDeleted();
+            pass.generationCleanupPending |= generationGc.pending();
+            if (!generationProofChanged[0]) {
+              pass.generationGcProof.close();
+              pass.generationGcProof = null;
+              pass.generationGcComplete = true;
+              break;
+            }
+            pass.generationGcProof.close();
+            pass.generationGcProof = reachabilityGuard.beginProof(accountId, tableId);
```
mrouvroy-floe, line 901: [blocker] **Generation GC re-run pairs a fresh epoch proof with a stale re-mark, deleting reused generations**

The generation-reclaim loop begins the epoch proof at line 785 and then builds the `tableReferenced` re-mark once at lines 790-832 (root chain at `rootTableRootChain`, line 829). On an epoch change (`generationProofChanged`), the loop closes and re-begins the proof (line 911) and resets the generation continuation, but it never rebuilds `tableReferenced`. The only in-loop refresh is `rootLivePinChains` inside the protection predicate, which re-roots *pinned* roots only — not a newly finalized current snapshot.

An inherited/reused stats generation is no longer copied into successors; its reachability lives solely in the current root's reuse index. I confirmed in `StatsRepository.deleteUnreferencedGenerations` that such a generation is protected by nothing else on this path: `activeStatsGeneration(snapshotId)` and `isActiveIndexGeneration(...)` key off the generation's *original* owning snapshot (superseded), and the min-age fence (lines 2143-2156) keys off the inherited manifest's `lastModified`, which reuse never rewrites. So protection depends entirely on `isProtectedManifestUri` → `tableReferenced.mightContain(...)`.

Failure scenario:

- Generation `G1` (owned by superseded `S0`) is currently unreferenced. GC builds `tableReferenced` without `G1`.
- An append-only finalize commits snapshot `S2` reusing `G1`, publishing under `reachabilityGuard.publishing(...)` and bumping the epoch. `S2` is not pinned.
- First delete attempt for `G1`: `deleteIfUnchanged` sees the changed epoch → aborts, sets `generationProofChanged`.
- Re-run: proof is re-begun at the new epoch; `tableReferenced` is still the pre-`S2` re-mark (no `G1`). With no further publication, the epoch is now stable, `isProtected(G1)` is false → `G1` is claimed and its blobs deleted while `S2` references them → silent, permanent data loss.

The base branch used `reachabilityGuard.exclusive(rid, ...)` (no re-run); this epoch-proof loop is new. The class doc promises a publication during the pass "forces a complete restart" and is "included by the fresh re-mark" — but the restart never produces a fresh re-mark.

Fix: on `generationProofChanged`, rebuild `tableReferenced` (re-walk the current root chain and pointer scans) after re-beginning the proof, or restart the whole table pass, so the re-mark and proof epoch always correspond to the same state.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
```diff
@@ -364,374 +435,93 @@ private ValidatedCaptureManifest validateManifestObject(
...
         manifest.getRealizedStatsSelectorsList(),
         manifest.getSourceFileCount());
-    if (manifest
-            .getCapturePolicy()
-            .getOutputsList()
-            .contains(ai.floedb.floecat.reconciler.rpc.CaptureOutput.CO_PARQUET_PAGE_INDEX)
-        && manifest.getIndexArtifactCount() != manifest.getSourceFileCount()) {
-      throw new IllegalArgumentException(
-          "snapshot capture manifest index artifacts do not cover planned files");
+    boolean capturedIndexes =
```
mrouvroy-floe, line 468: [minor] **Control-plane no longer cross-checks the manifest capture policy against the trusted lease**

`validateManifestObject` re-reads the finalizer-written capture manifest and re-validates identity, digest, and counts, but derives `capturedIndexes` purely from the manifest's *own* `getCapturePolicy().getOutputsList()` (line 458). The index-coverage check `indexArtifactCount == sourceFileCount` (line 460) and the index-predecessor check are both skipped when that self-declared policy omits `CO_PARQUET_PAGE_INDEX`.

The base branch cross-checked here via `validateCapturePolicy(lease, manifest.getCapturePolicy())` and `validateRealizedIndexSelectors(lease, manifest)`; both were removed. The append-only *base* path still cross-checks (`AppendOnlySnapshotBaseLoader` line 64, `capturePolicyMatches`), but the current-snapshot manifest path no longer does.

The finalizer is documented as trusted and stamps the policy from the lease scope, so this is not a fail-open against an untrusted producer. It is, however, a defense-in-depth regression on a boundary that otherwise independently re-validates the manifest: a finalizer bug that downgrades the capture policy would publish an index-incomplete snapshot marked `reusable_artifact_bundles_complete`, and future append-only reuse would inherit the gap undetected.

Consider restoring an explicit cross-check of the manifest's capture policy / realized index selectors against `lease.scope.capturePolicy()` here.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/SnapshotFinalizePublicationScheduler.java
```diff
@@ -0,0 +1,169 @@
...
+      every = "{floecat.reconciler.snapshot-finalize-publication.tick-every:250ms}",
+      concurrentExecution = Scheduled.ConcurrentExecution.SKIP,
+      skipExecutionIf = ReconcileJobGcScheduler.DisabledOrStopping.class)
+  void tick() {
+    if (inFlight.size() >= maxParallelism) {
+      return;
+    }
+    ReconcileJobStore.SnapshotFinalizeCommitPage page =
+        jobs.pendingSnapshotFinalizeCommits(pageSize, pageToken);
+    pageToken = page.nextPageToken();
```
mrouvroy-floe: [minor] **`pageToken` advances before dispatch, so a saturated page's tail is skipped until wraparound**

`tick()` advances `pageToken = page.nextPageToken()` (line 87) before the dispatch loop, but the loop `break`s as soon as `inFlight.size() >= maxParallelism` (lines 93-94). Any intents after the break point in the current page are never dispatched from that page — they are only revisited after `pageToken` cycles all the way back to `""`.

Under a large accepted-finalizer backlog with the default `maxParallelism=4` and `pageSize=100`, most of each page is skipped every tick, delaying publication of already-accepted finalizers by a full page cycle.

It self-corrects, so this is a fairness/latency issue rather than a correctness one. A fix is to not advance `pageToken` past intents that were not dispatched (e.g. only advance when the whole page was consumed, or track the first undispatched intent as the next token).

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/AppendOnlySnapshotBaseLoader.java
```diff
@@ -0,0 +1,124 @@
...
+        || base.sourceFileCount() > input.sourceFileCount()) {
+      throw new IllegalArgumentException("append-only snapshot base is not an earlier subset");
+    }
+    byte[] bytes = blobStore.get(base.manifestUri());
```
mrouvroy-floe, line 45: [minor] **Missing null-check on `blobStore.get` yields a bare NPE on a missing base blob**

`load()` does `byte[] bytes = blobStore.get(base.manifestUri());` (line 42) and immediately dereferences `bytes.length` (line 43); `loadAggregates()` does the same with `blobStore.get(descriptor.getPayloadUri()).length`.

`BlobStore.get` returns `null` for a missing object (e.g. `InMemoryBlobStore`), whereas the rest of the index store consistently null-checks and throws `StorageNotFoundException` (see `ReusableArtifactIndexStore.loadObject`). Here a missing base manifest or aggregate blob throws a bare `NullPointerException` with no message instead of the descriptive, fail-closed error used everywhere else.

Impact is limited to null-returning stores and error legibility (S3's `get` throws), so this is minor — but it is an inconsistency worth fixing by null-checking and throwing `StorageNotFoundException`.

---

connectors/catalogs/delta/src/main/java/ai/floedb/floecat/connector/delta/uc/impl/DeltaConnector.java
```diff
@@ -844,6 +863,218 @@ public Optional<SnapshotFilePlan> planSnapshotFiles(
...
+          "Delta change planning failed (versions "
+              + baseSnapshotId
+              + ".."
+              + targetSnapshotId
+              + ")",
+          e);
+    }
+    long completedNanos = System.nanoTime();
+
+    LinkedHashMap<String, SnapshotFileEntry> additions = new LinkedHashMap<>();
```
mrouvroy-floe, line 921: [minor] **`planSnapshotFileDelta` fully materializes the delta before checking `appendOnly()` (OOM vector)**

Both `planSnapshotFileDelta` implementations (Delta here at line 916; Iceberg at `IcebergConnector.java:813`) build the complete `additions` map of `SnapshotFileEntry` plus the `removals` set before the caller ever evaluates `appendOnly()`. Since `appendOnly()` is false the moment a single removal or delete-artifact change exists, a non-append commit in the base→target range causes all of that work and memory to be accumulated and then immediately discarded.

A single full-table `OPTIMIZE`/compaction (Delta) or a large `overwrite`/rewrite snapshot (Iceberg) within the range can rewrite millions of files, all of which land in the in-memory collections on the planning worker in one shot (for Delta, one such commit falls entirely within a single reader/chunk). Given this stack is explicitly about GC/OOM hardening (base branch `mc-gc-oom-fix`), this is a real memory-pressure regression vector.

The result is still correctly classified as non-append; this is purely a memory concern. An early exit — abandon accumulation and return a non-append result as soon as the first `RemoveFile` / removed data file / deletion-vector change is observed — eliminates it.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReusableArtifactIndexObjectCache.java
```diff
@@ -0,0 +1,96 @@
...
+  synchronized byte[] get(ReusableArtifactIndexObjectReference reference) {
+    CachedObject cached = objects.get(reference.getUri());
+    if (cached == null) {
+      return null;
+    }
+    if (cached.bytes().length != reference.getPayloadBytes()
+        || !MessageDigest.isEqual(cached.sha256(), reference.getPayloadSha256().toByteArray())) {
+      throw new IllegalArgumentException("reusable artifact index object metadata mismatch");
+    }
+    return cached.bytes();
```
mrouvroy-floe: [nit] **`get` returns the internal cached array without a defensive copy**

`put` stores `value.clone()` (line 50), but `get` returns `cached.bytes()` directly (line 63), so every caller of this process-wide cache shares the same backing array for a given cached object.

Current callers only `parseFrom` (which copies) or `Arrays.copyOfRange` the result, so there is no live mutation today. It is a latent aliasing hazard for a shared static cache — any future caller that mutates the returned array would corrupt the cached (and digest-authenticated) bytes for everyone. Returning a clone, or documenting the array as read-only, would remove the footgun.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/AppendOnlySnapshotBaseLoader.java
mcusack-floe: Not applying the suggested numeric `>=` guard. Snapshot IDs are opaque for Iceberg: a valid descendant can have ID 10 with parent ID 50, so numeric comparison would reject valid lineage. Iceberg proves the relationship by walking `parentId()` from the target to the exact base before emitting a delta; Delta uses its own native ordered versions. The persisted plan carries the connector-selected base identity and the finalizer authenticates that identity. Commit ebfc8bd9d adds a regression with base 50 and target 10 to ensure this remains true. Any additional generic defense-in-depth must use authenticated lineage evidence, not numeric snapshot ordering.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReusableArtifactIndexStore.java
mcusack-floe: Not addressed further in this PR. The Bloom-filter and authenticated object-cache components remain extracted, but the larger validator/writer/reader/reachability decomposition is deliberately out of scope for this correctness pass. This will be picked up in a future PR.

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
mcusack-floe: Fixed in ebfc8bd9d. On `generationProofChanged`, GC now begins the fresh proof and immediately calls `remarkTable` plus `rootLivePinChains` before resetting the generation continuation. That rebuilds the add-only `tableReferenced` protection set from the current root/pointers at the new epoch. The regression test publishes a real root inside `reachabilityGuard.publishing(...)` during the first deletion attempt and verifies the newly referenced generation survives the retry.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
mcusack-floe: Fixed in ebfc8bd9d. Restored `validateCapturePolicy(lease, manifest.getCapturePolicy())` and `validateRealizedIndexSelectors(lease, manifest)` before deriving captured-index state from the manifest. Tests now verify a manifest policy that differs from the trusted lease is rejected.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/SnapshotFinalizePublicationScheduler.java
mcusack-floe: Fixed in ebfc8bd9d. `pageToken` advances only after the complete page has been consumed; hitting `maxParallelism` leaves the token on the current page so its tail is retried on the next tick. Null intents now skip rather than terminate the page. Added saturated-page regression coverage.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/AppendOnlySnapshotBaseLoader.java
mcusack-floe: Fixed in ebfc8bd9d. Both the base manifest and aggregate payload reads now explicitly reject null `BlobStore.get(...)` results with `StorageNotFoundException` containing the missing URI. Added coverage for both missing-object paths.

connectors/catalogs/delta/src/main/java/ai/floedb/floecat/connector/delta/uc/impl/DeltaConnector.java
mcusack-floe: Fixed in ebfc8bd9d with bounded preflight paths. Delta now performs a non-retaining classification pass before a second pass materializes additions, so add actions that precede a remove do not accumulate in memory. Iceberg checks only current-snapshot manifest summaries for data removals or delete-artifact changes before invoking the file-change accessors; non-append lineage falls back without materializing either cached file-change list. Regression coverage also verifies opaque Iceberg snapshot IDs.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReusableArtifactIndexObjectCache.java
mcusack-floe: Fixed in ebfc8bd9d. `ReusableArtifactIndexObjectCache.get()` now returns `cached.bytes().clone()`, matching clone-on-insert and preventing callers from mutating the process-wide cached array. Added a mutation-isolation regression test.

mrouvroy-floe (commented):
**[major] Safety-critical validation logic spread across a 1,200-line executor class** — `reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotFinalizeReconcileExecutor.java:55`

`RemoteSnapshotFinalizeReconcileExecutor` (~1,225 lines) interleaves executor control flow with append-only base authentication, aggregate inheritance, coverage validation, reusable-bundle validation, and manifest-preparation inputs.

These checks are not incidental — they are the boundary that prevents a partial file-group output or stale inherited coverage from being published as a complete snapshot. Co-locating the happy path and the fail-closed checks in the same large method bodies makes it difficult to audit each invariant independently; a future edit to the executor loop can silently weaken a validation rule. The coverage-by-count gap (see separate finding) is a concrete example of safety logic being spread thin across the executor, its client, and a coverage service.

Extract artifact-validation, manifest-preparation, and coverage-gate logic into dedicated, independently-testable components. The executor should orchestrate; it should not own the safety boundary.

**[minor] Empty/unparseable scope set vends and caches an unscoped full-role credential** — `service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityResolver.java:513`

`scopedSessionPolicy` returns `null` when the normalized scope set is empty, and `assumeRoleOnce` then calls `AssumeRoleRequest.builder().policy(null)`, which mints a credential with the assumed role's full permissions (no session-policy scoping).

The null-policy behavior predates this PR, but the new credential cache amplifies the exposure: the unscoped credential is stored under key `(authority, secret, [])` and reused for every subsequent empty/unparseable-scope request. The path is hard to reach today because upstream `validateAuthorityCoversSessionScope` requires each scope to start with the authority's `s3://` prefix, but a scope list that normalizes to empty currently fails open toward broad permissions rather than closed.

Reject an empty normalized scope set with an explicit error (fail closed) instead of vending and caching a broad credential.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
```diff
@@ -257,6 +204,127 @@ public boolean persistSuccess(
...
 
+  public boolean publishAcceptedSnapshotFinalize(String jobId) {
+    ReconcileJobStore.SnapshotFinalizeCommitIntent intent =
+        jobs.snapshotFinalizeCommitIntent(jobId).orElse(null);
+    if (intent == null) {
+      return false;
+    }
+    long startedNanos = System.nanoTime();
+    ReconcileJobStore.LeasedJob lease =
+        requireLeasedSnapshotFinalizeJob(jobId, jobId, intent.leaseEpoch());
```
mrouvroy-floe: [major] **Deferred publication requires live worker lease, defeating crash-recovery design**

`publishAcceptedSnapshotFinalize` calls `requireLeasedSnapshotFinalizeJob(jobId, jobId, intent.leaseEpoch())`, which invokes `jobs.renewLease(jobId, leaseEpoch)`. Both job stores require the lease to be unexpired at renewal time.

This means the deferred scheduler can only publish while the *original* worker's lease is still live. If the service restarts after `beginSnapshotFinalizeCommit` records the intent but before `SnapshotFinalizePublicationScheduler` runs — or if publication is delayed past lease expiry — the scheduler will hit `FAILED_PRECONDITION` on every attempt, eventually requeue or fail a result that was already durably accepted. The snapshot then never records the reuse manifest or advances current state.

Issue a fresh lease (or a dedicated publication-only lease token) for the deferred publication path rather than requiring the original lease to remain valid.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/GrpcRemoteReconcileExecutorClient.java
```diff
@@ -1402,10 +1552,12 @@ public PreparedSnapshotFinalizeSuccess prepareSnapshotFinalizeSuccess(
                 .distinct()
                 .sorted()
                 .toList();
+    int inheritedFileCount = appendOnlyBase == null ? 0 : appendOnlyBase.sourceFileCount();
     if (stableFileGroups.size() != snapshotTask.fileGroupCount()
```
mrouvroy-floe, line 1556: [minor] **Cross-group coverage gate is count-based, not identity-based, allowing double-coverage to mask a missing file**

The only cross-group/base coverage gate is `sum(succeededFileCount) + inheritedFileCount == sourceFileCount`. Per-group file identity is enforced against the immutable plan, and cross-group/base overlap is caught by the reusable-artifact-index duplicate detector — but that detector only runs when per-file stats or index artifacts exist.

For a degenerate capture policy where file groups have succeeded files but request neither per-file stats nor indexes, there are no reusable-index entries, so nothing enforces file identity across the union. A plan that double-covers file `X` (in two groups) while omitting a real source file `Y` would satisfy `sum == sourceFileCount` and publish as complete.

Triggering this requires both a degenerate policy and a planner partition defect, so the latent risk is low. Nonetheless, consider adding an identity-level cross-group coverage check that is not gated on the presence of per-file index artifacts.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReusableArtifactIndexStore.java
```diff
@@ -0,0 +1,1479 @@
...
+        ? "inline:" + HexFormat.of().formatHex(reference.getPayloadSha256().toByteArray())
+        : "uri:" + reference.getUri();
+  }
+
+  private static BlockIdentity blockIdentity(ReusableArtifactIndexBlockReference reference) {
+    return new BlockIdentity(
+        objectIdentity(reference.getObject()), reference.getOffset(), reference.getLength());
+  }
+
+  private static void validateOwnedRuns(String prefix, ReusableArtifactIndexReference reference) {
```
mrouvroy-floe: [minor] **`validateOwnedRuns` checks manifest/filter URIs but not block-pack URIs inside manifests**

On the `append` path, `validateOwnedRuns` only verifies that each run's `manifest` and `filter` URIs start with the owning table prefix. It never loads the manifests to check the block/pack object URIs inside them. The read/GC path (`validateReferenceObjects`) does call `validateOwnedObject` on every pack object, so the two paths are asymmetric.

An inherited base whose manifest/filter URIs are correctly prefixed but whose block packs point at another table's prefix would be accepted by `append` and carried into the new reference. Packs are still content-authenticated by `blockSha256`, so this is a cross-tenant ownership/isolation gap rather than a content-integrity break, and it requires a forged or misrouted base (which is authenticated upstream).

Apply the same per-object ownership check on the `append` path that `validateReferenceObjects` already performs on the read/GC path.

---

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
```diff
@@ -824,56 +841,82 @@ private Result runPass(String accountId, long deadlineMs) {
...
+                          tableReferenced, pass.tableWalkedPinRoots, pass.tableWalkFailures);
+                      return pass.tableWalkFailures[0] > 0
+                          || tableReferenced.mightContain(normalized);
+                    },
+                    nowMs,
+                    minAgeMs,
+                    remainingGenerationBlobDeletes,
+                    deadlineMs,
+                    pass.generationGcContinuation,
+                    (isProtected, claim, restore) -> {
```
mrouvroy-floe, line 880: [minor] **Generation-protection recheck performs remote I/O while holding the table write lock**

On a `tableReferenced` miss, the `stillProtected` recheck runs inside the `deleteIfUnchanged` write-lock section and invokes `rootLivePinChains` → `rootTableRootChain`, which does remote blob/pointer I/O. This contradicts the PR's own design note ("remote blob I/O remains outside it") and the guard javadoc's claim that the check proceeds "without serializing concurrent publishers": every generation claim that reaches this step holds the table write lock across pin-chain I/O, blocking that table's publishers and query-pin registration for the duration.

The behavior is fail-safe for correctness (a walk failure is treated as "protected") and is largely carried over from the pre-PR `exclusive(...)` path, so this is a latency and design-consistency concern rather than a data-loss risk.

Consider building the pin-chain proof before taking the write lock and passing the result in, keeping the lock section free of remote I/O.

---

service/src/main/resources/application.properties
```diff
@@ -120,7 +120,7 @@ floecat.reconciler.job-store=${FLOECAT_RECONCILER_JOB_STORE:durable}
 floecat.reconciler.job-store.max-attempts=${FLOECAT_RECONCILER_JOB_STORE_MAX_ATTEMPTS:8}
 floecat.reconciler.job-store.base-backoff-ms=${FLOECAT_RECONCILER_JOB_STORE_BASE_BACKOFF_MS:500}
 floecat.reconciler.job-store.max-backoff-ms=${FLOECAT_RECONCILER_JOB_STORE_MAX_BACKOFF_MS:30000}
-floecat.reconciler.job-store.lease-ms=${FLOECAT_RECONCILER_JOB_STORE_LEASE_MS:120000}
+floecat.reconciler.job-store.lease-ms=${FLOECAT_RECONCILER_JOB_STORE_LEASE_MS:600000}
```
mrouvroy-floe: [minor] **Job-store lease default raised 5x (120s → 600s) with no call-out in deployment notes**

`floecat.reconciler.job-store.lease-ms` changes from `120000` to `600000`. This extends how long a dead worker's job is held before another worker can reclaim it, affecting failover latency for every reconcile job type, not just append-only captures.

The change is plausibly intentional (append-only finalize publication and larger captures can run longer), but the PR's "Config/env var changes" checkbox is left unchecked, so this operational change is easy to overlook during deploy planning.

Explicitly call out this lease change — and the new `snapshot-finalize-publication.*` settings — in the PR's deployment notes, and confirm the 5x increase is deliberate rather than incidental.

---

connectors/catalogs/iceberg/src/main/java/ai/floedb/floecat/connector/iceberg/impl/IcebergConnector.java
```diff
@@ -770,6 +783,106 @@ public Optional<SnapshotFilePlan> planSnapshotFiles(
...
+        IcebergPlanner.partitionJson(spec, dataFile.partition()),
+        spec == null ? 0 : spec.specId(),
+        List.of(),
+        sequenceNumber,
+        null,
+        List.of(),
+        IcebergPlanner.dataContentIdentity(sequenceNumber, dataFile.recordCount()));
+  }
+
+  private static boolean hasAny(Iterable<?> values) {
```
mrouvroy-floe: [nit] **Dead `hasAny` method and imprecise non-append delta fields**

`IcebergConnector.hasAny` is unused dead code and should be removed.

Separately, when a non-append change is detected, `SnapshotFileDelta.removedDataFilePaths` and `deleteArtifactsChanged` are only approximate: Iceberg always returns an empty removal list, and under the Delta connector's cancellation race an early-returning partition can report `deleteArtifactsChanged=true` with an empty removal list even when the real cause was a data-file removal. This is harmless today because the sole consumer reads only `appendOnly()`, but a future consumer trusting these fields as complete would be misled.

Remove the dead method, and either populate the removal fields accurately or add a doc comment stating that only `appendOnly()` is authoritative when it returns `false`.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
mcusack-floe: Fixed in 4577dd814. Deferred publication now loads a completion view for the durably accepted result instead of renewing the originating worker lease. Both stores distinguish the initial finalize-ownership flag from a complete publishable intent; a complete running intent survives lease expiry, while pre-intent/partial-intent records are reclaimed normally. Matching-epoch success and failure transitions can complete after expiry, and retry/reclaim clears the full flattened intent payload so a later attempt cannot resurrect stale data. Coverage includes publication after real lease expiry, pre-intent and partial-intent reclaim, stale-payload clearing, and cancellation recovery.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/GrpcRemoteReconcileExecutorClient.java
mcusack-floe: Fixed in 4577dd814. The finalizer now validates identity over the immutable planned file groups before processing results: every planned path must be unique across groups, and unique planned paths plus the authenticated append-only base file count must equal the declared source-file count. Combined with the existing per-group result-to-plan identity checks, this proves full coverage without depending on stats or index artifacts being present. Added a regression where duplicate cross-group paths preserve the count but omit a source file.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReusableArtifactIndexStore.java
mcusack-floe: Fixed in 4577dd814. append now applies validateReferenceObjects to the inherited base, loading inline/external run manifests and validating ownership of every filter, manifest, and block-pack object before carrying the reference forward. Added a cross-owner pack regression. This intentionally adds metadata reads proportional to inherited runs and their run manifests on the finalize path; it does not load every pack payload, and the work is bounded by the persisted run/block structure. That cost is now called out in the PR summary.

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
mcusack-floe: Fixed in 4577dd814. The generation-protection scan now runs before deleteIfUnchanged takes the table write lock. The epoch proof is retained across that scan; the lock section validates the epoch and performs the claim only. Any concurrent publication/query-pin registration bumps the epoch, rejects the claim, and causes the outer loop to re-mark before retrying. Added focused coverage that asserts the protection supplier executes outside the publication write lock; the existing publication-race regression still verifies the newly referenced generation survives.

service/src/main/resources/application.properties
mcusack-floe: Addressed in 4577dd814 and the PR body. The general lease default is back to 120000. That revert is a consequence of accepted snapshot-finalize publication now being independent of the originating worker lease lifetime, so the temporary 600000 mitigation is no longer needed. The three new snapshot-finalize-publication controls (tick-every, page-size, max-parallelism), defaults, and environment variables are documented, and the Config/env var changes checkbox is now checked.

connectors/catalogs/iceberg/src/main/java/ai/floedb/floecat/connector/iceberg/impl/IcebergConnector.java
mcusack-floe: Fixed in 4577dd814. Removed the dead Iceberg hasAny helper. Both connectors now classify non-append changes before materializing the append delta: Iceberg uses current-snapshot manifest summaries, and Delta uses a non-retaining first pass followed by materialization only for an append-only range. Delta cancellation is represented explicitly rather than fabricating deleteArtifactsChanged, and any cancelled materialization chunk fails closed with Optional.empty. No partially populated non-append SnapshotFileDelta is exposed to a future consumer.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/LeasedSnapshotFinalizeExecutionService.java
mcusack-floe: Cancellation edge clarification: a publishable finalizer already in JS_CANCELLING can still be admitted through the normal live-lease completion-view path. The expired-lease bypass itself is now JS_RUNNING-only. If the live path wins, durable completion CAS converts the job to JS_CANCELLED; the in-memory completion path removes the pending intent even when cancellation wins. If neither completes first, expiry reclaim clears the intent and finishes cancellation. Public cancellation also refuses a running finalizer once commit ownership is recorded, so this is a bounded race rather than an unbounded republish loop.

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/SnapshotFinalizePublicationScheduler.java
```diff
@@ -0,0 +1,176 @@
...
+          1L,
+          0L,
+          0L);
+      LOG.warnf(e, "Rejected accepted snapshot finalizer result jobId=%s", intent.jobId());
+    } catch (RuntimeException e) {
+      RetryState prior = retries.get(intent.jobId());
+      int attempts = prior == null ? 1 : prior.attempts() + 1;
+      if (attempts >= 8) {
+        retries.remove(intent.jobId());
+        jobs.markFailed(
```
mrouvroy-floe: [major] **Accepted finalize intent lost after transient publication failures**

After eight transient `RuntimeException`s, the scheduler calls `jobs.markFailed(...)`, which clears the accepted publication intent via `clearSnapshotFinalizeIntent()` and removes the manifest from `pendingSnapshotFinalizeCommits()`.

This breaks the durable publication contract: a manifest already accepted by `beginSnapshotFinalizeCommit(...)` can be silently dropped from recovery simply because publication was temporarily failing. The accepted-intent state should survive transient failures; only proven-invalid payloads should trigger a terminal clear.

Keep transient publication failures in the accepted-intent state and retry publication with durable or local backoff. Only promote to terminal failure when the payload is proven invalid, not when the remote endpoint is transiently unavailable.

---

connectors/catalogs/iceberg/src/main/java/ai/floedb/floecat/connector/iceberg/impl/IcebergConnector.java
```diff
@@ -770,6 +783,102 @@ public Optional<SnapshotFilePlan> planSnapshotFiles(
...
+
+    return Optional.of(
+        new SnapshotFileDelta(
+            List.copyOf(additions.values()),
+            List.of(),
+            false,
+            SchemaParser.toJson(schemaForSnapshot(table, target))));
+  }
+
+  private static boolean hasDataFileRemovals(Snapshot snapshot, FileIO io) {
```
mrouvroy-floe, line 852: [major] **Append-only gate may miss file removals, causing deleted files to be resurrected**

`hasDataFileRemovals`/`hasDeleteArtifactChanges` classify a snapshot as append-only by scanning only manifests whose `snapshotId()` equals the snapshot id and checking `deletedFilesCount()`/`addedFilesCount()`. This is the sole correctness gate for the entire reuse fast path.

If Iceberg drops an all-deleted manifest from the manifest list rather than rewriting it with DELETED entries, the scan sees no removal and misclassifies the snapshot as append-only. The planner then reuses the base and the finalizer inherits its file coverage, resurrecting removed files into the new snapshot — a silent data-correctness violation.

Derive append-only classification from `snapshot.operation()` (or an equivalent authoritative source) rather than from manifest delete-counts alone, which can be absent for certain removal patterns.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/impl/SnapshotFinalizePublicationScheduler.java
```diff
@@ -0,0 +1,176 @@
...
+      pageToken = page.nextPageToken();
+      if (pageToken.isBlank()) {
+        pageToken = "";
+      }
+    }
+  }
+
+  private void publish(ReconcileJobStore.SnapshotFinalizeCommitIntent intent) {
+    try {
+      if (publicationService.publishAcceptedSnapshotFinalize(intent.jobId())) {
```
mrouvroy-floe: [minor] **`retries` entry leaks when publication intent is cleared by another instance**

In `publish()`, when `publishAcceptedSnapshotFinalize` returns `false` (intent was cleared or requeued by another instance), `retries.remove(intent.jobId())` executes only inside the `true` branch. The `finally` block removes only from `inFlight`. Any pre-existing `retries` entry for a jobId that resolves via the `false` path is therefore never removed.

This is a memory/robustness issue rather than a correctness bug — overall progress is still bounded by the durable job attempt count — but the map grows unboundedly under concurrent multi-instance scenarios.

Remove the jobId from `retries` on both branches, e.g. by moving `retries.remove(intent.jobId())` into the `finally` block unconditionally.

---

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/DurableReconcileJobStore.java
```diff
@@ -143,7 +144,7 @@ public ConnectorDeletedException(String connectorId) {
   private static final int DEFAULT_MAX_ATTEMPTS = 8;
   private static final long DEFAULT_BASE_BACKOFF_MS = 500L;
   private static final long DEFAULT_MAX_BACKOFF_MS = 30_000L;
-  private static final long DEFAULT_LEASE_MS = 120_000L;
+  private static final long DEFAULT_LEASE_MS = 600_000L;
```
mrouvroy-floe: [minor] **`DEFAULT_LEASE_MS` not reverted to `120_000L` as stated in PR description**

The base branch had `DEFAULT_LEASE_MS = 120_000L`; this PR sets it to `600_000L`, contradicting the PR description ("reverting the temporary `600000` value") and the review reply claiming the default is back to 120000.

The packaged service is unaffected because `application.properties` supplies a `:120000` fallback via config, but the in-code constant is used by any test, embedding, or deployment that omits the property — quintupling failover latency for reclaiming a dead worker's job in those environments.

Revert `DEFAULT_LEASE_MS` to `120_000L` to match the stated intent. If the 600s fallback is deliberate, correct the PR description and document the rationale.

---

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityResolver.java
```diff
@@ -249,6 +473,33 @@ StsClient buildStsClient(StorageAuthority authority, AwsCredentialsProvider prov
...
+
+  private record AssumeRoleCacheKey(
+      String authorityFingerprint, String sourceCredentialFingerprint, List<String> scopes) {
+    private static AssumeRoleCacheKey of(
+        StorageAuthority authority,
+        AuthCredentials authoritySecret,
+        List<String> sessionScopeLocations) {
+      return new AssumeRoleCacheKey(
+          sha256(authority.toByteArray()),
+          sha256(authoritySecret == null ? new byte[0] : authoritySecret.toByteArray()),
```
mrouvroy-floe: [minor] **STS cache key uses non-deterministic protobuf serialization, causing spurious cache misses**

`AssumeRoleCacheKey.of` fingerprints the source credential with `sha256(authoritySecret.toByteArray())`. `AuthCredentials` contains `map<string,string>` fields (`properties`, `headers`), and protobuf `toByteArray()` does not guarantee deterministic ordering of map entries.

When those maps are non-empty, two logically-identical secrets can serialize to different bytes, producing different SHA-256 hashes and defeating the cached/single-flight STS vending this PR adds. The failure mode is safe (extra STS calls, never merged credentials), but it silently undermines the caching goal.

Build the fingerprint from a canonical form — e.g. sort map entries by key before hashing, or use a deterministic serialization — rather than relying on `toByteArray()`.

---

mrouvroy-floe (commented):
**[minor] Empty normalized scope vends and caches an unscoped full-role credential** — `service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityResolver.java:513`

`scopedSessionPolicy` returns `null` when the normalized scope set is empty, and `assumeRoleOnce` then passes `policy(null)` to `AssumeRoleRequest.builder()`, minting a credential with the assumed role's full permissions — no session-policy downscoping.

The new credential cache amplifies this: the unscoped credential is stored under key `(authority, secret, [])` and reused for every subsequent empty-scope request. The PR's stricter `s3://`-only parsing in `normalizeS3Scopes` also slightly widens the set of inputs that normalize to empty (non-S3 URIs are now dropped silently).

The path is currently guarded by `validateAuthorityCoversSessionScope`, but a scope list that normalizes to empty after validation could still reach this path. Either reject empty normalized scopes explicitly, or ensure `scopedSessionPolicy` returns a maximally-restrictive policy (e.g. deny-all) rather than `null` when the scope set is empty.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotFinalizeReconcileExecutor.java
```diff
@@ -760,7 +861,7 @@ static DeduplicatedSnapshotArtifacts deduplicateSnapshotArtifacts(
       for (var metadata : bundle.getFileStatsList()) {
         String target =
             StatsTargetIdentity.storageId(StatsTargetIdentity.fileTarget(metadata.getFilePath()));
-        if (ownerByTarget.getOrDefault(target, -1) == index) {
+        if (!ownerByTarget.containsKey(target) || ownerByTarget.get(target) == index) {
```
mrouvroy-floe: [minor] **Dedup emits reuse-bundle stats metadata for un-owned targets**

The dedup guard changed from "emit only when this bundle owns the target" to `!ownerByTarget.containsKey(target) || ownerByTarget.get(target) == index`. Because `ownerByTarget` is populated only from committed `descriptors` (file-stats), a reuse-bundle `fileStats` entry whose file target has no corresponding committed `StatsObjectDescriptor` is no longer dropped — it is emitted in *every* bundle that carries it.

If such a target can occur, the normalized bundles then carry that file path in multiple bundles. Downstream in `LeasedSnapshotFinalizeExecutionService.publishCaptureArtifacts`, `reusableStatsFiles.add(...)` rejects the duplicate ("duplicate or invalid stats metadata") or the artifact-count check fails — a valid finalize is rejected rather than published. This is a false-reject (liveness), not data acceptance: the new condition only ever emits a superset of the old one, so it cannot under-count needed metadata.

Worth confirming whether `artifacts.fileStats()` can ever be a proper subset of a bundle's file targets; if so, restore the owner-present requirement (`ownerByTarget.getOrDefault(target, -1) == index`) or add a targeted test pinning the invariant.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ReusableArtifactIndexStore.java
```diff
@@ -0,0 +1,1468 @@
...
+            > 0) {
+      throw new IllegalArgumentException("reusable artifact block reference is invalid");
+    }
+    validateObjectReference(reference.getObject());
+    if (reference.getObject().getPayloadBytes() > TARGET_PACK_BYTES) {
+      throw new IllegalArgumentException("reusable artifact index pack is too large");
+    }
+    long length = Integer.toUnsignedLong(reference.getLength());
+    if (reference.getLength() < 0
+        || reference.getOffset() > reference.getObject().getPayloadBytes()
```
mrouvroy-floe, line 1218: [minor] **Block-range validation omits negative-offset lower bound**

`validateBlockReference` bounds the block range with `offset > payloadBytes` and `length > payloadBytes - offset`, but never rejects a negative `getOffset()`. With a negative offset the second check's right-hand side grows, so a crafted reference passes validation and reaches `blobStore.getRange(uri, negativeOffset, length)`.

In practice this fails closed — the subsequent block SHA-256/length re-check in `decodeBlock` (or `copyOfRange`/`toIntExact`) rejects the bytes, so no unauthenticated data is returned. But if any `BlobStore.getRange` implementation interprets a negative offset as an HTTP-style suffix range, the guard is the only thing standing between it and a surprising read.

Add `reference.getOffset() < 0` to the invalid-range condition so the bound is explicit rather than relying on the downstream digest check.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/RemoteSnapshotPlanningReconcileExecutor.java
```diff
@@ -461,13 +559,222 @@ private List<ReconcileFileGroupTask> enrichFileGroupTasks(
...
+          historical.snapshot().getSnapshotId());
+      return Optional.empty();
+    }
+    if (delta == null || !delta.appendOnly()) {
+      return Optional.empty();
+    }
+    String baseSchema = historical.snapshot().getSchemaJson();
+    if (baseSchema.isBlank()
+        || delta.executionSchemaJson().isBlank()
+        || !baseSchema.equals(delta.executionSchemaJson())) {
```
mrouvroy-floe: [nit] **Append-only schema gate couples two distinct proto fields by exact bytes**

The reuse gate requires `historical.snapshot().getSchemaJson()` (catalog `Snapshot.schema_json`, set from the enumeration descriptor) to be byte-identical to `delta.executionSchemaJson()` (connector-derived). Both currently flow from the same connector serialization (`SchemaParser.toJson` for Iceberg, `DataTypeJsonSerDe.serializeStructType` for Delta), so equal schemas produce equal strings and the fast path engages — the check is safe today.

The risk is silent: if either serialization path ever normalizes differently (field ordering, whitespace, metadata) for an otherwise-unchanged schema, this comparison silently disables append-only reuse table-wide with no error — the feature becomes a no-op and nothing signals it.

Consider comparing parsed/canonicalized schemas instead of raw JSON, or add a test that pins `schema_json == execution_schema_json` for an unchanged schema on both connectors so a future serialization change fails loudly rather than quietly turning off the optimization.

---

mrouvroy-floe (approved):

PR #418, opened 2026-07-27
GetUserObjects - 5: preserve scanner SPI compatibility and document view bases

mrouvroy-floe:
## Summary

Keep explicit-context name resolution backward-compatible for existing CatalogOverlay implementations. Document that eager view base relations are emitted with input_index = -1.

## Type of Change

- [ ] feat (new functionality)
- [ ] fix (bug fix)
- [x] docs (documentation only)
- [ ] refactor (no behavior change)
- [ ] test (adds/updates tests)
- [x] chore (build/tooling/maintenance)

## Testing

Describe how this was validated.

- [x] `make fmt`
- [x] `make verify`
- [ ] Added/updated tests for changed behavior

## Deployment and Compatibility

- [x] No breaking changes
- [ ] Breaking changes (describe below)
- [ ] Config/env var changes (describe below)

## Checklist

- [x] PR is scoped and focused
- [x] Commit messages follow conventional commit style where practical
- [x] Documentation updated where needed
- [x] No credentials, tokens, or secrets are included
- [x] This PR does not disclose a security vulnerability publicly (see `SECURITY.md`)

PR #432, opened 2026-08-03
concurrency: bounded ordered fan-out with request-context propagation

mrouvroy-floe:
Foundation for parallelizing metadata work: the parallelism substrate plus the request-context carrier it needs. No domain wiring yet — the first production caller arrives in PR 4.

### What's here

- **`BoundedFanout`** — runs independent, mostly-blocking tasks under a concurrency bound: at most *N* in flight, a completed task immediately reopens a slot, results delivered in **input order**. The first failure reachable in input order propagates immediately; active siblings are cancelled or abandoned rather than drained. Fan-out methods are **package-private on purpose** — the narrow surface the later admission tier drives, so a per-unit store call cannot reach the store un-admitted.
- **`PropagatedContext`** — captures a request's ambient context and re-establishes it on worker threads. Four carriers travel together: the OpenTelemetry context (with the call's gRPC server span grafted on), the forked `io.grpc.Context`, the `ResolvedCallContext`, and the log MDC.
- **`ResolvedCallContexts`** — gains `withCurrentCallSpan` (the span-graft rule, previously duplicated verbatim in `BaseServiceImpl`, which now delegates) and `callWithoutRequestScope` (an explicitly-empty scope, distinct from "no scope").
- **`Futures.propagate`** — surfaces a task's original `RuntimeException`/`Error` instead of an execution wrapper.

### Contracts worth reviewing

These are the properties the scheduler actually promises, each stated in the class javadoc because each has a sharp edge:

- **First-failure precedence holds from the very first submit.** An initial-fill or refill submission failure (cancellation, executor rejection) is reconciled against every already-terminal task, so a real store failure always wins over a masking rejection. Reconciliation consults futures directly rather than the completion queue, because a future's terminal transition happens-before its queue notification.
- **`permits` bounds concurrency, not buffered results.** The window refills past a still-running input-order head, so undelivered results are bounded by `items.size()`, not by `permits`. Deliberate — the alternative stalls a fast later task behind a slow earlier one — but it means one slow head can hold every downstream result in heap.
- **Return does not imply completion.** Cancellation/failure *abandon* siblings (`cancel(true)` interrupts but does not wait), so a task may still be inside a non-interruptible store call when this returns. This class does not hold the store ceiling; an admitted call holds its permit until the call itself returns (PR 2).
- **Executor requirement.** The executor must run tasks independently of the caller, and must run-or-reject every task it accepts — never silently discard. There is deliberately no wait deadline: any ceiling low enough to catch a wedged executor would also abort legitimately slow metadata reads.

### Cross-request isolation

`PropagatedContext` forks and carries the caller's gRPC context rather than leaving the worker's own. That matters because several readers consult **only** gRPC keys — the session/authorization tokens `AuthResolutionContexts` reads, and `PrincipalProvider`/`EngineContextProvider`'s fallbacks. Carrying the captured context both delivers this request's credentials to fan-out tasks *and* isolates an off-request capture (it carries an empty context instead of inheriting whatever foreign request the worker happened to bear). This mirrors flight's `ContextSnapshot`, deliberately, so the two dispatch points agree on gRPC semantics.

MDC is handled separately: on a Quarkus duplicated Vert.x context the MDC is shared by every concurrent task, so `supply` skips MDC propagation there and warns once per JVM rather than corrupting it — a logging degradation instead of failing every task on a misconfigured executor.

**Tests:** `BoundedFanoutTest` (37), `PropagatedContextTest` (12), `FuturesTest` (3), plus a relocated invariant in `InboundContextInterceptorContextTest` pinning that `populateMdc` writes only the request ids and never the dispatch dimensions.

### Second commit
`test: cover transaction root resync` — adds the failure-branch test for `resyncRootOrLeaveMarker` (a failed resync must leave a durable re-drive marker, the only convergence path for REST-only tables). Unrelated to fan-out; folded here rather than blocking on its own PR.

---
Part of the GetUserObjects parallel re-cut. Supersedes #415–#418. Base: `main`.

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
+++ b/service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
@@ -1,0 +1,882 @@
+/*
+ * Copyright 2026 Yellowbrick Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ai.floedb.floecat.service.concurrent;
+
+import ai.floedb.floecat.service.context.PropagatedContext;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.Future;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.function.BooleanSupplier;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.IntConsumer;
+import org.jboss.logging.Logger;
+
+/**
+ * Runs independent, mostly-blocking tasks with a concurrency bound.
+ *
+ * <p>At most the configured number of tasks are submitted at a time. A completed task immediately
+ * opens one submission slot, independent of input order. Completed results are retained by index;
+ * mapping and ordered consumption deliver each contiguous input-order prefix as soon as it is
+ * available. Once that prefix exposes a failure, active siblings are cancelled or abandoned and the
+ * original failure returns without waiting for unrelated work.
+ *
+ * <p>This is orchestration only — it applies no I/O admission of its own. Its fan-out methods are
+ * package-private on purpose: an I/O-admission tier is meant to wrap them, so widening them to
+ * {@code public} would let a caller fan out per-unit store calls that bypass admission entirely.
+ * Keep them package-private.
+ *
+ * <p>{@code permits} bounds concurrency, not buffered results. The window refills past a
+ * still-running input-order head, so a completed later result is retained until the head lets the
+ * contiguous prefix advance. Undelivered results are therefore bounded by {@code items.size()}, not
+ * by {@code permits}: one slow head can hold every downstream result in heap at once. That is the
+ * deliberate trade for not stalling a fast later task behind a slow earlier one — size the batch
+ * (or the per-element footprint) accordingly rather than assuming a {@code permits}-sized buffer.
+ *
+ * <p>Return does not imply completion. Cancellation and failure <em>abandon</em> active siblings —
+ * {@code Future.cancel(true)} interrupts them but does not wait — so this class can return with
+ * tasks still running inside a non-interruptible store call (an S3/JDBC round trip that ignores the
+ * interrupt). This class holds no admission permit of its own.
+ *
+ * <p><b>Executor requirement:</b> {@code executor} must be able to run submitted tasks
+ * independently of the calling thread. The scheduler blocks the caller awaiting completions, so a
+ * bounded executor that the caller thread <em>itself</em> occupies deadlocks — the caller holds the
+ * only worker while its own submissions queue behind it, with no timeout. An unbounded or
+ * virtual-thread executor is always safe, as is any pool separate from the caller's. Use one of
+ * those. This cannot be detected here (the class does not know the caller's pool), so it is the
+ * caller's contract.
+ *
+ * <p>A same-pool {@link ForkJoinPool} is <b>no exemption at all</b>, despite {@link
+ * ForkJoinPool#managedBlock}. Compensation covers the <em>scheduler's</em> wait — one credit per
+ * managed block, and {@code tryCompensate} often satisfies it by lowering the active count rather
+ * than adding a worker. Nothing compensates for a submitted task that blocks, because that park is
+ * unmanaged, and mostly-blocking tasks are this class's whole premise. So the pool stops growing
+ * after a small constant and every later submission starves. Measured on a single-worker pool, with
+ * the pool capping at three threads: a cancellable caller stalls permanently with two blocking
+ * tasks, and {@link #NEVER_CANCELLED} stalls from three onward. Neither path is safe — the sentinel
+ * merely fails one task later. Do not dispatch to the pool the caller is running on.
+ *
+ * <p>{@code managedBlock} is still used, and still earns its keep for the supported topology: a
+ * scheduler running on an FJP worker that dispatches tasks <em>elsewhere</em> lends the pool a
+ * worker for the duration of its wait. Off an FJP it degrades to an ordinary block.
+ *
+ * <p>{@code executor} must also <b>run or reject</b> every task it accepts, never silently drop
+ * one. The scheduler's only exits from its wait are a completion notification, an observed
+ * cancellation, or an interrupt: a task accepted and then discarded never notifies, so {@code
+ * active} never drains and the calling thread parks. A rejection is fine (it surfaces through the
+ * ordered failure path); a {@link java.util.concurrent.ThreadPoolExecutor.DiscardPolicy} or {@code
+ * DiscardOldestPolicy} is not, and neither is calling {@code shutdownNow()} on the executor while a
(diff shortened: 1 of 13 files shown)


service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
```diff
@@ -0,0 +1,675 @@
...
+   * contiguous prefix is delivered — a reachable failure throws here — before the submission
+   * failure itself is surfaced.
+   */
+  private static <O> void surfaceSubmissionFailure(
+      Throwable submissionFailure,
+      TaskRuntime<?, O> runtime,
+      List<CompletionSlot<O>> active,
+      List<TaskOutcome<O>> outcomes,
+      OrderedOutcomeConsumer<O> ordered) {
+    for (CompletionSlot<O> ready : runtime.drainCompleted()) {
```
jchauzi-floe: [major] **Terminal-future race masks ordered failures**

`surfaceSubmissionFailure` treats only slots already present in `completions` as completed. That publication is later than the future's terminal transition: `FutureTask` publishes its terminal state before invoking `done()`, and `CompletableFuture` stores its result before running the `whenComplete` action. A sibling can therefore already be terminal while its slot is still absent from this drain.

In that window, an initial-fill or refill rejection/cancellation is thrown instead of an earlier completed store failure. The same gap before the refill decision at line 180 can submit inline work even though the next ordered failure has already completed; if that work blocks, the supposedly immediate failure is delayed indefinitely. Both outcomes violate the failure-precedence and fail-fast guarantees highlighted in this PR.

Reconcile terminal futures in `active` directly before refill and submission-failure decisions, while deduplicating their eventual queue notifications (for example with per-slot publication/consumption state). Add a coordinated regression test that holds completion notification after the future becomes terminal and verifies that the ordered failure wins without another submission.

---

service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
```diff
@@ -0,0 +1,730 @@
...
+      OrderedOutcomeConsumer<O> ordered,
+      TaskRuntime<I, O> runtime) {
+    validatePermits(permits);
+    List<CompletionSlot<O>> active = new ArrayList<>(permits);
+    int next = 0;
+    try {
+      while (next < items.size() && active.size() < permits) {
+        try {
+          runtime.beforeSubmit(active);
+          active.add(runtime.submit(next, items.get(next++)));
```
jchauzi-floe: [major] **`Initial fill` ignores reachable failures**

The initial-fill loop submits until `active.size() == permits` without reconciling a task that became terminal during the preceding `submit`. With an inline or caller-runs executor, `submit(0)` returns with input 0 already failed, but the loop still calls `submit(1)` before the scheduler ever records or surfaces that reachable failure.

This contradicts the immediate first-reachable-failure guarantee and can turn a prompt store error into an indefinite wait if input 1 blocks on the caller thread. It is not only theoretical for the supported executor contract: a direct executor deterministically runs input 1 after an inline input-0 failure, and the class already has explicit caller-runs handling elsewhere.

After every initial submission, reconcile terminal slots and stop before the next submission when `ordered.earliestReachableFailure()` is non-null. Add a regression in which input 0 fails inline and input 1 records a side effect (or blocks), then assert input 1 is never invoked.

---

service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
```diff
@@ -0,0 +1,730 @@
...
+        // Refill the window before delivering this completion so a fast later task is already in
+        // flight — but only while no failure is yet reachable in input order. Once a completion
+        // exposes a contiguous-prefix failure the operation is terminal: a further submit would be
+        // wasted work on an async executor and, on an inline/caller-runs executor, would run that
+        // task on this thread and could stall before the failure is ever surfaced. Skip the refill
+        // in that case and fall through to deliverReady, which throws the reachable failure. A
+        // submission failure on the success-path refill is reconciled against every terminal
+        // completion (see surfaceSubmissionFailure) so a reachable store failure keeps precedence
+        // over the refill's own cancellation or executor rejection.
+        if (ordered.earliestReachableFailure() == null && next < items.size()) {
```
jchauzi-floe: [major] **`recordTerminalCompletions` collapses the window**

`recordTerminalCompletions` can remove several completed siblings from `active`, but this single `if` submits only one replacement. If three initial tasks become terminal before the scheduler handles the first notification, `active` drops from three to zero and only one new task is submitted before delivery.

The lost slots are never recovered: each subsequent single completion is replaced by one submission, so the rest of the batch runs with concurrency one. A deterministic direct-executor run with `permits = 3` and six inputs has only four total submissions at the first result delivery, demonstrating that a burst of completions permanently defeats the configured parallelism. A slow ordered consumer makes the impact worse because only that one replacement can progress while delivery is blocked.

Refill until the window is full (or inputs are exhausted), repeating terminal/failure and cancellation checks between submissions so inline completion remains safe. Add a test that coordinates multiple sibling completions before one notification is consumed and asserts all reopened slots are replenished before ordered delivery.

---

service/src/main/java/ai/floedb/floecat/service/context/PropagatedContext.java
```diff
@@ -0,0 +1,152 @@
...
+   * duration and torn down afterward, so its ambient reads behave as they did on the capturing
+   * thread. Checked exceptions from {@code body} propagate as {@link RuntimeException} (via {@link
+   * ResolvedCallContexts#callWith}).
+   */
+  public <T> T supply(Supplier<T> body) {
+    try (Scope ignored = otel.makeCurrent()) {
+      Map<String, Object> priorMdc = snapshotMdc();
+      replaceMdc(sourceMdc);
+      try {
+        if (call == null) {
```
jchauzi-floe, line 124: [major] **`call == null` preserves a foreign scope**

When the captured caller has no `ResolvedCallContext`, this branch invokes `body` directly instead of installing an explicitly empty scope. Consequently, if the execution thread currently has another `ResolvedCallContexts` scope, ambient principal and engine reads inside `body` see that foreign request even though the captured caller was off-request. OTel and MDC are replaced correctly; the security-sensitive call context is the only carrier left inherited.

This violates the snapshot/isolation contract and can attribute off-request work to the wrong principal. The behavior is deterministic by capturing outside a request and then calling `captured.supply(...)` inside `ResolvedCallContexts.callWith(foreign, ...)`: `currentOrNull()` returns `foreign`.

Always enter `ResolvedCallContexts.callWith(call, ...)`, since that API explicitly accepts `null` and restores the prior scope afterward; only guard `populateMdc(call)` and the call-derived overlays behind `call != null`. Add a test where an empty capture runs on a thread bearing a foreign scope and assert the body sees `null` while the foreign scope is restored afterward.

---

service/src/main/java/ai/floedb/floecat/service/context/PropagatedContext.java
```diff
@@ -0,0 +1,212 @@
...
+        // The caller's LIVE gRPC context, deliberately not fork(): fork() builds a context with no
+        // cancellableAncestor, so getDeadline() returns null and isCancelled() is permanently false
+        // — the body would silently lose the inbound deadline and never observe client
+        // cancellation.
+        // Carrying the live context (rather than leaving the worker's own) is also what isolates
+        // the
+        // body: readers that consult only gRPC keys — AuthResolutionContexts' session/authorization
+        // tokens, PrincipalProvider/EngineContextProvider's fallbacks — see THIS request's values,
+        // and a body captured off-request carries an empty context instead of a foreign one.
+        io.grpc.Context.current(),
```
jchauzi-floe, line 98: [major] **`grpc` snapshot can carry mismatched credentials**

`capture()` obtains the authoritative `ResolvedCallContext` and `io.grpc.Context.current()` independently. That raw gRPC slot is the known unreliable carrier this codebase introduced `ResolvedCallContexts` to avoid: it can be empty during the Quarkus duplicated-context race or stale on a reused worker, even while `ResolvedCallContexts.currentOrNull()` correctly returns request A.

`supply()` then installs request A's explicit principal/engine scope but attaches the independently captured gRPC context. `AuthResolutionContexts.fromInboundContext()` reads the session and authorization values only from those gRPC keys, although the authoritative `ResolvedCallContext` already contains both values. A fan-out task can therefore resolve connector credentials with blank tokens or, worse, request B's stale tokens while the rest of the task runs as request A. This defeats the cross-request isolation this class is meant to provide.

Make the remaining credential reader prefer `ResolvedCallContexts.currentOrNull()` (falling back to gRPC only for legacy callers), or otherwise derive a coherent gRPC carrier from the resolved record while preserving the captured deadline/cancellation and explicitly masking absent header values. Add a regression that captures under `ResolvedCallContexts.callWith(requestA, ...)` while the raw gRPC context is empty and then foreign-request B, and assert that the worker always observes request A's session/authorization values and never B's.

---

jchauzi-floe (approved):

PR #433, opened 2026-08-03
concurrency: bounded metadata-I/O admission with cancellable calls

mrouvroy-floe:
Stages the process-wide metadata-I/O admission infrastructure. It deliberately has no production store caller in this PR; PR #434 wires catalog, namespace, table, and view reads through it.

### Architecture

- **One JVM-lifetime gate** — the JDK-only `floecat-metadata-admission-runtime` module owns only the shared semaphore and configured capacity. Application reloads therefore cannot create independent ceilings, and the parent-first holder retains no application-generation object.
- **One virtual thread per admitted call** — `MetadataIoRunner` acquires admission and starts a fresh virtual-thread task. There is no shared worker pool, executor lifecycle, nested-lease protocol, or application-object registry.
- **Task-owned permit lifetime** — cancellation or caller interruption can promptly abandon the admission/result wait, but the task retains its permit until the downstream operation actually exits. An interruption-insensitive store call therefore continues to count against capacity.
- **Explicit context isolation** — virtual workers do not inherit request-local thread state. The supported telemetry, gRPC, resolved-call, MDC, and context-classloader state is captured and restored explicitly for the operation.

### Configuration and telemetry

- `floecat.query.metadata-io.max-concurrency` defaults to `64`; values must be integers from `1` through `256`, and invalid configured values fail startup.
- The first application generation fixes capacity until JVM restart, preventing overlapping reload generations from admitting through separate gates.
- Typed telemetry exposes capacity, permits in use, admission waiters, and saturated arrivals.
- Downstream clients must enforce their own request/socket timeout: Java cannot safely reclaim a virtual thread still executing non-returning foreign code.

### Scope

This is intentionally infrastructure-only. Publishing and wiring the first admitted repository reads is the responsibility of stacked PR #434.

### Validation

- Process-wide holder and reload-isolation tests
- Permit ownership, cancellation, context propagation, and classloader-retention tests
- Configuration and telemetry contract tests
- Full `make verify` succeeds on the stacked PR #434 tip

---

Part of the GetUserObjects parallel re-cut. Based directly on `main`.


service/src/main/java/ai/floedb/floecat/service/concurrent/MetadataIoRunner.java
```diff
@@ -0,0 +1,647 @@
...
+   * <p>Clearing the latch is enough: {@link #sharedRuntime()} replaces a closed runtime once it is
+   * allowed to build one, so the next caller gets a fresh one without this having to null the
+   * reference and race a concurrent resolution.
+   */
+  static void reopenSharedRuntime() {
+    SHUTDOWN_LATCHED.set(false);
+    // Drop the dead reference too. sharedRuntime() would replace it on the next resolution anyway,
+    // so this changes no outcome; it just stops a closed runtime being pinned until someone happens
+    // to ask, and leaves the next lifecycle in the same state a fresh process starts in.
+    SHARED.updateAndGet(current -> current != null && current.isClosed() ? null : current);
```
jchauzi-floe: [major] **`reopenSharedRuntime()` replaces live admission state**

`RuntimeState.close()` gives up after five seconds when an interruption-insensitive store call is still executing, but `reopenSharedRuntime()` removes that closed state unconditionally. The next call creates a new `RuntimeState` with a full `Semaphore` while the old worker can still be performing metadata I/O under a permit from the discarded semaphore.

That breaks the central process-wide ceiling during the in-JVM restart scenarios this lifecycle explicitly supports. With capacity `1`, one old call can ignore shutdown and continue while the restarted lifecycle admits a second call, producing two concurrent round trips; the old worker can also keep its prior classloader pinned. The restart tests only reopen after all old operations have exited, so they do not cover this case.

Keep outstanding admission state across runtime generations (for example, reuse a stable process-wide semaphore while replacing only the executor), or prevent a new runtime from admitting work until the previous executor has actually terminated. Add a restart test that keeps an `UninterruptibleBlocker` held across `reopenSharedRuntime()` and verifies that new work cannot exceed the configured capacity.

---

service/src/main/java/ai/floedb/floecat/service/concurrent/MetadataIoLifecycle.java
```diff
@@ -0,0 +1,72 @@
...
+   * Closes the shared runtime at shutdown.
+   *
+   * <p>Ordering caveat, unresolved on purpose: {@code ShutdownEvent} fires at the start of
+   * shutdown, and whether the HTTP/gRPC layer has finished draining in-flight calls by then is a
+   * Quarkus-version detail. If it has not, a request still doing store I/O sees a rejected
+   * admission rather than completing. Nothing routes store I/O through this tier yet, so the
+   * question cannot be settled by test here; settle it in the change that wires the first caller,
+   * against a live request, rather than by guessing at container ordering.
+   */
+  void closeSharedMetadataIoRuntime(@Observes ShutdownEvent event) {
```
jchauzi-floe: [major] **`ShutdownEvent` closes before teardown consumers**

Closing from a `ShutdownEvent` observer does not preserve other beans' shutdown-time metadata reads. Quarkus fires this event before Arc invokes application-bean `@PreDestroy` callbacks; `MetadataIoRunner.close()` itself acknowledges this ordering at lines 441–444, so every such callback runs after this observer has latched the shared runtime closed.

Once the next PR routes metadata reads through this tier, any teardown callback that flushes or reads metadata will receive `RejectedExecutionException`. This is the same cross-bean destruction-order failure that the PR description says moving away from an instance `@PreDestroy` is intended to avoid, only earlier in the lifecycle.

Move shared-runtime cleanup to a phase that runs after its consumers have drained, or explicitly coordinate and order all metadata-I/O consumers before closing the runtime. Cover the chosen lifecycle with a container-level test containing a second bean that performs an admitted call during teardown.

---

core/engine-utils/src/main/java/ai/floedb/floecat/engine/concurrent/ProcessWideAdmission.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+package ai.floedb.floecat.engine.concurrent;
+
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Owns admission permits that must survive a Quarkus runtime-classloader restart.
+ *
+ * <p>This is deliberately a normal runtime dependency, rather than application code. Quarkus loads
```
mcusack-floe: [major] **The supposedly persistent admission gate is reloadable**

The latest fix assumes that moving the static semaphore into `floecat-core-engine-utils` places it in the persistent base runtime classloader. The actual Quarkus DEV application model reports this artifact as `(compile, runtime classpath, reloadable)`. Quarkus loads reloadable artifacts in the runtime classloader and recreates them on reload, so the replacement lifecycle receives a new `ProcessWideAdmission.CURRENT` and a full semaphore while an old worker can still hold a permit from the previous copy.

That leaves the previously reported ceiling breach unresolved through a newly introduced classloader assumption. The same-classloader unit test cannot exercise this. Put the gate in a genuinely non-reloadable/parent-first runtime artifact or another process-stable owner, and cover it with an actual Quarkus/dev-mode restart that keeps an old call alive.

service/src/main/java/ai/floedb/floecat/service/concurrent/MetadataIoLifecycle.java
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * Drop the telemetry sink before this container's beans are destroyed.
+   *
+   * <p>The runtime deliberately remains open: neither {@code ShutdownEvent} nor any CDI destruction
+   * phase is after every potential consumer, including {@code @Singleton} beans. Closing there
+   * would reject their teardown metadata I/O. Daemon workers and idle-timeout pool reclamation make
+   * retaining the runtime safe across in-JVM restarts while process-wide admission remains shared.
+   */
+  void clearSaturationSinkAtShutdown(@Observes ShutdownEvent event) {
```
mcusack-floe: [major] **Production shutdown no longer closes the worker runtime**

The response to the teardown-order finding removed the only production call to `closeSharedRuntimeIfStarted()`; this observer now clears only telemetry. That contradicts the PR contract that shutdown reclaims the platform-worker pool across dev-mode and test restarts.

Consequently an old lifecycle remains able to submit through its runtime, idle workers retain the old context classloader until their 60-second timeout, and an interruption-insensitive store call retains it indefinitely. This is distinct from closing too early: cleanup has been removed rather than moved or coordinated. Please provide a post-consumer lifecycle owner, or explicitly revise the design and account for retained runtimes/classloaders.

service/src/main/java/ai/floedb/floecat/service/concurrent/MetadataIoRunner.java
```diff
@@ -0,0 +1,676 @@
...
+ * timeout.</b> A permit is held until the callable returns, and there is no deadline in this tier:
+ * a client that hangs rather than fails pins its permit for the life of the process. With one
+ * process-wide semaphore, {@code capacity} such hangs wedge all metadata I/O — the gauges show
+ * {@code in_use == capacity} but nothing here can name the stuck call. Verify the timeout for each
+ * client as it is wired in.
+ */
+@ApplicationScoped
+public class MetadataIoRunner {
+  private static final Logger LOG = Logger.getLogger(MetadataIoRunner.class);
+  static final String MAX_CONCURRENCY_PROPERTY = "floecat.query.metadata-io.max-concurrency";
```
mcusack-floe, line 47: [major] **The implemented configuration key differs from the advertised key**

The PR documents `floecat.query.metadata_io.max_concurrency`, but the implementation and shipped application properties use `floecat.query.metadata-io.max-concurrency`. These are distinct SmallRye configuration names. An operator following the PR contract and setting the underscore form leaves this key unchanged and silently runs with the default ceiling.

Please make the public contract and implementation use one exact property name.

service/src/main/java/ai/floedb/floecat/service/concurrent/MetadataIoRunner.java
```diff
@@ -0,0 +1,676 @@
...
+    try {
+      parsed = Integer.parseInt(raw.trim());
+    } catch (NumberFormatException badValue) {
+      throw new IllegalStateException(
+          MAX_CONCURRENCY_PROPERTY + " must be an integer; got \"" + raw + "\"", badValue);
+    }
+    // Out of range is rejected, not clamped. Clamping is silent, and 0 or a negative value clamps
+    // to 1 — serialising every metadata round trip in the process behind a single permit, with one
+    // WARN line as the only trace.
+    if (parsed != clampConfiguredCapacity(parsed)) {
```
mcusack-floe: [major] **Out-of-range values are rejected rather than clamped**

The PR specifies a ceiling that is “clamped 1–256”, but startup throws for any value outside that range. Thus `0` and `300` do not become `1` and `256`; they prevent deployment. The application-properties comment documents the rejection behavior, so the implementation and PR contract currently disagree.

Either implement the promised clamping or update the PR contract to state that out-of-range configuration fails startup.

service/src/main/java/ai/floedb/floecat/service/concurrent/MetadataIoRunner.java
mrouvroy-floe: Fixed in 13a5f7e94. ProcessWideAdmission now stores only JDK-owned state in the JVM system-properties table, rather than a static in a reloadable classloader. reopenSharedRuntime() can replace the runtime/executor, but overlapping generations share the same semaphore.

The restart test holds an UninterruptibleBlocker across close/reopen at capacity 1 and verifies that replacement work cannot start until the old call releases its permit.

service/src/main/java/ai/floedb/floecat/service/concurrent/MetadataIoLifecycle.java
mrouvroy-floe: Agreed. Production ShutdownEvent no longer closes the shared runtime; it only clears the CDI-backed telemetry sink. This leaves teardown-time metadata reads available to application and singleton consumers.

closeSharedRuntimeIfStarted() is now limited to controlled non-CDI lifecycles that have already stopped every consumer. MetadataIoLifecycleQuarkusTest covers a teardown consumer performing admitted work after ShutdownEvent.

service/src/main/java/ai/floedb/floecat/service/concurrent/MetadataIoLifecycle.java
mrouvroy-floe: Agreed that simply retaining the runtime was insufficient. The retained pool now uses JDK-owned thread-factory plumbing, and each completed metadata call resets the worker context classloader to the platform loader. Thus an idle worker does not retain a completed pre-reload application generation.

A downstream call that never returns still necessarily retains its own task, classloader, and permit until its client timeout fires; that limitation is explicit in the lifecycle and runner contracts. The PR description now documents this lifecycle choice.

service/src/main/java/ai/floedb/floecat/service/concurrent/MetadataIoRunner.java
mrouvroy-floe: Agreed. The implementation and shipped configuration use floecat.query.metadata-io.max-concurrency. The PR description has been corrected from the obsolete underscore spelling.

service/src/main/java/ai/floedb/floecat/service/concurrent/MetadataIoRunner.java
mrouvroy-floe: Agreed. Values outside 1–256 are rejected; they are not clamped. The strict parser is now shared by CDI startup and the supported non-CDI runner path, with direct-construction coverage for blank, malformed, unresolved, and out-of-range values. The PR description now says that invalid values fail startup.

core/engine-utils/src/main/java/ai/floedb/floecat/engine/concurrent/ProcessWideAdmission.java
mrouvroy-floe: Fixed in 13a5f7e94. The gate no longer depends on the classloader that defines ProcessWideAdmission: its immutable state is held in the JDK-owned JVM system-properties table, keyed by the process-wide configuration. A reloadable application copy therefore obtains the same semaphore while an older generation still holds a permit.\n\nThe held-call restart test verifies that exact overlapping-generation invariant at capacity 1. The implementation is intentionally independent of the classloader placement of floecat-core-engine-utils.

service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
```diff
@@ -192,6 +192,17 @@ static <I, O> void forEachOrdered(
...
+    // no-fan-out-under-admission invariant is checked rather than only documented. One surface,
+    // not the invariant: supplyAsync, a ManagedExecutor, a Mutiny hop and a store client's own pool
+    // all reach the same deadlock unguarded. A tripwire on the path most likely to take it, not
+    // proof the tier catches the mistake. Units dispatched to
+    // other threads acquire their own permits while the caller still holds one, so a fan-out
+    // started inside an admitted operation can wedge the process-wide ceiling.
+    // Unconditional, including for zero inputs. Gating this on item count would make a wiring
+    // mistake data-dependent: silent wherever a caller happens to have no rows, and fatal in
+    // production the first time it does. The caller is structurally wrong either way.
+    MetadataIoRunner.rejectFanOutFromAdmittedOperation("BoundedFanout.forEachOrdered");
```
jchauzi-floe: [blocker] **The admission boundary leaks into generic fan-out**

This makes the general-purpose fan-out scheduler depend on `MetadataIoRunner`, while the surrounding comment explicitly concedes that the check catches only this one dispatch surface and misses `supplyAsync`, managed executors, Mutiny hops, and client-owned pools. That is a scattered tripwire, not an enforceable invariant. More importantly, needing it at all shows that a permit can wrap orchestration which is still allowed to create more store work; the ThreadLocal re-entrancy protocol in `MetadataIoRunner` is another compensating mechanism for the same misplaced ownership boundary.

Move admission to the one canonical leaf where an actual metadata-store round trip begins, after fan-out has selected and dispatched each unit, and ensure only that layer wraps the call. Parent repository/orchestration methods should never hold a store permit while they schedule children. With that boundary, delete this dependency, `rejectFanOutFromAdmittedOperation`, and the ThreadLocal re-entry mode. If arbitrary admitted scopes truly must remain, enforce their dispatch contract at an executor/context boundary rather than adding one-off checks to individual concurrency utilities.

---

service/src/main/java/ai/floedb/floecat/service/concurrent/CancellableCallRunner.java
```diff
@@ -0,0 +1,626 @@
...
+      BooleanSupplier cancelled,
+      BooleanSupplier closed,
+      Supplier<T> operation,
+      FailureMessages messages,
+      Runnable onSaturation) {
+    // Allocate the lease BEFORE acquiring: after acquire() a permit exists that only the lease can
+    // return, so an allocation Error between the two would consume it for the life of the process.
+    PermitLease permitLease = new PermitLease(permits);
+    acquire(permits, cancelled, closed, messages, onSaturation, ADMITTED_CANCELLATION_POLL_MILLIS);
+    SubmittedCallHandle<T> submitted =
```
jchauzi-floe: [blocker] **Do not ship a second concurrency runtime for one blocking call**

A single call is represented simultaneously by a `CompletableFuture`, a `FutureTask`, `CallLifecycle` with three atomics and a volatile task, and an idempotent lease with several release sites. Start, cancel, shutdown discard, executor rejection, and completion races each update different subsets of those objects. The 626-line implementation and its new 1,123-line test file are evidence that this abstraction is not containing complexity; it is manufacturing a bespoke concurrency protocol whose invariants have to be reconstructed from load-bearing comments.

Make one small `AdmittedTask<T>` the sole owner of result, runner cancellation, and permit release, and await that task directly; keep the semaphore polling as a separate, boring acquisition helper. Prefer the JDK `FutureTask` completion/cancellation arbitration. If queued-versus-running cleanup genuinely needs extra state, encode it once inside that task (for example, a compact lifecycle enum) and make its one terminal transition release the lease. Delete the mirrored `CompletableFuture`, `SubmittedCallHandle`, and `CallLifecycle`. If queue tombstone removal is required, accept the actual `ThreadPoolExecutor` contract instead of accepting generic `Executor` and recovering it with a downcast.

---

core/engine-utils/src/main/java/ai/floedb/floecat/engine/concurrent/ProcessWideAdmission.java
```diff
@@ -0,0 +1,88 @@
...
+   * <p>The first runtime in a JVM establishes the ceiling. Later runtime generations deliberately
+   * retain it even if their configuration differs: replacing the semaphore could let a caller that
+   * already resolved the old gate admit work outside the new gate. A full JVM restart is required
+   * to apply a changed ceiling.
+   */
+  public static State resolve(int capacity) {
+    if (capacity < 1) {
+      throw new IllegalArgumentException("process-wide metadata-I/O capacity must be positive");
+    }
+    synchronized (System.getProperties()) {
```
jchauzi-floe: [blocker] **System properties cannot be the JVM object registry**

This commandeers the JVM system-properties table as a two-entry object store for an `Integer` and a `Semaphore`, then synchronizes unrelated code on the global `Properties` object. System properties are a string configuration boundary: non-string values are invisible to `getProperty` and can break code that lists, serializes, copies, or replaces the table. The paired keys, identity checks, and conditional clearing form an implicit cross-classloader protocol with no typed owner. Putting this service-specific mechanism in `core/engine-utils` and adding a module dependency does not make its lifetime explicit.

Give the cross-generation gate a real JVM owner: a deliberately parent-loaded holder or Quarkus runtime integration with a typed atomic `State` and a narrow resolve/observe API. Configure that component's classloading explicitly if parent ownership is required, while leaving system properties solely as boot-time strings. Keep the metadata-specific implementation in the service/runtime layer instead of leaking it into generic engine utilities.

---

service/src/main/java/ai/floedb/floecat/service/concurrent/MetadataIoRunner.java
```diff
@@ -0,0 +1,651 @@
...
+        // establish a different configured ceiling without replacing a live gate.
+        ProcessWideAdmission.clearIfIdle(closing.permits);
+      }
+    }
+  }
+
+  /**
+   * Re-arm the shared runtime for a new application lifecycle.
+   *
+   * <p>A no-op under Quarkus: this is an application class, so a dev-mode reload or
```
jchauzi-floe: [major] **The runtime lifecycle model contradicts itself**

This method says a Quarkus reload creates a fresh `Class` with fresh statics and therefore makes `reopenSharedRuntime` a no-op; the `SHARED` documentation at line 521 says the opposite and justifies replaceable static state because Quarkus reuses the runtime classloader. `MetadataIoLifecycle` likewise orders startup around a latch supposedly left by the previous lifecycle. At the same time the class supports a CDI-wide facade, public non-CDI construction, a test-only owned runtime, controlled same-classloader close/reopen, per-call runtime replacement, and a manually reset static telemetry sink. The result is a 651-line facade with no coherent owner, and several branches exist only to simulate lifecycle shapes that no production caller currently establishes.

Model the lifetimes separately and explicitly: one immutable JVM-lifetime admission gate survives reloads, while each application generation receives a fresh facade/executor owner from CDI. For non-CDI embedding, return an explicit `AutoCloseable` runtime instance rather than mutating the production singleton through static close/reopen hooks. Inject typed configuration and telemetry into the generation owner, and construct isolated owners directly in tests. That removes `ownsRuntime`, per-call shared-runtime resolution, the latch/CAS replacement loop, startup priority coupling, and the static sink reset protocol.

---

service/src/main/java/ai/floedb/floecat/service/concurrent/BoundedFanout.java
```diff
@@ -192,6 +192,9 @@ static <I, O> void forEachOrdered(
       IntConsumer completionBarrier) {
     // Before capture() or any allocation: an invalid bound should cost nothing and name itself.
     validatePermits(permits);
+    // Check even an empty batch so invalid wiring fails deterministically rather than only when
+    // production data happens to contain work.
+    MetadataIoRunner.rejectFanOutFromAdmittedOperation("BoundedFanout.forEachOrdered");
```
jchauzi-floe: [blocker] **Admission is enforced at the wrong ownership boundary**

This generic fan-out scheduler now depends on `MetadataIoRunner` to reject one particular dispatch shape. The guard is inherently incomplete: `CompletableFuture`, Mutiny, managed executors, and client-owned pools can still dispatch from the same admitted scope. It also makes orchestration know about metadata admission while failing to make the invariant enforceable. The re-entrant ThreadLocals, forked leases, and lineage invalidation elsewhere in this PR are compensating mechanisms for the same misplaced boundary.

Move admission to the canonical leaf that performs exactly one metadata-store round trip. Fan-out should happen first, and each dispatched leaf should acquire exactly once. Then delete this dependency, `rejectFanOutFromAdmittedOperation`, `HELD_ADMISSION`/`CURRENT_ADMISSION`, `callAlreadyAdmitted`, and the fork/invalidate lease protocol. If arbitrary admitted scopes truly must remain, enforce their dispatch contract at an executor/context boundary rather than adding one-off tripwires to individual concurrency helpers.

---

service/src/main/java/ai/floedb/floecat/service/concurrent/MetadataIoRunner.java
```diff
@@ -0,0 +1,626 @@
...
+      // acquire plus a watchdog to interrupt waiters.
+      this.permits = sharedPermits != null ? sharedPermits : new Semaphore(capacity);
+    }
+
+    /**
+     * Create the pool on first use. {@code closed} is sticky and there is no reopen path, so this
+     * is start-once-then-throw, never a restart: a closed runtime stays closed for the life of the
+     * process.
+     */
+    private synchronized ThreadPoolExecutor start() {
```
jchauzi-floe: [blocker] **The worker runtime duplicates the semaphore and manufactures lifecycle complexity**

Admission is acquired before dispatch, yet every runtime also creates `capacity` platform workers and a `capacity` queue. This second concurrency mechanism is what requires queued-task removal, shutdown-discard callbacks, `closingExecutor`, sticky closure, CAS replacement, bounded shutdown waits, and worker classloader cleanup. The result is two 600-line production classes plus a new 1,406-line test file for executing one blocking call; the abstraction is not containing complexity. The nested path already runs the same blocking operation on a virtual thread, so the codebase is carrying the heavier platform-pool design without a demonstrated constraint that requires it.

After moving admission to the store leaf, acquire the process-wide permit and start one task-owned virtual thread for that admitted call. Let that task alone own its result, interrupt target, and idempotent permit release. A completed virtual thread has no idle pool or queue to retain, so `MetadataIoExecutors`, queue tombstone/discard handling, executor-generation drain state, and most close/reopen machinery disappear. If a store client genuinely forbids virtual threads, document that concrete constraint and introduce the smallest dedicated dispatcher that satisfies it instead of preserving this generic runtime.

---

service/src/main/java/ai/floedb/floecat/service/concurrent/MetadataIoRunner.java
```diff
@@ -0,0 +1,626 @@
...
+  /**
+   * The runtime this facade owns, or null when it follows the process-wide one.
+   *
+   * <p>Never cache the shared runtime in a field: a facade outlives it. The
+   * {@code @ApplicationScoped} bean and any long-held reference would keep serving the instance
+   * captured at construction, which a shutdown closes for good — {@code closed} is sticky, and
+   * {@link #reopenSharedRuntime()} lowers a flag; it cannot revive a closed runtime. Resolve per
+   * use instead.
+   */
+  private final RuntimeState ownedRuntime;
```
jchauzi-floe: [major] **One facade conflates CDI, process, test, and embedding lifetimes**

`MetadataIoRunner` simultaneously acts as an application-scoped bean, a public non-CDI facade, a package-private owned test runtime, and a controller for a replaceable static runtime. Supporting all of those modes introduces `ownsRuntime`, per-call runtime resolution, `SHARED`, `SHUTDOWN_LATCHED`, public close/reopen hooks, startup-priority coordination, and a separately reset static telemetry sink. The tests themselves acknowledge that Quarkus reloads receive fresh service-class statics, so much of this state machine exists only for a hypothetical same-classloader embedding lifecycle; there is no production caller of the close hook.

Separate the lifetimes explicitly. Keep a tiny immutable JVM-lifetime admission gate, make the CDI runner generation-scoped with typed configuration and injected telemetry, and provide a distinct `AutoCloseable` runtime only if a real non-CDI embedding needs controlled shutdown. Tests should construct that owner directly. This removes the mode flag, static runtime replacement protocol, startup latch choreography, and global callback reset instead of asking every method to reason about all ownership modes.

---

service/src/main/resources/application.properties
```diff
@@ -14,6 +14,10 @@
 
 ## application name
 quarkus.application.name=floecat-service
+# ProcessWideAdmission must survive a dev-mode or in-JVM Quarkus reload. Its artifact has only JDK
+# dependencies, so loading it parent-first makes its static holder stable without pinning an
+# application generation.
+quarkus.class-loading.parent-first-artifacts=ai.floedb.floecat:floecat-core-engine-utils
```
jchauzi-floe: [major] **A metadata gate makes an unrelated shared artifact parent-first**

To preserve one 70-line metadata semaphore holder, this marks the entire `floecat-core-engine-utils` artifact parent-first. That artifact also contains the AWS client refresh machinery and engine identity utilities and is consumed by storage, connectors, scanners, and other core modules. The change therefore gives unrelated present and future classes a process-stable class-identity constraint that is visible only in the service deployment configuration. This is a broad classloader boundary for a narrow service concern.

Put the process-lifetime holder in a dedicated minimal runtime/bootstrap artifact (or an explicit Quarkus runtime integration) and mark only that artifact parent-first. Keep the API typed and narrow, but do not turn a general-purpose utility jar into the JVM ownership boundary for metadata admission.

---

jchauzi-floe (approved):

PR #434, opened 2026-08-03
concurrency: automatic metadata-I/O admission + admission-free fan-out stage

mrouvroy-floe:
Wires automatic metadata-I/O admission into the catalog, namespace, table, and view repository families, and introduces the configured fan-out stage consumed by the next resolver PR.

### Repository admission boundary

- **One typed read seam** — `RepositoryReads.bind(...)` creates pointer/blob capabilities from a `ReadPolicy`. `MetadataRepositoryFactory` composes CDI repositories with `MetadataResourceReader`; manual constructors keep direct reads for compatibility.
- **Leaf-store admission** — query reads acquire the PR #433 process-wide permit only at actual pointer/blob store access. Catalog, namespace, table, and view read surfaces are covered.
- **Mutation protocols remain raw** — create/update/delete transactions and their prerequisite reads use the raw stores, so admission and request cancellation cannot split a write protocol.
- **Cache-aware placement** — immutable cache probes stay outside admission. Only the single-flight loader's cold backend miss acquires a permit; cache hits and followers consume none.

### Cancellation and fan-out staging

- `BaseServiceImpl.run()` binds a fresh live cancellation signal for every unary RPC subscription, combining subscriber termination with the captured gRPC context. A disconnected request abandons a queued admission wait and cannot start its backend read later.
- Lazy PlannerStats target streaming carries the same per-subscription signal into its iterator and binds it only around the admitted table lookup. Cancellation escapes the iterator instead of being materialized as a corrupted or per-target error result.
- `PropagatedContext` carries cancellation across supported dispatch. PR #435 adds resolver and user-object-stream-specific propagation and consumes `MetadataFanout`.
- `MetadataFanout.serial()` and `MetadataFanout.concurrent(permits)` expose ordered cancellable and uncancellable operations while keeping `BoundedFanout` internal. Serial work runs inline; concurrent work is bounded by its configured fan-out width before each unit reaches repository admission.

### Scope

Admission covers catalog, namespace, table, and view reads. Snapshot, stats-store, current-pointer, and table-root reads remain outside this ceiling in this PR. PlannerStats' nested table lookup is admitted; its stats orchestration/storage path remains outside the ceiling.

### Validation

- Reflection-backed wiring coverage for all four repository families
- Cache-hit/follower and cold-miss admission tests
- Mutation prerequisite regression coverage for update and both delete paths
- Saturated unary and lazy-stream cancellation regressions verify that waiters drop and queued backends never start
- Cancellation, context, fan-out, repository, and runner tests
- Full 35-module `make verify` succeeds

---

Part of the GetUserObjects parallel re-cut. Stacked on PR #433 (`mr-guo-2-admission`).


service/src/main/java/ai/floedb/floecat/service/repo/impl/CatalogRepository.java
```diff
@@ -73,54 +74,65 @@ public boolean deleteWithPrecondition(ResourceId catalogResourceId, long expecte
...
 
   /** Pointer-only meta (no blob HEAD, blank etag) for metadata-graph consumers. */
+  @BoundMetadataIo
   public MutationMeta pointerMetaForSafe(ResourceId catalogResourceId) {
     return repo.pointerMetaForSafe(
         new CatalogKey(catalogResourceId.getAccountId(), catalogResourceId.getId()));
   }
 
   /** Blob-direct read for graph hydration from resolved metadata; empty if the blob moved. */
+  @BoundMetadataIo
```
jchauzi-floe: [major] **`getByBlobUri` admits before the cache**

`getByBlobUri` delegates to `GenericResourceRepository.getByBlobUri`, which first consults `ImmutableBlobCache`. A hit performs no store access, and concurrent misses for one URI are single-flighted by Caffeine, but this outer `@BoundMetadataIo` makes every hit and every follower acquire a process-wide permit before reaching that cache. With the default ceiling, 64 callers for one cold URI can therefore consume all 64 permits while only one backend round trip is running.

That turns the I/O gate into a global cache/request throttle: unrelated pointer reads queue behind threads that are either returning cached data or waiting for the same cache loader, and the admission metrics report those callers as in-flight I/O. A slow single-flight load can consequently stall the whole admitted metadata surface even though the store has only one request from this group.

Apply admission to the cache loader/miss path (or split out an admitted uncached loader) so cache probes, hits, and single-flight followers remain outside the permit. The same outer annotation has this problem on the corresponding `getByBlobUri` methods in `NamespaceRepository`, `TableRepository`, and `ViewRepository`.

---

service/src/test/java/ai/floedb/floecat/service/concurrent/MetadataIoAdmissionWiringTest.java
```diff
@@ -0,0 +1,163 @@
...
+        CatalogRepository.class,
+        read("getById", ResourceId.class),
+        read("getByName", String.class, String.class),
+        read("list", String.class, int.class, String.class, StringBuilder.class),
+        read("count", String.class),
+        read("metaFor", ResourceId.class),
+        read("metaFor", ResourceId.class, Timestamp.class),
+        read("metaForSafe", ResourceId.class),
+        read("pointerMetaForSafe", ResourceId.class),
+        read("getByBlobUri", String.class),
```
jchauzi-floe: [minor] **Wiring guard omits most read methods**

The PR promises that this test guards every exposed metadata-read entry point, but `assertMetadataReadsBound` checks only the two `metaFor` overloads and `metaForSafe`, plus the five special cases above it. It does not guard `getById`, name/path reads, `list`, `count`, ID/ref reads, `listTokenAfter`, `getByBlobUri`, or `getByBlobUriLive`.

All of those methods are annotated today, but removing any of those annotations later would leave this test green, so the central guarantee that repository callers cannot accidentally bypass admission is not protected.

Enumerate every current read signature, or reflect over public repository methods with an explicit allowlist for constructors, writes, and pure helpers so a newly added or accidentally unbound read fails the wiring test.

---

service/src/main/java/ai/floedb/floecat/service/concurrent/MetadataIoAdmissionInterceptor.java
```diff
@@ -0,0 +1,95 @@
...
+    // in front of this — so every read below takes the uncancellable path until then.
+    BooleanSupplier cancelled = PropagatedContext.currentCancellation();
+    try {
+      // Repository reads are thread-safe blocking I/O, so run each on the shared admission pool
+      // (tier 2) rather than the caller thread: the ceiling bounds the pool, the caller's fan-out
+      // (tier 1) supplies parallelism, and — crucially — a cancellable read that blocks mid-flight
+      // is abandoned promptly on cancel (the caller returns while the worker unwinds) instead of
+      // pinning the request thread. A read that re-enters admission on the pool worker reuses that
+      // worker's permit and runs inline (see MetadataIoRunner), so nesting does not re-dispatch.
+      return cancelled == null
```
jchauzi-floe: [blocker] **CDI interceptor owns an off-thread, abandonment-capable call**

**Issue**

This interceptor is not merely enforcing a policy around a synchronous call: it dispatches InvocationContext.proceed() to another thread and, on cancellation, can return to the proxy caller while that invocation is still running. That hidden lifetime/thread boundary is why the implementation needs checked-exception tunneling and extensive re-entrancy/deadlock rules, and it makes every annotated repository method silently asynchronous internally. CDI interception is the wrong abstraction for owning abandonment-capable work; the resulting control flow is magical and much harder to reason about than the repository/store operation it wraps.

**Remedy**

Move the off-thread admission call into one explicit metadata-read module owned by the repository layer (for example, a MetadataResourceReader that owns pointer reads, cache probes, and blob loads through a typed admission service). Repository facades should delegate to that module normally; do not dispatch InvocationContext.proceed(). This keeps callers unable to bypass admission without turning the CDI invocation itself into a background task, and it lets the annotation/interceptor, exception tunnel, re-entrant interceptor nesting, and reflection-based method-coverage machinery disappear rather than be maintained as parallel control flow.

---

service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java
```diff
@@ -48,6 +49,8 @@ public class GenericResourceRepository<T, K extends ResourceKey> extends BaseRes
   private static final Logger log = Logger.getLogger(GenericResourceRepository.class);
 
   private final ResourceSchema<T, K> schema;
+  // Null for repository families outside metadata-I/O admission, and for direct unit-test setup.
```
jchauzi-floe: [major] **Nullable admission bean leaks a feature policy into the generic repository**

**Issue**

The cache-miss exception to the annotation scheme leaks a feature-specific, nullable admission bean into the shared generic repository. Admission now depends on which constructor overload happened to be used: null means bypass, while a non-null object whose Java body is only loader.get() becomes admitted through CDI magic. The four public two-argument repository constructors explicitly select the bypass mode, so this is not actually an intrinsic property of the read boundary, and the growing GenericResourceRepository now carries a policy that applies to only four families.

**Remedy**

Extract blob hydration/cache loading behind a non-null, typed reader policy with explicit admitted and unadmitted implementations, selected when composing each repository family. Prefer making the admitted implementation call the admission service directly. That removes the null branch, constructor churn, subclassable identity bean, and service.concurrent dependency from GenericResourceRepository while making excluded families an explicit composition choice instead of a silent fallback.

---

service/src/main/java/ai/floedb/floecat/service/concurrent/MetadataFanout.java
```diff
@@ -0,0 +1,121 @@
...
+
+  /**
+   * Apply {@code unit} to each element in input order, delivering each result to {@code consumer}
+   * as its input-order prefix completes. The first failure reachable in input order — from a unit
+   * or from {@code consumer} — propagates immediately; concurrent siblings are cancelled rather
+   * than drained. {@code permits} bounds how many of this request's units run at once (concurrent
+   * mode only); {@code cancelled} must be non-blocking and thread-safe, and observed cancellation
+   * throws {@link CancellationException}.
+   */
+  public static <I, O> void forEachOrdered(
```
jchauzi-floe: [major] **`MetadataFanout` leaks `BoundedFanout`'s identity-based no-cancellation protocol**

**Issue**

MetadataFanout claims to be the canonical public stage, but it does not encapsulate BoundedFanout's no-cancellation contract. BoundedFanout recognizes its NEVER_CANCELLED supplier by identity to avoid polling; this facade exposes only a BooleanSupplier parameter and does not provide a no-cancellation overload or canonical token. Callers therefore either pass an equivalent lambda and pay the polling path (as this PR's tests do) or must reach through the facade to an internal scheduler detail. The wrapper is not earning its boundary if consumers still need to understand the engine's identity protocol.

**Remedy**

Make MetadataFanout the sole public API: add map/forEach overloads for uncancellable work that internally pass BoundedFanout.NEVER_CANCELLED, keep BoundedFanout package-private, and align BoundedFanout's stale documentation with the admission-free facade. If cancellation modes continue to grow, expose a small typed cancellation policy owned by MetadataFanout rather than a magic BooleanSupplier identity.

---

service/src/main/java/ai/floedb/floecat/service/repo/impl/CatalogRepository.java
```diff
@@ -74,41 +93,55 @@ public boolean deleteWithPrecondition(ResourceId catalogResourceId, long expecte
   }
 
   public Optional<Catalog> getById(ResourceId catalogResourceId) {
```
jchauzi-floe, line 80: [blocker] **Admission is a scattered convention instead of a repository invariant**

Every read entry point in this class—and again in `NamespaceRepository`, `TableRepository`, and `ViewRepository`—must remember to wrap itself in `metadataReads.read(...)`, while `getByBlobUri` is a special exception whose cold path is admitted through a second policy in `GenericResourceRepository`. That spreads one safety-critical rule over dozens of methods and five files. A new read method can compile and silently bypass the ceiling, and `MetadataIoAdmissionWiringTest` only exercises one representative `count` method per family, so it does not close that hole. The duplicated CDI constructor, concrete-to-interface cast, policy constructor, and silent two-argument `DIRECT` fallback are further evidence that the ownership boundary is wrong.

Move admission to one canonical composed read boundary. For example, give `BaseResourceRepository`/`GenericResourceRepository` separate admitted read delegates (or read-only `PointerStore`/`BlobStore` decorators) while retaining raw delegates for writes; the immutable cache should invoke the admitted blob delegate only from its loader. Then these four facades return to direct delegation, cache hits/followers stay admission-free naturally, and the per-method wrappers, constructor casts, and `BlobLoadPolicy` bridge can disappear. If legacy direct construction must remain, make it an explicit named factory or require `MetadataReadPolicy.DIRECT` at the call site rather than silently changing the semantics of the same repository type.

---

service/src/main/java/ai/floedb/floecat/service/concurrent/MetadataFanout.java
```diff
@@ -0,0 +1,148 @@
...
```
jchauzi-floe: [major] **The boolean mode exposes two schedulers through one invalid-state API**

`concurrent` switches this method between delegation to `BoundedFanout` and a separately implemented inline loop, and it makes `permits` conditionally meaningful: serial callers are expected to pass an ignored value (the tests deliberately pass `0`), while concurrent callers reject the same value. Four overloads, a process-wide executor, conditional validation, and duplicated cancellation/ordering logic are a large public surface for what is mostly a thin wrapper. Callers can construct nonsensical combinations, and future changes must keep two orchestration implementations behaviorally aligned.

Use one canonical ordered-fan-out abstraction. Either deepen `BoundedFanout` with explicit serial support and expose that single implementation, or provide separate serial and concurrent entry points/configurations so serial has no permit argument and concurrent construction requires a positive bound. This removes the boolean branch and ignored parameter, gives executor ownership one home, and makes the legal execution model visible in the API instead of encoded by convention.

---

jchauzi-floe (approved):

PR #436, opened 2026-08-03
catalog bundle: decompose chunk assembly into structural stages

mrouvroy-floe:
## Summary

Make `GetUserObjects` bounded, cancellable, ordered, and easier to follow.

## RPC flow

The request is assembled in this order:

```mermaid
flowchart TD
    A["Client calls GetUserObjects<br/>UserObjectBundleService"]
    B["Resolve requested names and IDs<br/>RelationResolutionMemo"]
    C["Choose the exact table snapshot<br/>QueryInputResolver + SnapshotPinMemo"]
    D["Protect snapshot/blob references<br/>QueryPinCommitter"]
    E["Warm statistics with shared limits<br/>StatsProviderFactory + StatsWarmLimiter"]
    F["Build relation metadata<br/>RelationBundleBuilder"]
    G["Run optional engine decoration<br/>EngineRelationDecorator"]
    H["Reuse payload when token matches<br/>RelationPayloadPolicy"]
    I["Frame results in request order<br/>BundleStreamFramer"]
    J["Send final counts and complete"]

    A --> B --> C --> D --> E --> F --> G --> H --> I --> J

    X["Client disconnects or stream is cancelled"]
    Y["Stop admission waits and release temporary pin/blob protections<br/>StreamTelemetryState + CancelledQueryPinCleanup"]
    X -.-> Y
    Y -.-> C
    Y -.-> E
    Y -.-> F
```

## What changed

- Split the former bundle coordinator into focused stages for resolution, snapshot protection, statistics, relation assembly, engine decoration, payload reuse, and response framing.
- Run independent relation work concurrently only when the overlay, decorator, and other callback-bearing collaborators explicitly support worker-thread execution.
- Preserve original request order and the existing chunk/header/end-message protocol.
- Memoize repeated relation and snapshot-pin resolution within one request.
- Commit snapshot references before emitting each response batch, and release temporary roots on cancellation and every failure path.
- Bound statistics warming across active requests while retaining a per-request fairness limit.
- Make cancellation state explicit so cancelled requests stop waiting and do not start abandoned backend work.
- Reuse payloads only when the client has the exact schema, engine decoration, and content version being served.
- Replace confusing internal names with responsibility-based names such as `RelationPayloadPolicy`, `StatsWarmLimiter`, `QueryPinCommitter`, `BundleStreamFramer`, and request-scoped `Memo` types.
- Remove the obsolete mutable-map snapshot-pin compatibility adapter; callers now use the request-scoped `SnapshotPinMemo`.

## Compatibility

Wire format and query semantics are unchanged. Explicit snapshot and AS-OF behavior remain exact, repeated CURRENT references remain stable within a request, and response ordering is preserved.

## Verification

- `make fmt`
- `git diff --check`
- `mvn -pl service -am -DskipTests compile`
- `make verify` (full 35-module reactor, including service and REST integration tests)


service/src/main/java/ai/floedb/floecat/service/query/resolver/QueryInputResolver.java
```diff
@@ -169,6 +183,43 @@ public SnapshotSet snapshotSet() {
     }
   }
 
+  /**
+   * Per-request single-flight cache for CURRENT-snapshot table pins. Failed lookups are evicted so
+   * a later caller may retry, while successful pins remain frozen for the cache lifetime.
+   */
+  public static final class CurrentSnapshotPinCache {
```
jchauzi-floe: Seems like this could be a `SnapshotPinCache`, nothing inherent to the current snaphsot?

service/src/main/java/ai/floedb/floecat/service/query/resolver/LegacyCurrentSnapshotPinCaches.java
```diff
@@ -0,0 +1,89 @@
...
+
+import ai.floedb.floecat.common.rpc.ResourceId;
+import ai.floedb.floecat.query.rpc.TablePin;
+import ai.floedb.floecat.service.concurrent.Futures;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+/** Adapts the legacy completed-pin map contract to the resolver's single-flight cache. */
+final class LegacyCurrentSnapshotPinCaches {
```
jchauzi-floe: Is legacy needed?

jchauzi-floe (approved):

PR #444, opened 2026-08-10
Support catalog-vended storage credentials for Iceberg REST catalogs

asoufane-ops:
## Summary

Makes Iceberg REST credential vending actually reachable, so a connector against a
delegating catalog (Polaris) can capture data with **no storage authority configured**.

Three gaps, each of which independently broke the path:

1. **Connectors couldn't surface vended credentials.** `FloecatConnector` gains an optional
   `vendStorageCredentials`; `IcebergConnector` implements it by copying the `s3.*` keys off a
   delegated `loadTable`'s `FileIO` via an explicit allowlist.
2. **The service never asked.** `StorageAuthorityServiceImpl` now falls back to the source
   catalog when no authority covers a location — and resolves the table from the **reconcile
   lease**, because the reconcile worker sends `table_id: None` and the previous
   `request.hasTableId()` guard made the whole path unreachable from the only caller that matters.
3. **Leased data files were rejected as out of scope.** A lease scoped credentials to the table
   location, so a file group planned over data registered in place (`add_files`) was denied its
   own files. Requests for a file the lease actually planned are now admitted, scoped to that file.

## Type of Change

## Testing

Validated end to end on `floedb-dev-eks`: 8/8 TPC-H tables captured from Polaris on vended
credentials alone, 44 files, 0 errors, 6m47s, with the `polaris-test-data` storage authority
disabled. 94 × `vended storage credentials from source catalog`, 0 `INVALID_ARGUMENT`,
0 `PERMISSION_DENIED`.

- [X] `make fmt` — 0 reformatted
- [ ] `make verify` — not run; Docker daemon unavailable locally. `mvn -pl service test` passes:
      **1547 tests, 0 failures**

New tests: 6 in `IcebergConnectorVendedCredentialsTest` (including one asserting the catalog's
own OAuth token is never copied into storage credentials), 4 in `StorageAuthorityServiceImplTest`,
2 in `ServerSideStorageConfigResolverTest`. The two behaviour-defining tests were verified
non-vacuous by stubbing the new logic and confirming they fail — the scope test fails with
`expected INVALID_ARGUMENT but was PERMISSION_DENIED`, the exact production symptom.

- [X] No breaking changes
- [X] No config/env var changes

Every new line sits inside `if (authority == null)`. Any request where a storage authority
matches takes the identical prior path, so no working configuration changes behaviour. The only
observable difference is for requests that **already failed** with "no storage credential
authority is configured": they may now succeed, or fail with a different error if the catalog
refuses.

## Checklist

- [X] This PR does not disclose a security vulnerability publicly

## Additional Notes

**Wants a follow-up before production traffic.** `vendFromSourceCatalog` builds a connector and
performs an OAuth exchange *per file group* — 94 vends for 44 files in the run above. It wants a
cache keyed on the table and bounded by the credential expiry, which the vended credential already
carries. Left out deliberately: the invalidation policy is a design decision, not an
implementation detail, and it's called out in a comment at the call site.

**Reviewer guidance.** The security-relevant line is the allowlist in `IcebergConnector` — a
delegated `loadTable` returns the catalog's own bearer `token` in the same property map as the
storage credentials, so the copy enumerates three keys rather than filtering. `neverCopiesTheCatalogAuthToken`
is what keeps it that way.

The scope change is also worth attention: it admits a requested location only on **exact**
membership in the lease's planned file paths, never a prefix match, so a caller can't widen its own
scope by asking for a parent directory. `parentPrefixOfALeasedFileIsStillOutOfScope` pins that.

**Squash note.** `173da9ebf` and `91565ddbb` share a title ("reconciler: cover the delegation
vending bypass") — worth collapsing on merge.


reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ServerSideStorageConfigResolver.java
```diff
@@ -253,6 +263,19 @@ private ConnectorConfig resolveStorageBackedConnector(
...
+    // A catalog that vends its own credentials needs no storage authority: loadTable returns
+    // storage-credentials and the catalog client's FileIO uses them. Returning the config untouched
+    // is the same thing the null-locationPrefix branch below already does for non-S3 URIs, which is
+    // how GCS- and Azure-backed connectors run today without an authority -- this makes that
+    // existing path reachable for s3:// when the caller has explicitly asked for delegation.
+    //
+    // Deliberately not a fallback on vending failure: if delegation is declared we never call the
+    // authority service at all, so a missing authority cannot mask a catalog that silently ignored
+    // the header. The failure then surfaces from the catalog or the read itself.
+    if (isNonBlank(config.options().get(ICEBERG_ACCESS_DELEGATION_PROP))) {
```
mcusack-floe: **Severity: High — delegation configured through `auth.headerHints()` is missed.**

This checks only `config.options()`, but `IcebergConnectorFactory` also converts `config.auth().headerHints()` into `header.<name>` catalog properties. A connector configured through the supported header-hint route therefore falls through and still calls storage-authority vending. Please derive an effective delegation setting from both sources and add a test for the header-hint configuration path.

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityServiceImpl.java
```diff
@@ -598,6 +656,120 @@ private CredentialScope resolveDiscoveryTableLocation(
...
+   * and a fresh HTTP client. Vending happens per file group, so this wants a cache keyed on the
+   * table and bounded by the credential expiry before it carries production traffic. Left out here
+   * because the invalidation policy is a design decision, not an implementation detail.
+   */
+  private ResolveStorageAuthorityResponse vendFromSourceCatalog(
+      String accountId, ResourceId requestedTableId, String responseLocationPrefix) {
+    Table table;
+    try {
+      table = loadVisibleTable(scopedTableId(accountId, requestedTableId, correlationId()));
+    } catch (RuntimeException e) {
```
mcusack-floe: **Severity: High — repository/infrastructure failures are silently treated as an unavailable fallback.**

Catching every `RuntimeException` from `loadVisibleTable` also catches database, corruption, timeout, and other infrastructure failures. Those failures then fall through to the generic missing-authority `INVALID_ARGUMENT`, masking an outage. Please catch only the expected not-found condition, or avoid the second table load by carrying forward the table already loaded during authorization.

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityServiceImpl.java
```diff
@@ -598,6 +656,120 @@ private CredentialScope resolveDiscoveryTableLocation(
...
+    }
+
+    LinkedHashMap<String, String> storageConfig = new LinkedHashMap<>();
+    storageConfig.put("type", "s3");
+    storageConfig.putAll(vended.get().properties());
+    VendedStorageCredential.Builder credential =
+        VendedStorageCredential.newBuilder()
+            .setPrefix(responseLocationPrefix == null ? "" : responseLocationPrefix)
+            .putAllConfig(Map.copyOf(storageConfig));
+    Instant expiresAt = vended.get().expiresAt();
```
mcusack-floe: **Severity: High — missing or malformed expiry disables credential renewal.**

For execution-bound requests, the reconciler registers its refresh provider only when access key, secret, session token, and `expires_at` are all present. If `s3.session-token-expires-at-ms` is absent or malformed, these credentials are installed statically and no re-vend is attempted before they expire. This path also bypasses the normal `SCU_CLIENT` invariant that client credentials have a known expiry. Please define and enforce the intended behavior for credentials without a usable expiry (for example, reject them for refreshable execution/client usage).

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityServiceImpl.java
```diff
@@ -598,6 +656,120 @@ private CredentialScope resolveDiscoveryTableLocation(
...
+          "source-catalog vending skipped: upstream connector %s of table %s not found",
+          upstream.getConnectorId().getId(), requestedTableId.getId());
+      return null;
+    }
+
+    String namespaceFq = String.join(".", upstream.getNamespacePathList());
+    Optional<FloecatConnector.VendedStorageCredentials> vended;
+    try (FloecatConnector source =
+        ConnectorFactory.create(resolveConnectorCredentials(connector))) {
+      vended = source.vendStorageCredentials(namespaceFq, upstream.getTableDisplayName());
```
mcusack-floe: **Severity: High — deterministic catalog authorization failures lose terminal semantics.**

Exceptions from connector creation or delegated `loadTable` escape this method and are mapped to gRPC `INTERNAL`. The reconciler does not classify that status as terminal, so a persistent catalog refusal can be retried indefinitely. Please preserve/translate authentication and authorization failures to terminal statuses such as `UNAUTHENTICATED` or `PERMISSION_DENIED`, while leaving genuinely transient catalog failures retryable.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ServerSideStorageConfigResolver.java
```diff
@@ -253,6 +263,19 @@ private ConnectorConfig resolveStorageBackedConnector(
...
+      return config;
+    }
+
     String locationPrefix = storageAuthorityLookupLocation(storageLocation, config);
```
mcusack-floe: **Severity: Medium — this appears inconsistent with the PR's compatibility claim.**

The delegation short-circuit above runs before location and authority lookup. Consequently, a connector that has both delegation configured and a working storage authority now bypasses that authority, whereas the PR says existing authority-backed configurations retain identical behavior and that source-catalog vending is only a fallback when no authority matches. Please either preserve authority precedence here or update the compatibility statement and add a test that pins the intended precedence.

mcusack-floe (commented):
Review of the Iceberg REST vended-credential path. The inline findings focus on error classification, credential renewal, delegation detection, and the stated compatibility behavior.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ServerSideStorageConfigResolver.java
asoufane-ops: You're right, and the PR description was wrong rather than the intent. The short-circuit sat at line 275, before storageAuthorityLookupLocation at 279, so adding a delegation header to a connector with a working authority silently stopped using it.

Fixed in 064be859: delegation is now a fallback taken only when the lookup reports no authority, matching StorageAuthorityServiceImpl where source-catalog vending is likewise reached only when none matches — both layers now agree on precedence. Pinned by workingStorageAuthorityWinsOverDeclaredDelegation and delegatingRestConnectorFallsBackToCatalogWhenNoAuthorityMatches.

Confirmation the review was right: my old delegatingRestConnectorSkipsStorageAuthorityVending failed immediately on the change, because it encoded exactly the behaviour you flagged.

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityServiceImpl.java
asoufane-ops: Fixed in 064be859. catalogFailureStatus maps recognisable authentication and authorization refusals to UNAUTHENTICATED / PERMISSION_DENIED; anything unrecognised stays INTERNAL so genuinely transient catalog failures remain retryable.

One caveat worth flagging: classification is by inspecting the exception chain for 401/403/unauthorized/forbidden, which is inherently approximate — Iceberg's REST client doesn't surface a typed status. If there's an existing helper for this in the codebase I'd rather use it than pattern-match on messages.

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityServiceImpl.java
asoufane-ops: Confirmed, and worse than described. The worker's is_refreshable() is exactly expires_at.is_some(), and on false it takes the with_embedded_credentials path and never re-vends — so these expire mid-read with no recovery.

Two problems, both mine: absentExpiryIsNullNotGuessed asserted the behaviour was intended, and its comment documented a "callers treat null as do-not-cache" contract that the real consumer doesn't honour. So the test was pinning the bug.

Fixed in 064be859: requireUsableExpiry rejects with FAILED_PRECONDITION at vend time, with two tests. Connector still reports absence faithfully rather than inventing a TTL; the comment now states the actual contract.

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityServiceImpl.java
asoufane-ops: Fixed in 064be859. Now catches StatusRuntimeException and rethrows anything that isn't NOT_FOUND, so repository timeouts, connection failures and corruption propagate instead of being reported as the generic missing-authority INVALID_ARGUMENT.

I took the narrow-catch option rather than carrying the table forward — threading the authorization-time Table up through resolveExecutionBoundLocation → resolveLeaseScopedLocation is a larger refactor, and the narrow catch removes the masking. Happy to do the carry-forward as a follow-up if you'd rather the redundant second load were gone too.

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ServerSideStorageConfigResolver.java
asoufane-ops: Confirmed. IcebergConnectorFactory:119 does headerHints.forEach((k,v) -> catalogProps.put("header." + k, v)), so the hint route never appears in options() and fell straight through to authority vending.

Fixed in 064be859: declaresDelegation() now derives from both options() and auth().headerHints(). Added delegationViaHeaderHintsIsDetected covering the hint path specifically.

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityServiceImpl.java
```diff
@@ -598,6 +657,214 @@ private CredentialScope resolveDiscoveryTableLocation(
...
+   * table at all.
+   *
+   * <p>Prefers an explicit {@code table_id}, then the table the reconcile lease is bound to. The
+   * lease is what makes this reachable in practice: the reconcile worker vends by location and
+   * sends no {@code table_id}, so without the leased job there is no table identity to resolve a
+   * connector from.
+   */
+  private ResourceId sourceCatalogTableId(
+      VendStorageCredentialsRequest request, String accountId, ReconcileJobStore.ReconcileJob job) {
+    if (request.hasTableId()) {
```
mrouvroy-floe, line 709: [major] **Execution-bound lease table identity can be overridden by caller**

`authorizeAndResolveLocation()` validates the requested location against the reconcile lease, but `sourceCatalogTableId()` then prefers `request.table_id` over the table bound to that same lease.

A caller with a valid lease for table A can send a `table_id` for table B. When no authority matches, the fallback asks table B's upstream connector/catalog to vend under table A's lease-scoped request. This breaks the lease invariant that the job controls both the storage scope and the table identity used for source-catalog vending, and can route the request to the wrong upstream connector/catalog entirely.

For execution-bound requests, derive the source table from the lease first and reject any explicit `table_id` that differs from the leased table identity.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ServerSideStorageConfigResolver.java
```diff
@@ -531,6 +565,33 @@ private static String normalize(String value) {
...
+  /**
+   * Whether the connector asks the catalog to vend credentials.
+   *
+   * <p>Both configuration routes count. Options carry {@code header.X-Iceberg-Access-Delegation}
+   * directly, while {@code auth().headerHints()} is turned into the same {@code header.<name>}
+   * catalog property by IcebergConnectorFactory -- so reading options alone misses a connector
+   * configured through header hints, which then falls through to storage-authority vending despite
+   * having asked for delegation.
+   */
+  private static boolean declaresDelegation(ConnectorConfig config) {
```
mrouvroy-floe: [major] **Non-vended delegation modes (e.g. `remote-signing`) incorrectly bypass missing storage authority**

`declaresDelegation()` treats any non-blank `X-Iceberg-Access-Delegation` header value as permission to swallow the storage-authority service's `INVALID_ARGUMENT` and continue with the untouched connector config.

The header has more than one valid value: `vended-credentials` returns storage credentials on `loadTable`, while `remote-signing` signs requests instead and returns no credentials. With `remote-signing` (or any other non-vended value) and no matching storage authority, the catch swallows the intended missing-authority failure and returns the untouched config — but the client's `FileIO` holds no vended credentials, so the read fails later with a far less diagnostic error. The fallback is also broader than the PR's stated intent: only `vended-credentials` / `vended_credentials` should bypass a missing storage authority.

Parse the header value and match only against the credential-vending token(s) specifically rather than accepting any non-blank value.

---

connectors/catalogs/iceberg/src/main/java/ai/floedb/floecat/connector/iceberg/impl/IcebergConnector.java
```diff
@@ -1473,4 +1474,65 @@ private static void collectNotNullConstraints(
...
+
+  /**
+   * Storage credential properties copied out of a delegated {@code loadTable}.
+   *
+   * <p>An explicit allowlist rather than a prefix match. The FileIO property map from a delegated
+   * load also contains the catalog's own {@code token}, and an "everything that looks like a
+   * credential" filter would copy it into the storage path -- handing a catalog credential to
+   * whatever reads data files.
+   */
+  private static final List<String> VENDED_STORAGE_KEYS =
```
mrouvroy-floe, line 1728: [major] **Allowlist `VENDED_STORAGE_KEYS` drops non-secret S3 routing properties (`s3.region`, `s3.endpoint`, `s3.path-style-access`)**

`VENDED_STORAGE_KEYS` copies only the three credential keys and the expiry. The security rationale only requires excluding the auth token, but the allowlist also silently drops `s3.region`, `s3.endpoint`, and `s3.path-style-access` — non-secret routing properties that a delegated `loadTable`'s `FileIO` carries for anything that isn't a default-region AWS bucket. The `assertEquals(3, props.size())` in the test pins this exclusion as intentional.

The reconcile worker then receives `type=s3` plus keys only, with no region or endpoint. The current validated Polaris-in-EKS run happened to work because the observed property map had no region and the worker's default region matched — but this feature's premise is capturing from an arbitrary delegating catalog, and any non-default-region or custom-endpoint bucket will silently misconfigure the worker's `FileIO`.

Extend the allowlist to include non-secret routing properties and update the test assertion accordingly.

---

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityServiceImpl.java
```diff
@@ -598,6 +657,214 @@ private CredentialScope resolveDiscoveryTableLocation(
...
+          .withCause(cause)
+          .asRuntimeException();
+    }
+    if (forbidden) {
+      return io.grpc.Status.PERMISSION_DENIED
+          .withDescription(detail)
+          .withCause(cause)
+          .asRuntimeException();
+    }
+    return io.grpc.Status.INTERNAL.withDescription(detail).withCause(cause).asRuntimeException();
```
mrouvroy-floe: [minor] **Substring-based error classification can incorrectly mark transient catalog failures as terminal**

`catalogFailureStatus` classifies terminality by lowercasing the whole cause chain and substring-matching `401`/`unauthorized`/`invalid_client` and `403`/`forbidden`/`access denied`.

The dangerous direction is a false *terminal*: a genuinely transient failure whose message happens to contain one of these tokens — a 503 or gateway page whose body echoes `Access Denied`, an S3-level `Access Denied` surfaced through the catalog during IAM-propagation lag, or a URL/error string containing `403` — becomes `PERMISSION_DENIED`. The reconciler treats that as terminal and stops retrying, permanently failing a job that would have recovered.

Use structured status codes or typed exception inspection rather than substring matching on the full cause chain to distinguish true auth failures from transient errors.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ServerSideStorageConfigResolver.java
```diff
@@ -257,15 +267,39 @@ private ConnectorConfig resolveStorageBackedConnector(
...
+                      locationPrefix,
+                      tableId,
+                      executionJobId,
+                      executionLeaseEpoch));
+    } catch (StatusRuntimeException e) {
+      // INVALID_ARGUMENT here is "no storage credential authority is configured for this table".
+      // With delegation declared that is not fatal: the catalog client holds vended credentials of
+      // its own, so hand back the untouched config and let it use them. Any other status is a real
+      // failure and must not be masked.
+      if (e.getStatus().getCode() == Status.Code.INVALID_ARGUMENT && declaresDelegation(config)) {
```
mcusack-floe: **Severity: High — the delegation fallback masks unrelated `INVALID_ARGUMENT` failures.**

This catch treats the canonical status as if it uniquely meant “no storage authority matches,” but `vendStorageCredentials` uses the same status for other validation failures (for example `account_id`, `execution_binding`, or `location_prefix` problems, plus generic `IllegalArgumentException` mapping). A delegating connector therefore turns a real request/configuration error into `return config`, and the eventual catalog/FileIO failure obscures the cause.

The repository error contract already carries structured Floecat details; please give the specific “no matching authority” condition a distinct application error/message key (or other stable structured discriminator), decode that here, and fall back only for that condition. Matching every `INVALID_ARGUMENT` remains unsafe.

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityServiceImpl.java
```diff
@@ -598,6 +657,214 @@ private CredentialScope resolveDiscoveryTableLocation(
...
+   * credentials statically and never re-vends, so they expire mid-read with no recovery. Client
+   * usage carries the same invariant. Failing at vend time makes that visible here instead of as an
+   * opaque 403 partway through a file group.
+   */
+  static void requireUsableExpiry(
+      FloecatConnector.VendedStorageCredentials vended, String namespaceFq, String tableName) {
+    if (vended.expiresAt() != null) {
+      return;
+    }
+    throw io.grpc.Status.FAILED_PRECONDITION
```
mcusack-floe: **Severity: High — missing-expiry rejection is still retryable indefinitely.**

Rejecting these credentials prevents the static-installation bug, but this emits a generic `FAILED_PRECONDITION`. On the initial resolution path, `JavaConnectorCaptureEngine` passes it through `ReconcileFailureClassifier`, which marks only `UNAUTHENTICATED`/`PERMISSION_DENIED` (and separately identified lease failures) terminal; executors default the unclassified error to `RETRYABLE`. Since a catalog consistently omitting/mangling expiry will not recover on retry, the job can now loop forever instead of failing mid-read.

Please surface a structured, specifically classifiable error and make that condition terminal. Avoid classifying every `FAILED_PRECONDITION` as terminal because lease-precondition failures have separate semantics.

connectors/catalogs/iceberg/src/main/java/ai/floedb/floecat/connector/iceberg/impl/IcebergConnector.java
```diff
@@ -1473,4 +1474,65 @@ private static void collectNotNullConstraints(
...
+    for (String key : VENDED_STORAGE_KEYS) {
+      String value = ioProps.get(key);
+      if (value != null && !value.isBlank()) {
+        vended.put(key, value);
+      }
+    }
+    // A catalog that does not delegate still returns a perfectly good table -- its FileIO simply
+    // carries no credentials. That is "unavailable", not a failure, so fall back to empty and let
+    // the caller use a storage authority.
+    if (!vended.containsKey("s3.access-key-id")) {
```
mcusack-floe, line 1776: **Severity: High — partial credentials can recreate the non-refreshable installation bug.**

Presence of `s3.access-key-id` alone is not a valid delegated session credential. In particular, an access key + secret + expiry but no `s3.session-token` passes this method and `requireUsableExpiry`; downstream it satisfies `isExecutionBoundStorageCredential` but fails `isRefreshableExecutionCredential`, so the reconciler embeds it statically instead of registering the refresh provider. That recreates the renewal defect the expiry fix was intended to close. A missing secret takes a different late/internal-error path.

Please validate the complete delegated tuple—nonblank access key, secret key, session token, and usable expiry—at the vending boundary, and add malformed-response tests for each required field.

mcusack-floe (approved):

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityServiceImpl.java
```diff
@@ -495,6 +543,17 @@ private CredentialScope resolveExecutionBoundLocation(
...
+      // An Iceberg table may reference data anywhere -- registering files in place (add_files)
+      // leaves them outside the table location entirely. The lease already enumerates the exact
+      // files planned for this job, so a request for one of them is in scope even though it sits
+      // under a different prefix.
+      //
+      // Scope to the requested file rather than widening the table scope: authority lookup and the
+      // minted session policy then follow the location actually being read, and a table whose data
+      // does live under its own location is unaffected.
+      if (isLeasedFilePath(job, requestedLocationPrefix)) {
+        return CredentialScope.forSingleLocation(requestedLocationPrefix);
```
mrouvroy-floe: [major] **Exact file scope mints wildcard access — leased-file escape hatch is wider than stated contract**

When a planned file is admitted via `CredentialScope.forSingleLocation(requestedLocationPrefix)`, that scope is later passed to `StorageAuthorityResolver.scopedSessionPolicy`. For every non-root scope, the policy builder grants both the exact object ARN **and** `objectArn/*`. A lease for `s3://bucket/path/part-000.parquet` therefore also mints credentials for `s3://bucket/path/part-000.parquet/...`, violating the PR's stated contract of "scoped to that file".

Impact: any holder of a per-file lease can access child keys under the file's path, broadening the blast radius beyond a single object.

Fix: introduce an exact-object scope mode for the planned-file case (e.g. a flag or a distinct `CredentialScope` variant) and suppress the child-wildcard/list-child-prefix clause in the policy builder when the scope originates from `fileGroupTask.filePaths()` membership.

---

🤖 _Generated with GPT-5.5 and Claude Opus 5._

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityServiceImpl.java
```diff
@@ -598,6 +657,240 @@ private CredentialScope resolveDiscoveryTableLocation(
...
+      return CredentialResolverSupport.apply(base, auth.getCredentials());
+    }
+    if (!connector.hasResourceId()
+        || auth.getScheme().isBlank()
+        || "none".equalsIgnoreCase(auth.getScheme())) {
+      return base;
+    }
+    return credentialResolver
+        .resolve(connector.getResourceId().getAccountId(), connector.getResourceId().getId())
+        .map(c -> CredentialResolverSupport.apply(base, c, AuthResolutionContext.empty()))
```
mrouvroy-floe: [major] **No-authority fallback uses `AuthResolutionContext.empty()`, breaking token-exchange connector auth modes**

`resolveConnectorCredentials` applies stored connector credentials with `AuthResolutionContext.empty()`. This is not equivalent to the normal connector resolution path: `ConnectorsImpl` uses `AuthResolutionContexts.fromInboundContext()`, and `CredentialResolverSupport` requires a subject token for `RFC8693_TOKEN_EXCHANGE`, `AZURE_TOKEN_EXCHANGE`, and `GCP_TOKEN_EXCHANGE` credential schemes.

Impact: a delegating source catalog configured with any of those auth modes will fail to construct the source connector before it can call `loadTable` on the new no-authority fallback path, even though the storage-authority RPC has the inbound authorization/session headers available. Client-credentials-authenticated connectors still work, but three supported auth modes become unreachable on this new code path.

Fix: propagate the inbound request context (e.g. via `AuthResolutionContexts.fromInboundContext()`) into `resolveConnectorCredentials` instead of using the empty context.

---

🤖 _Generated with GPT-5.5 and Claude Opus 5._

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityServiceImpl.java
```diff
@@ -598,6 +657,240 @@ private CredentialScope resolveDiscoveryTableLocation(
...
+          connector.getResourceId().getId(), namespaceFq, upstream.getTableDisplayName());
+      return null;
+    }
+
+    requireRefreshableCredentials(vended.get(), namespaceFq, upstream.getTableDisplayName());
+
+    LinkedHashMap<String, String> storageConfig = new LinkedHashMap<>();
+    storageConfig.put("type", "s3");
+    storageConfig.putAll(vended.get().properties());
+    VendedStorageCredential.Builder credential =
```
mrouvroy-floe: [major] **`vendFromSourceCatalog` drops `s3.region`/`s3.endpoint` on the execution-bound path**

`vendFromSourceCatalog` places routing properties (`s3.region`, `s3.endpoint`) only into `storage_credentials[0].config` and never into `client_safe_config`. On the reconcile worker's execution-bound path, `ServerSideStorageConfigResolver.mergeResolvedStorageConfig` is called with `refreshableExecutionCredentials=true` (because both job id and lease epoch are present). In that branch it applies `response.getClientSafeConfigMap()`, then merges only the refreshable-provider keys or the credential triple from `ResolvedStorageCredentials.asS3Properties()`. The `getStorageCredentials(0).getConfigMap()` merge at the end of the method is never reached on this branch.

Impact: execution-bound file-group vends silently lose the region and endpoint overrides required to route requests to the correct S3-compatible endpoint, causing runtime failures or silent misrouting on the worker side.

Fix: also populate `client_safe_config` with the routing properties in `vendFromSourceCatalog`, or ensure the execution-bound merge path includes `getStorageCredentials(0).getConfigMap()`.

---

🤖 _Generated with GPT-5.5 and Claude Opus 5._

connectors/catalogs/iceberg/src/main/java/ai/floedb/floecat/connector/iceberg/impl/IcebergConnector.java
```diff
@@ -1473,4 +1474,76 @@ private static void collectNotNullConstraints(
...
```
mrouvroy-floe, line 1776: [major] **`vendStorageCredentials` cannot reliably distinguish catalog-vended credentials from connector-configured ones**

`table.io().properties()` contains more than what the catalog vended. `IcebergConnectorFactory.applyStorageProperties(catalogProps, storageProps)` writes the connector's own configured `s3.*` values into the REST catalog properties, and Iceberg merges them into every table's `FileIO` property map. Presence of `s3.access-key-id` therefore does not reliably indicate that delegation occurred.

Impact: this is not a corner case. `resolveConnectorCredentials` → `CredentialResolverSupport.applyAwsCredentials` writes `s3.access-key-id`, `s3.secret-access-key`, and `s3.session-token` for `aws`, `aws-assume-role`, and `aws-web-identity` schemes. A Glue or S3-Tables connector authenticated that way passes the `containsKey("s3.access-key-id")` gate with three of the four required fields present, causing incorrect branching in `vendStorageCredentials`.

Fix: use a dedicated sentinel (e.g. a well-known property key set only by the catalog vend path, or a separate out-of-band flag) to distinguish catalog-delegated credentials from connector-configured ones, rather than relying on the presence of `s3.access-key-id`.

---

🤖 _Generated with GPT-5.5 and Claude Opus 5._

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityServiceImpl.java
```diff
@@ -598,6 +657,240 @@ private CredentialScope resolveDiscoveryTableLocation(
...
+          .setId(leasedTableId)
+          .build();
+    }
+
+    // No table on the lease. Either the request is not execution-bound at all -- authorized by
+    // table.read on the requested id -- or it is a discovery planner job, where
+    // resolveDiscoveryTableLocation has already checked the requested table against the leased
+    // source and connector before we get here.
+    if (request.hasTableId()) {
+      return request.getTableId();
```
mrouvroy-floe, line 710: [minor] **Caller-supplied `table_id` is unvalidated for `PLAN_VIEW` bootstrap leases — confused-deputy risk**

The comment above this line claims the non-leased case is safe because `resolveDiscoveryTableLocation` has already validated the requested table. This does not hold for `PLAN_VIEW`: `allowsConnectorBootstrapScope` admits `PLAN_VIEW` jobs, but `resolveDiscoveryTableLocation` returns `null` for any job whose `jobKind != PLAN_TABLE`, and `leasedTableId` finds nothing on a `PLAN_VIEW` job. The scope therefore comes from `resolvePlannerBootstrapLocation`, which validates only the *location* against the connector root and never inspects `request.table_id`.

Impact: a holder of a `PLAN_VIEW` lease can name any table in the account and have that table's upstream connector vend credentials for it — the same confused-deputy shape the leased-table path was designed to prevent.

Fix: validate `request.table_id` against the `PLAN_VIEW` job's own authorized table set inside `resolvePlannerBootstrapLocation`, or explicitly reject non-`PLAN_TABLE` jobs before reaching the unvalidated scope-resolution path.

---

🤖 _Generated with GPT-5.5 and Claude Opus 5._

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityServiceImpl.java
```diff
@@ -598,6 +658,366 @@ private CredentialScope resolveDiscoveryTableLocation(
...
+   * <p>Reached only when no storage authority covers the location. Returning null means "the
+   * catalog does not delegate" and lets the caller fail with the usual missing-authority error; it
+   * does not swallow a catalog that delegates but refused, which surfaces as a thrown exception.
+   *
+   * <p>NOTE: builds a connector per call, which for an Iceberg REST catalog means an OAuth exchange
+   * and a fresh HTTP client. Vending happens per file group, so this wants a cache keyed on the
+   * table and bounded by the credential expiry before it carries production traffic. Left out here
+   * because the invalidation policy is a design decision, not an implementation detail.
+   */
+  private ResolveStorageAuthorityResponse vendFromSourceCatalog(
```
mrouvroy-floe, line 741: [major] **Source-catalog vending logic should be extracted into a focused collaborator**

`StorageAuthorityServiceImpl` now owns the entire source-catalog vending workflow: lease-derived table selection, upstream lookup, delegation detection, connector credential resolution, connector construction, `loadTable` vending, credential validation, routing-property shaping, and catalog failure classification.

This PR adds roughly 459 lines to a service class already responsible for the storage-authority API. The new path is therefore embedded in the least-focused place for security-sensitive credential plumbing. Invariants are spread across private helpers and mirrored comments, making future fixes likely to span API handling, lease authorization, connector auth, and error classification simultaneously.

Move the catalog-vending path behind a focused collaborator that owns this workflow end-to-end, leaving `StorageAuthorityServiceImpl` as a thin delegator.

---

reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/ServerSideStorageConfigResolver.java
```diff
@@ -531,6 +569,55 @@ private static String normalize(String value) {
...
+  /**
+   * Whether the connector asks the catalog to vend credentials.
+   *
+   * <p>Both configuration routes count. Options carry {@code header.X-Iceberg-Access-Delegation}
+   * directly, while {@code auth().headerHints()} is turned into the same {@code header.<name>}
+   * catalog property by IcebergConnectorFactory -- so reading options alone misses a connector
+   * configured through header hints, which then falls through to storage-authority vending despite
+   * having asked for delegation.
+   */
+  private static boolean declaresDelegation(ConnectorConfig config) {
```
mrouvroy-floe: [minor] **Delegation-header parsing is duplicated across two modules and must stay in lockstep**

`declaresDelegation` / `declaresVendedCredentials` (options + `auth().headerHints()`, comma-split, case-insensitive, `_`/`-` normalized, matching only `vended-credentials`) is now implemented independently in both `ServerSideStorageConfigResolver` (reconciler) and `StorageAuthorityServiceImpl.connectorDeclaresVendedDelegation` (service).

These two gates must agree: the service decides whether to *attempt* source-catalog vending, and the reconciler decides whether to *swallow* the resulting missing-authority error and fall back to the untouched config. If they drift — e.g. one gains support for a new delegation token or a whitespace/casing edge — a connector could attempt vending on one side but not have its no-authority failure absorbed on the other, surfacing as an opaque late `FileIO` error.

Extract the parsing logic into a single shared utility and have both sides call it.

---

core/storage-spi/src/main/java/ai/floedb/floecat/storage/errors/SourceCatalogVendingGrpcStatus.java
```diff
@@ -0,0 +1,142 @@
...
+ * {@code INVALID_ARGUMENT} is also what {@code vendStorageCredentials} returns for account_id,
+ * execution_binding and location_prefix validation failures, so a delegating connector matching on
+ * the code alone turned real configuration errors into a silent fallback. {@code
+ * FAILED_PRECONDITION} is shared with lease-precondition failures, which have their own retry
+ * semantics, so it cannot simply be classified terminal.
+ *
+ * <p>Same shape as {@link ReconcileLeaseGrpcStatus}: an ErrorInfo detail carrying a stable domain
+ * and reason, plus a matching predicate.
+ */
+public final class SourceCatalogVendingGrpcStatus {
```
mrouvroy-floe, line 44: [minor] **Storage error protocol class lives in reconciler package but is owned and thrown by the service layer**

`SourceCatalogVendingGrpcStatus` declares a storage-domain error protocol (`ERROR_DOMAIN = "ai.floedb.floecat.storage"`) yet lives under `reconciler.impl` and is thrown directly from `service/storage/impl`. This makes the storage service depend on a reconciler implementation package for conditions like `NO_MATCHING_STORAGE_AUTHORITY` and `VENDED_CREDENTIALS_NOT_REFRESHABLE`.

This is exactly the kind of cross-layer coupling that becomes sticky: the service, reconciler, and client-side resolver all need the same structured reasons, but none of them cleanly owns the protocol.

Move this class to a shared API/error package — or a storage-owned package that both service and reconciler can depend on — rather than letting it accumulate dependents in its current location.

---

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityServiceImpl.java
```diff
@@ -598,6 +658,366 @@ private CredentialScope resolveDiscoveryTableLocation(
...
+            "source catalog %s refused credentials for %s.%s: %s",
+            connector.getResourceId().getId(), namespaceFq, tableName, cause);
+
+    // Typed exceptions only. Substring-matching the cause chain for 401/403/"access denied" gets
+    // the risk backwards: a transient failure whose text merely contains one of those tokens -- a
+    // gateway page echoing "Access Denied", an S3 denial during IAM propagation lag, a URL with 403
+    // in it -- would be classified terminal and stop the reconciler retrying a job that would have
+    // recovered. Iceberg's REST client raises NotAuthorizedException for 401 and ForbiddenException
+    // for 403, so classification uses those and nothing else.
+    for (Throwable c = cause; c != null; c = c.getCause()) {
```
mrouvroy-floe: [nit] **`catalogFailureStatus` cause-chain walk guards only self-cycles, not two-node cycles**

The classification loop walks the cause chain with only a self-reference guard (`if (c.getCause() == c) break;`). A two-node cycle (`a.cause == b`, `b.cause == a`) is never caught by that check, so the loop would spin forever.

In practice Iceberg exceptions do not form cycles, so this is latent rather than live — but it is an inconsistency worth closing. The sibling `SourceCatalogVendingGrpcStatus.hasReason` introduced in this same PR already walks a cause chain safely with a `HashSet<Throwable> seen` guard.

Mirror that approach here: track visited throwables in a set and break when one repeats, rather than only guarding the self-reference case.

---

service/src/main/java/ai/floedb/floecat/service/storage/impl/StorageAuthorityResolver.java
```diff
@@ -396,7 +448,10 @@ String objectStatementJson() {
         }
         String objectArn = jsonEscape("arn:aws:s3:::%s/%s".formatted(bucket, scope.keyPrefix()));
```
mrouvroy-floe, line 476: [major] **Exact object policy widens on IAM wildcard metacharacters in object keys**

`exactObjectScope` only strips the trailing `/*` from the ARN, but leaves the raw object key embedded in the IAM resource ARN. S3 object keys can legally contain `*` and `?`, which are IAM wildcard metacharacters, so a planned key like `s3://warehouse/orders/data/part-*.parquet` produces a session policy resource that matches sibling objects rather than only the literal named file.

This breaks the guarantee that a planned-file exception is scoped to exactly one object. The current test only covers a safe key (`part-000.parquet`), so no existing assertion pins the widening behavior.

Reject keys containing IAM wildcard metacharacters before minting the session policy, or introduce an escaping/authorization strategy that preserves literal `*` and `?` as non-wildcard characters in the ARN.

---

service/src/main/java/ai/floedb/floecat/service/storage/impl/SourceCatalogCredentialVendor.java
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    requireRefreshableCredentials(vended.get(), namespaceFq, upstream.getTableDisplayName());
+
+    LinkedHashMap<String, String> storageConfig = new LinkedHashMap<>();
+    storageConfig.put("type", "s3");
+    storageConfig.putAll(vended.get().properties());
+    VendedStorageCredential.Builder credential =
+        VendedStorageCredential.newBuilder()
+            .setPrefix(responseLocationPrefix == null ? "" : responseLocationPrefix)
```
mrouvroy-floe, line 191: [major] **Catalog-vended fallback loses exact-object scope, returning over-broad table credentials**

When no storage authority matches and the fallback path in `vendFromSourceCatalog(...)` runs, `CredentialScope.forSingleObject(...)` is constructed but its `exactObjectScope` never reaches an STS session policy. The vendor stamps only `credentialScope.responseLocationPrefix()` into `VendedStorageCredential.prefix` and returns the raw table-level credentials from `loadTable` directly.

This is materially broader than the authority-backed path, where `exactObjectScope` is embedded in the STS session policy to suppress child wildcards. The reconciler and gateway merge paths consume `storage_credentials[0].config` as FileIO properties without enforcing the response `prefix`, so a planned external file can silently receive wider table-level access instead of the narrower single-object scope.

Ensure that the fallback path either (a) synthesizes an STS session policy using `exactObjectScope` before returning credentials, or (b) propagates the scope into all downstream credential consumers so the single-object restriction is actually enforced.

---

service/src/main/java/ai/floedb/floecat/service/storage/impl/ServerSideFileIoPropertiesResolver.java
```diff
@@ -58,6 +59,20 @@ public Map<String, String> resolve(Table table, String location) {
...
     StoreOperationSummary.put("storage_authority_source", "load");
     var authority = StorageAuthorityResolver.resolveBest(authorities, locationPrefix).orElse(null);
+    if (authority == null) {
+      // No authority covers this location. Before failing, ask the source catalog -- exactly as the
+      // vend RPC does, and for the same reason: a table captured through a delegating Iceberg REST
+      // catalog has no authority to configure, because the catalog vends its own credentials. This
+      // path is what lets such a table be *read back*. Without it capture succeeds through
+      // delegation and every scan of the result dies on NO_MATCHING_STORAGE_AUTHORITY.
+      ResolveStorageAuthorityResponse vended =
+          sourceCatalogVendor.vendForTable(table, locationPrefix);
```
mrouvroy-floe: [minor] **Query read-back path performs unbounded per-scan catalog round-trips with no cache or follow-up note**

The newly-added query read-back branch calls `sourceCatalogVendor.vendForTable(...)` on every invocation of `buildTableInfo`. Because `buildTableInfo` runs once per query scan session (`ScanBundleService:123`), every scan of a delegated table pays a full `ConnectorFactory.create` + OAuth exchange + `loadTable` round-trip.

The PR description and the `vendFromSourceCatalog` comment in `StorageAuthorityServiceImpl` already acknowledge this cost on the reconcile RPC path and call it a required pre-production follow-up ("cache keyed on the table and bounded by the credential expiry"). The query path introduced in this commit shares `SourceCatalogCredentialVendor` and has the same unbounded per-call cost, but carries no equivalent note, making it easy for the follow-up work to miss this call site.

Add a `// TODO` or tracking comment on this call site mirroring the one on the reconcile path, and include it in the scope of the planned caching work.

---

service/src/main/java/ai/floedb/floecat/service/storage/impl/SourceCatalogCredentialVendor.java
```diff
@@ -0,0 +1,378 @@
...
+    String vendedAccessKey = vended.get().properties().get("s3.access-key-id");
+    if (configuredAccessKey != null && configuredAccessKey.equals(vendedAccessKey)) {
+      LOG.infof(
+          "source-catalog vending skipped: connector %s returned its own configured credentials"
+              + " for %s.%s (catalog did not delegate)",
+          connector.getResourceId().getId(), namespaceFq, upstream.getTableDisplayName());
+      return null;
+    }
+
+    requireRefreshableCredentials(vended.get(), namespaceFq, upstream.getTableDisplayName());
```
mrouvroy-floe: [minor] **Query read-back path inherits the reconcile-worker refreshable-credential requirement**

`vendForTable` calls `requireRefreshableCredentials` unconditionally, and that method's entire justification is reconcile-worker-specific: it mirrors the worker's `is_refreshable() == expires_at.is_some()` and the static-embedding defect. But `vendForTable` is also the query read-back path (`ServerSideFileIoPropertiesResolver.resolve`), where credentials are handed straight to the scan engine's `FileIO` for immediate reads and no refresh provider is ever registered.

The impact is a latent divergence: a table captured while a storage authority covered it never went through source-catalog vending, so its catalog was never required to vend a session token + expiry. If that authority later stops matching and read-back falls into delegation, the query now rejects credentials with a terminal-classified `FAILED_PRECONDITION` even for creds that would read fine. In practice a `vended-credentials` catalog almost always returns STS session tokens with an expiry, so the trigger is narrow.

Consider scoping the refreshable requirement to the execution-bound/reconcile caller (e.g. a flag on `vendForTable`) rather than enforcing it on the read-only query path, or add a comment at line 161 recording that the query path deliberately shares the same invariant.

---

mrouvroy-floe (approved):

PR #446, opened 2026-08-12
docs(integrations): finalize catalog integration design (stack 1/10)

mcusack-floe:
## Summary

Finalizes `catalog-integration-design.md` as the design and delivery contract for the catalog integration stack. It incorporates the implementation decisions reflected by the completed stack, links the eventual Catalog Access SPI, and fixes the documentation-area link.

This is PR 1 of 10 and intentionally contains documentation only.

## Type of Change

- [ ] fix (bug fix)

## Testing

- `make test-site`

- [x] `make fmt`
- [ ] `make verify`

- [x] No breaking changes

## Checklist

## Additional Notes

Stack position: **1/10**. Review this PR as the design baseline for the seven implementation PRs above it.

### DIFF
--- a/docs/catalog-integration-design.md
+++ b/docs/catalog-integration-design.md
@@ -5,7 +5,7 @@ implemented in the current PR stack.
 
 This document records both the target architecture and the boundary of each stacked change. It is
 aligned with the Catalog Integration and Catalog Overlay SQL specification. It distinguishes code
-already present below this change, code in the following PR, and work that remains deferred.
+implemented in the following stacked changes from work that remains deferred.
 
 ## Stack implementation map
 
@@ -13,7 +13,7 @@ already present below this change, code in the following PR, and work that remai
 | --- | --- | --- |
 | #424, Catalog Integration and Overlay APIs | CRUD resources, typed authentication, write-only credential storage and rotation, shared SQL catalog-name reservation, idempotency, optimistic concurrency, dependencies, cascade deletion, and atomic persistence primitives | Upstream connectivity, provider discovery, validation RPCs, capture, reconciliation, and query visibility |
 | #425, Shell CLI | Integration and Overlay CRUD commands, typed authentication input, write-only credential input, authentication rotation, pagination, name resolution, namespace filters, cascade, and etag preconditions | Connectivity validation, discovery, capture, reconciliation, and query visibility |
-| #439, this document | Architecture decisions and delivery boundaries | Production behavior |
+| #446, this document | Architecture decisions and delivery boundaries | Production behavior |
 | #440, catalog-access SPI | Connector-independent catalog client SPI plus an Iceberg REST provider with validation, discovery, table loading, OAuth2, SigV4, and renewable AWS credential support | Wiring Catalog Integration resources to the SPI, Integration validation/listing RPCs, persistence, scheduling, and capture |
 
 ## Goals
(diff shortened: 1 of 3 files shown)


docs/catalog-integration-design.md
```diff
@@ -0,0 +1,486 @@
...
+Connectors will remain a separate operational path until they are retired through an explicit
+later migration, not through runtime fallback.
+
+## Resource and ownership model
+
+```text
+CatalogIntegration "prod-glue"     connection and authentication
+ ├── CatalogOverlay "crm_data"     upstream selection -> owns Catalog "crm_data"
+ └── CatalogOverlay "finance"      a different selection -> owns Catalog "finance"
+
```
kwesterfeld-floe, line 70: As discussed, I don't think a 1:1 rigid ownership of catalog<->catalog overlay makes sense here.  It would be better to allow ownership at the namespace level.  Doing a cross namespace (schema) in PG is done all the time.  Its a Yellowbrick feature that we enabled cross database (ie. catalog) queries to exist at all. While we can do them with Floe, its a bit artificially limiting to make this ownership choice here when we already do not do that.

mrouvroy-floe:
I like the idea of splitting the Connector model.
But I'm not sure about a few things from this doc:
- `CatalogOverlay`
  This seems to be doing a lot of things: source selection, the destination catalog, lifecycle of materialized objects, and eventually policy/state like refresh and retention. Wouldn't it be better to divide ownership of these into a few different components?
  As per this doc, one overlay owns exactly one top-level catalog, and all upstream namespaces appear under that catalog without remapping. That's fine for now, but what happens if we later need namespace-level mounts, multiple integrations under one catalog, or local/imported content mixed in the same catalog? The overlay would likely have to be rewritten vs one specific and smaller bit.
  I also don't particularly like the name itself: as described, the `CatalogOverlay` isn't a virtual overlay and is more like a materialized catalog mapping or mount. Is it really worth the renaming churn in unrelated parts of the stack vs choosing a more descriptive name for this?
- Shared capture work & duplicated objects
  If two overlays point at the same upstream table, Floecat creates two separate Floecat tables. [...] The integration scheduler can share discovery/capture work across overlays.
  
  I think the doc should be clearer about what "shared capture work" means. Sharing external I/O across overlays seems fine, but persisted resulting objects like snapshots/statistics/validation should be duplicated per overlay table, right? Otherwise we would still need a shared upstream table identity of some kind that both could point to.
- Updating overlays/integrations
  Many fields of the `CatalogIntegration` & `CatalogOverlay` are said to be immutable, which is safe. But that will make it much more complicated to operate day to day on those objects IMO.
  If the upstream catalog moves to a new endpoint for instance, instead of just updating the integration, it sounds like you have to do something like:
  - create a new integration
  - replace the overlay
  - retire old materialized metadata
  - create new materialized metadata
  - delete old integration
  That feels pretty heavy for what could just be an endpoint change. Maybe that's intentional, but then the doc should probably spell out what replacement actually means and what state is preserved vs destroyed.
- Reconciliation under failure
  I think the doc should be more precise here: what happens when we fail in the middle of reconciliation? Do we publish nothing? Only what succeeded? From there, what is considered "stale"? Only tables we replaced, or any table that isn't there anymore? If the latter, then partial publishing could destroy valid data.
- View behaviour
  The view rule is correct, but shouldn't that be a planner/runtime consideration rather than a Floecat connector/integration concern? I don't think connectors should have to bother with this at all.

mrouvroy-floe (approved):

kwesterfeld-floe (approved):

docs/catalog-integration-design.md
mcusack-floe: I've removed the 1:1 ownership. Catalog Overlays don't own a catalog. Two overlays can reconcile into the same Floecat catalog now.

PR #449, opened 2026-08-12
feat(repo): add catalog integration durability primitives (stack 3/10)

mcusack-floe:
## Summary

Adds the repository durability primitives needed by catalog integrations and overlays: atomic mutation helpers, repository-backed idempotency operations, and the generic resource-repository behavior used by later API slices.

This PR deliberately separates foundational repository work from the catalog-specific resources that consume it.

## Type of Change

- [ ] fix (bug fix)

## Testing

Focused `IdempotencyGuard` and repository tests passed.

- [x] `make fmt`
- [ ] `make verify`

- [x] No breaking changes

## Checklist

## Additional Notes

Stack position: **3/10**. Base: `mc-catalog-graph-view-rename`. Review this as storage/repository infrastructure; no catalog integration API is introduced here.

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/common/IdempotencyGuard.java
+++ b/service/src/main/java/ai/floedb/floecat/service/common/IdempotencyGuard.java
@@ -25,6 +25,7 @@ import ai.floedb.floecat.service.repo.IdempotencyRepository;
 import ai.floedb.floecat.service.repo.model.Keys;
 import ai.floedb.floecat.service.repo.util.BaseResourceRepository;
 import ai.floedb.floecat.storage.errors.StorageAbortRetryableException;
+import ai.floedb.floecat.storage.spi.PointerStore;
 import com.google.protobuf.Duration;
 import com.google.protobuf.Timestamp;
 import com.google.protobuf.util.Timestamps;
@@ -32,6 +33,8 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.Base64;
 import java.util.Map;
+import java.util.Optional;
+import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import org.jboss.logging.Logger;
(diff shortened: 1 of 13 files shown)


service/src/main/java/ai/floedb/floecat/service/common/IdempotencyGuard.java
```diff
@@ -298,16 +472,36 @@ public static <T> Result<T> runOnce(
...
-        } catch (Throwable deleteError) {
-          LOG.warnf(deleteError, "idempotency.delete_failed key=%s corr=%s", key, corrId.get());
-        }
+        deletePendingIfOwned(store, key, opName, requestHash, now, expiresAt, corrId);
       }
       throw t;
     }
   }
 
+  private static void deletePendingIfOwned(
```
mrouvroy-floe, line 481: This checks that the record is still the pending record we own, but the actual delete is a separate non-conditional `store.delete(key)`.

A concurrent retry could finalize the same idempotency key to `SUCCEEDED` after this check, and this cleanup path could then delete the success receipt. I think the repository needs a conditional delete for this exact pending pointer/blob/version instead of a read followed by an unconditional delete.

service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java
```diff
@@ -692,6 +1260,59 @@ public boolean deleteWithPrecondition(K key, long expectedCanonicalVersion) {
...
+
+  /**
+   * Returns the pointer operations that would delete this resource at its current versions, without
+   * committing anything, so a caller can fold them into another repository's atomic mutation. The
+   * counterpart to {@link #prepareCreateOps}.
+   *
+   * <p>Returns empty when the resource is already absent, which lets a cascading caller treat a
+   * partially completed earlier attempt as work already done rather than as an error.
+   */
+  public List<PointerStore.CasOp> prepareDeleteOps(K key) {
```
mrouvroy-floe, line 1286: This helper returns delete operations without calling `guardSystemObject(key)`, while the direct delete paths do enforce that backstop.

Since this is a generic atomic-cascade primitive, it should preserve the same system-object immutability invariant by default. Please call `guardSystemObject(key)` before reading/assembling delete ops.

service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java
```diff
@@ -610,6 +1010,128 @@ private boolean classifyUpdateConflict(
...
+            throw new IllegalArgumentException("duplicate replacement canonical pointer");
+          }
+          Pointer replacementReserved =
+              reserve(
+                  replacementCanonical, replacementBlobUri, replacementValue, replacementBlobBytes);
+          ops.add(new PointerStore.CasUpsert(replacementCanonical, 0L, replacementReserved));
+
+          Set<String> allSecondaries = new HashSet<>(currentSecondaryKeys);
+          allSecondaries.addAll(replacementSecondaryKeys);
+          for (String pointerKey : allSecondaries) {
```
mrouvroy-floe, line 1076: This treats any secondary pointer collision with a canonical pointer as an illegal state, but some existing schemas intentionally expose the canonical key as a secondary too.

`Schemas.SNAPSHOT` has this shape with `byId`, and create/update already de-dupe it. This replacement path should mirror that behavior: skip secondary keys already covered by the current/replacement canonical operation, and only fail on real extra collisions.

service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java
```diff
@@ -458,112 +780,190 @@ private void cleanupCreateIfAbsentBlobOnCasMiss(
...
-            return true;
+            return Optional.of(
+                committedMeta(
+                    Optional.of(committedCanonical),
+                    canonicalPointer,
+                    blobUri,
+                    Timestamps.fromMillis(clock.millis())));
           }
-          return classifyUpdateConflict(canonicalPointer, expectedCanonicalVersion, blobUri, toAdd);
+          if (!pointerConditionsStillMatch(requiredPointerVersions, requiredAbsentPointers))
```
mrouvroy-floe, line 962: If a companion `CasUpsert(..., expectedVersion = 0, ...)` loses because that companion name already exists, the conflict is terminal.

Right now the failed batch is classified using only the main resource conditions/markers/secondary names, so this can fall through as `AbortRetryableException` and make callers spin. Please add the same companion-conflict classification that `createWithMeta` already has.

mrouvroy-floe (commented):
I’d also add a happy-path test for the new atomic `SuccessCommitter` flow: prepare the success receipt, fold it into the resource pointer batch, commit, then verify the guard reads back the `SUCCEEDED` receipt. The current tests cover failure/adoption/replay paths, but not the main protocol this PR introduces.

mrouvroy-floe (approved):

service/src/main/java/ai/floedb/floecat/service/common/IdempotencyGuard.java
mcusack-floe: I see, I fixed this. code now validates the pending record and conditionally deletes the exact observed pointer version.

service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java
mcusack-floe: prepareDeleteOps now calles guardSystemObject() before reading the resource or assembling delete ops.

service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java
mcusack-floe: Replacement path now skips a current-only secondary alias of the canonical.

service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java
mcusack-floe: Update path now calls classifyCompanionConflict() before classifying the resource conflict and after confirming pointer and marker conditions still match. We should get a NameConflictException instead of a retryable.

PR #450, opened 2026-08-12
feat(account): fence and resume account deletion (stack 4/10)

mcusack-floe:
## Summary

Makes account deletion fenced and resumable, and extends repository/storage deletion support so catalog-owned resources can be cleaned up safely. This establishes the lifecycle behavior required before catalog integration and overlay resources are added.

## Type of Change

- [ ] fix (bug fix)

## Testing

Focused account-deletion and repository/storage tests passed.

- [x] `make fmt`
- [ ] `make verify`

- [x] No breaking changes

## Checklist

## Additional Notes

Stack position: **4/10**. Base: `mc-catalog-durability-primitives`. The catalog-specific deletion hooks are added by the resource PRs above this one.

### DIFF
--- a/core/storage-spi/src/main/java/ai/floedb/floecat/storage/spi/PointerStore.java
+++ b/core/storage-spi/src/main/java/ai/floedb/floecat/storage/spi/PointerStore.java
@@ -123,10 +123,8 @@ public interface PointerStore {
   List<Pointer> listPointersByPrefix(
       String prefix, int limit, String pageToken, StringBuilder nextTokenOut);
 
-  default List<Pointer> listPointersByPrefixConsistent(
-      String prefix, int limit, String pageToken, StringBuilder nextTokenOut) {
-    return listPointersByPrefix(prefix, limit, pageToken, nextTokenOut);
-  }
+  List<Pointer> listPointersByPrefixConsistent(
+      String prefix, int limit, String pageToken, StringBuilder nextTokenOut);
 
   /**
    * Returns a page token that resumes a {@link #listPointersByPrefix} scan immediately after the
@@ -145,9 +143,7 @@ public interface PointerStore {
 
   int countByPrefix(String prefix);
 
-  default int countByPrefixConsistent(String prefix) {
-    return countByPrefix(prefix);
-  }
+  int countByPrefixConsistent(String prefix);
 
   boolean isEmpty();
 
--- a/core/storage-spi/src/main/java/ai/floedb/floecat/storage/spi/PointerStore.java
+++ b/core/storage-spi/src/main/java/ai/floedb/floecat/storage/spi/PointerStore.java
@@ -123,6 +123,11 @@ public interface PointerStore {
   List<Pointer> listPointersByPrefix(
       String prefix, int limit, String pageToken, StringBuilder nextTokenOut);
 
+  default List<Pointer> listPointersByPrefixConsistent(
+      String prefix, int limit, String pageToken, StringBuilder nextTokenOut) {
+    return listPointersByPrefix(prefix, limit, pageToken, nextTokenOut);
+  }
+
   /**
    * Returns a page token that resumes a {@link #listPointersByPrefix} scan immediately after the
    * given pointer key, as if a previous page had ended exactly at that key. This lets callers that
@@ -139,6 +144,10 @@ public interface PointerStore {
   int deleteByPrefix(String prefix);
 
   int countByPrefix(String prefix);
+
+  default int countByPrefixConsistent(String prefix) {
+    return countByPrefix(prefix);
+  }
 
   boolean isEmpty();
 
--- a/core/storage-spi/src/main/java/ai/floedb/floecat/storage/spi/PointerStore.java
+++ b/core/storage-spi/src/main/java/ai/floedb/floecat/storage/spi/PointerStore.java
@@ -17,6 +17,7 @@
 package ai.floedb.floecat.storage.spi;
 
 import ai.floedb.floecat.common.rpc.Pointer;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -140,6 +141,33 @@ public interface PointerStore {
   }
 
   int deleteByPrefix(String prefix);
+
+  /**
+   * Deletes every pointer under {@code prefix} except the exact {@code excludedKey}.
+   *
+   * <p>The exclusion is continuous: implementations must never delete and recreate the excluded
+   * key. Account teardown uses this to sweep an account partition while preserving its durable
+   * deletion fence.
+   */
+  default int deleteByPrefixExcluding(String prefix, String excludedKey) {
+    int deleted = 0;
+    String token = "";
+    var seenTokens = new HashSet<String>();
+    do {
+      var next = new StringBuilder();
+      List<Pointer> page = listPointersByPrefixConsistent(prefix, 500, token, next);
+      for (Pointer pointer : page) {
+        if (!pointer.getKey().equals(excludedKey) && delete(pointer.getKey())) {
+          deleted++;
+        }
+      }
+      token = next.toString();
+      if (!token.isBlank() && !seenTokens.add(token)) {
+        throw new IllegalStateException("stagnant pointer scan token: " + token);
+      }
+    } while (!token.isBlank());
+    return deleted;
+  }
 
   int countByPrefix(String prefix);
 
--- a/service/src/main/java/ai/floedb/floecat/service/account/impl/AccountServiceImpl.java
+++ b/service/src/main/java/ai/floedb/floecat/service/account/impl/AccountServiceImpl.java
@@ -31,7 +31,6 @@ import ai.floedb.floecat.account.rpc.ListAccountsRequest;
 import ai.floedb.floecat.account.rpc.ListAccountsResponse;
 import ai.floedb.floecat.account.rpc.UpdateAccountRequest;
 import ai.floedb.floecat.account.rpc.UpdateAccountResponse;
-import ai.floedb.floecat.catalog.rpc.Catalog;
 import ai.floedb.floecat.common.rpc.MutationMeta;
 import ai.floedb.floecat.common.rpc.Pointer;
 import ai.floedb.floecat.common.rpc.ResourceId;
@@ -45,19 +44,13 @@ import ai.floedb.floecat.service.common.MutationOps;
 import ai.floedb.floecat.service.credentials.DefaultCredentialResolver;
 import ai.floedb.floecat.service.error.impl.GrpcErrors;
 import ai.floedb.floecat.service.metagraph.overlay.user.UserGraph;
+import ai.floedb.floecat.service.reconciler.jobs.DurableReconcileJobStore;
 import ai.floedb.floecat.service.repo.IdempotencyRepository;
 import ai.floedb.floecat.service.repo.impl.AccountRepository;
-import ai.floedb.floecat.service.repo.impl.CatalogRepository;
-import ai.floedb.floecat.service.repo.impl.ConnectorRepository;
-import ai.floedb.floecat.service.repo.impl.NamespaceRepository;
-import ai.floedb.floecat.service.repo.impl.StorageAuthorityRepository;
-import ai.floedb.floecat.service.repo.impl.TableRepository;
 import ai.floedb.floecat.service.repo.impl.TableRootRepository;
-import ai.floedb.floecat.service.repo.impl.ViewRepository;
 import ai.floedb.floecat.service.repo.model.Keys;
 import ai.floedb.floecat.service.repo.model.PointerReferences;
 import ai.floedb.floecat.service.repo.util.BaseResourceRepository;
-import ai.floedb.floecat.service.repo.util.MarkerStore;
 import ai.floedb.floecat.service.security.impl.Authorizer;
 import ai.floedb.floecat.service.security.impl.PrincipalProvider;
 import ai.floedb.floecat.service.storage.impl.StorageAuthorityResolver;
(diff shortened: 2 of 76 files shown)


service/src/main/java/ai/floedb/floecat/service/account/impl/AccountServiceImpl.java
```diff
@@ -443,6 +473,63 @@ public Uni<DeleteAccountResponse> deleteAccount(DeleteAccountRequest request) {
...
+
+  private void clearAccountDeletionFence(String accountId, MutationMeta expectedMeta) {
+    String key = Keys.accountDeletionMarker(accountId);
+    pointerStore
+        .get(key)
+        .filter(marker -> decodeDeletionMeta(marker).equals(expectedMeta))
+        .ifPresent(marker -> pointerStore.compareAndDelete(key, marker.getVersion()));
+  }
+
   private void cleanupAccountResources(ResourceId accountId) {
```
mrouvroy-floe, line 538: cleanupAccountResources() only deletes connectors and catalogs. Existing StorageAuthority resources are account-scoped and can carry credentials/secrets, but they are never listed or deleted during account deletion.

After the account row is removed and the deletion fence is left behind, those storage-authority pointers and their secret-manager entries become durable orphans. Also,  credential-vending paths can still list authorities by account_id, making this a credentials leak.

service/src/main/java/ai/floedb/floecat/service/account/impl/AccountServiceImpl.java
mrouvroy-floe, line 710: purgeSnapshotsAndStats() deletes only Keys.snapshotRootPrefix(...) (covering /accounts/{a}/tables/{t}/snapshots/) plus the table-root pointer. 
What about  table-scoped constraint pointers such as Keys.snapshotConstraintsPointerPrefix(...) under /accounts/{a}/tables/{t}/constraints/by-snapshot/ ?

If a table has constraints, account deletion removes the table record but leaves those constraint pointers/blobs behind with no remaining cleanup path to reach them. This breaks the resumable-delete guarantee and leaves durable account-owned state after the account is gone.

service/src/main/java/ai/floedb/floecat/service/account/impl/AccountServiceImpl.java
```diff
@@ -405,36 +408,63 @@ public Uni<DeleteAccountResponse> deleteAccount(DeleteAccountRequest request) {
...
+                      clearAccountDeletionFence(accountId.getId(), fencedMeta);
+                      throw failedPrecondition;
+                    }
+                    advanceAccountDeletionFence(accountId.getId(), fencedMeta, current);
+                    throw new BaseResourceRepository.AbortRetryableException(
+                        "account changed while deletion fence was installed");
                   }
-
-                  return DeleteAccountResponse.newBuilder().setMeta(out).build();
+                  cleanupAccountResources(accountId);
```
mrouvroy-floe, line 471: The deletion marker written by ensureAccountDeletionFence is a durable tombstone that is never removed on the successful completion path. After cleanupAccountResources finishes, the marker at Keys.accountDeletionMarker(id) remains in place forever.

Every create call asserts CasCheckAbsent(accountDeletionMarker), and resolveAccountId allows callers to supply their own account id via request.hasAccountId(). This means deleting account foo and later attempting to recreate foo will be rejected forever with AccountDeletionInProgressException → MC_PRECONDITION_FAILED.account.deletion.in.progress ("Account "foo" is being deleted"), which is misleading since deletion completed long ago.

core/storage-spi/src/main/java/ai/floedb/floecat/storage/spi/PointerStore.java
```diff
@@ -140,6 +145,10 @@ default String pageTokenAfterKey(String key) {
 
   int countByPrefix(String prefix);
 
+  default int countByPrefixConsistent(String prefix) {
```
mrouvroy-floe, line 148: Is that ever used ?

service/src/main/java/ai/floedb/floecat/service/account/impl/AccountServiceImpl.java
mcusack-floe: Fixed to also delete the constraints pointers.

service/src/main/java/ai/floedb/floecat/service/account/impl/AccountServiceImpl.java
mcusack-floe: This is intentional. Once deletion commits the account ID is tombstoned.

core/storage-spi/src/main/java/ai/floedb/floecat/storage/spi/PointerStore.java
mcusack-floe: Yes, used in PR 451. Just estabilishing it as a primitive here.

service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java
```diff
@@ -363,8 +379,14 @@ private CreateCommit createInternal(
     return observeRepository(
         "create",
         () -> {
+          K resourceKey = schema.keyFromValue.apply(value);
+          String accountDeletionMarker = Keys.accountDeletionMarker(resourceKey.accountId());
+          if (mutationPointerStore.get(accountDeletionMarker).isPresent()) {
```
mrouvroy-floe, line 384: The new AccountDeletionInProgressException can be thrown by repo.create after callers have already performed out-of-band side effects. Specifically, ConnectorsImpl.createConnector calls storeAuthCredentials(...) before connectorRepo.create(connector), and only compensates on NameConflictException.

If account deletion wins the race, the fence rejects the create, no connector pointer is published, and cleanupAccountResources cannot discover the random connector id to delete its secret. The result is an orphaned credential secret for a deleted account — a persistent secret leak.

service/src/main/java/ai/floedb/floecat/service/account/impl/AccountServiceImpl.java
```diff
@@ -405,36 +413,63 @@ public Uni<DeleteAccountResponse> deleteAccount(DeleteAccountRequest request) {
...
-                          corr,
-                          "account",
-                          Map.of("id", accountId.getId()));
-
-                  if (out.getPointerVersion() == meta.getPointerVersion()
-                      && out.getEtag().equals(meta.getEtag())) {
-                    cleanupAccountResources(accountId);
+                  MutationOps.BaseServiceChecks.enforcePreconditions(
+                      corr, meta, request.getPrecondition());
+                  MutationMeta fencedMeta = ensureAccountDeletionFence(accountId.getId(), meta);
```
mrouvroy-floe, line 430: Installing Keys.accountDeletionMarker(accountId) only protects mutations routed through GenericResourceRepository. A transaction already in TS_PREPARED state before the fence is installed can still call commitTransaction; TransactionIntentApplierSupport.applyTransactionAtomically writes table, connector, and name pointers via raw PointerStore.compareAndSetBatch without ever checking the account-deletion marker.

This means a prepared transaction can commit after cleanupAccountResources has listed and deleted descendants, recreating account-scoped resources under an account row that has already been deleted. The delete can return successfully while orphaned tables/connectors remain reachable.

service/src/main/java/ai/floedb/floecat/service/account/impl/AccountServiceImpl.java
mcusack-floe: Account cleanup now pages through all storage auths deleting each secret first, then removing repo artifacts.

service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java
mcusack-floe: All connector create paths now delete the credential if a create is rejected by the account delete fenche.

service/src/main/java/ai/floedb/floecat/service/account/impl/AccountServiceImpl.java
mcusack-floe: We now pre-check the account deletion marker. If deletion installs the fence the entire batch fails without publishing.

service/src/main/java/ai/floedb/floecat/service/account/impl/AccountServiceImpl.java
```diff
@@ -582,10 +707,10 @@ private void cleanupTable(
     CLEANUP_LOG.infof(
         "account_delete_cleanup_table account_id=%s namespace_id=%s table_id=%s",
         tableId.getAccountId(), table.getNamespaceId().getId(), tableId.getId());
-    tableRepo.delete(tableId);
+    purgeSnapshotsAndStats(tableId, summary);
```
mrouvroy-floe: StatsRepository.TargetStatsStorage and IndexArtifactRepository.registerChunk build raw PointerStore.CasUpsert batches without checking Keys.accountDeletionMarker(...). GenericResourceRepository mutations include CasCheckAbsent(Keys.accountDeletionMarker(...)), but these two writers do not, so they can succeed after ensureAccountDeletionFence has installed the marker and after cleanupTable has already deleted the relevant prefixes — recreating stats/index pointers and blobs under the deleting account.

Add the deletion-marker precondition to every raw pointer transaction that writes stats or index artifacts, and preflight the marker before writing associated blobs.

service/src/main/java/ai/floedb/floecat/service/account/impl/AccountServiceImpl.java
mrouvroy-floe: purgeSnapshotsAndStats now deletes snapshot/constraint pointer prefixes and the table root pointer; cleanupTable removes the table record. Neither step deletes the table-owned blob subtree under Keys.tableBlobPrefix(...), which contains every blob family owned by the table (e.g. target-stats/, index-sidecars/). The pointer-prefix deletes do not remove those blobs, and the CAS blob GC path does not issue a table-prefix delete once the table row has already been removed.

We should add a regression test that asserts no blobs remain under the table prefix after account deletion.

service/src/main/java/ai/floedb/floecat/service/account/impl/AccountServiceImpl.java
mcusack-floe: All raw stats/index publication CAS operations now include CasCheckAbsent in the same transaction.

kwesterfeld-floe (commented):
## Caveat

These findings come from diffing the implementation I did for recursive delete.  The findings refer to "v2" which is my second version of the recursive delete PR.

## Findings

1. The teardown parses every blob, and one bad blob fences the account forever.

This is the big one. His walk enumerates through connectorRepo.listConsistent(...), catalogRepo.listConsistent(...), tableRepo.listConsistent(...) — all of which return parsed protos, so every blob gets read. CorruptionException extends RepoException, not AbortRetryableException, so nothing retries it.

v2 hit exactly this and moved the entire sweep to pointer-row-driven enumeration (forEachId), because identity is all teardown needs and the repositories' delete already tolerates an unreadable blob. One truncated connector or catalog blob otherwise orphans every resource behind it, permanently.

In his design the consequence is worse than it was in v2. The tombstone is never cleared on the success path — clearAccountDeletionFence is called from exactly one place, the failed-precondition branch. So a corrupt blob leaves the account: cleanup unfinished, marker permanently up, every create rejected with AccountDeletionInProgressException, and every DeleteAccount retry re-entering cleanupAccountResources to die on the same blob. The account id becomes permanently unwritable and permanently undeletable. v2 could at least make progress on retry.

2. Residual index rows after a corrupt-blob delete.

Deleting a resource whose blob can't be read removes the canonical pointer but not the by-name row — because the name that row indexes lives in the blob that can't be read. Those rows survive teardown with nothing left to enumerate them. v2 added a paged, versioned residual sweep over both the by-id and by-name families for catalogs and connectors. I don't see an equivalent here.

3. Children markers are never swept.

catalogChildrenMarker is /accounts/{a}/catalogs/{c}/markers/children and namespaceChildrenMarker is /accounts/{a}/namespaces/{n}/markers/children. Neither sits under a by-id or by-name prefix, so PointerGc — which scans /accounts/{a}/<kind>/by-id/ and by-name — never sees them, and his teardown doesn't touch them. His blobStore.deletePrefix(accountRootPrefix) sweeps blobs, not pointer rows.

Result: one permanent pointer row per catalog and per namespace the account ever had. This is pre-existing on main, but his is the PR claiming account deletion is now complete, so it lands in scope.

mcusack-floe:
> ## Caveat
> These findings come from diffing the implementation I did for recursive delete. The findings refer to "v2" which is my second version of the recursive delete PR.
> 
> ## Findings
> 1. The teardown parses every blob, and one bad blob fences the account forever.
> 
> This is the big one. His walk enumerates through connectorRepo.listConsistent(...), catalogRepo.listConsistent(...), tableRepo.listConsistent(...) — all of which return parsed protos, so every blob gets read. CorruptionException extends RepoException, not AbortRetryableException, so nothing retries it.
> 
> v2 hit exactly this and moved the entire sweep to pointer-row-driven enumeration (forEachId), because identity is all teardown needs and the repositories' delete already tolerates an unreadable blob. One truncated connector or catalog blob otherwise orphans every resource behind it, permanently.
> 
> In his design the consequence is worse than it was in v2. The tombstone is never cleared on the success path — clearAccountDeletionFence is called from exactly one place, the failed-precondition branch. So a corrupt blob leaves the account: cleanup unfinished, marker permanently up, every create rejected with AccountDeletionInProgressException, and every DeleteAccount retry re-entering cleanupAccountResources to die on the same blob. The account id becomes permanently unwritable and permanently undeletable. v2 could at least make progress on retry.
> 
> 2. Residual index rows after a corrupt-blob delete.
> 
> Deleting a resource whose blob can't be read removes the canonical pointer but not the by-name row — because the name that row indexes lives in the blob that can't be read. Those rows survive teardown with nothing left to enumerate them. v2 added a paged, versioned residual sweep over both the by-id and by-name families for catalogs and connectors. I don't see an equivalent here.
> 
> 3. Children markers are never swept.
> 
> catalogChildrenMarker is /accounts/{a}/catalogs/{c}/markers/children and namespaceChildrenMarker is /accounts/{a}/namespaces/{n}/markers/children. Neither sits under a by-id or by-name prefix, so PointerGc — which scans /accounts/{a}//by-id/ and by-name — never sees them, and his teardown doesn't touch them. His blobStore.deletePrefix(accountRootPrefix) sweeps blobs, not pointer rows.
> 
> Result: one permanent pointer row per catalog and per namespace the account ever had. This is pre-existing on main, but his is the PR claiming account deletion is now complete, so it lands in scope.

The account-deletion path now:
- Enumerates resources from canonical by-ID pointer keys without parsing blobs, so dangling or corrupt blobs cannot wedge deletion.
- Performs credential/secret cleanup before the terminal sweep.
- Cleans reconcile jobs, attributable global lookup and lease-expiry rows, and supports empty or corrupt cleanup manifests.
- Fences account-scoped reconcile writes during deletion.
- Sweeps the complete account pointer subtree while continuously preserving /deleting, then verifies only the fence remains.
- Deletes account blobs and invalidates metadata/root caches after the pointer sweep.
Supporting fixes include:
- Correct pagination and encoded-ID handling during reconcile cleanup.
- Transaction-budget accounting that includes fence checks.
- Convergence handling for ready, state, expiry, and cancellation indexes after canonical jobs disappear.
- Best-effort orphan-index cleanup on read paths.
- Atomic removal of secondary pointers when deleting corrupt resources.
- Percent-encoding fixes for reconcile lease keys.
- Compile-time requirements for teardown-critical store methods.
- Removal of obsolete tree-walk and mutation-read APIs.
- Storage-layout documentation updates.

kwesterfeld-floe (approved):
This is better than what I came up with. 

Two things I'd consider adding, and if they are easy to do now both might be worth it.

1. There is value in a bit more bounded detail in logs when things go wrong.  For example:
> "account pointer sweep left 3 rows under /accounts/x/" 

this tells you it failed but nothing about what to do about it. Logging the first N surviving keys would make this actionable (ie. go fix the mess leftover, develop a follow-on patch, etc).

2. There is no fence enforcement for follow on code.  This one is tricky, but the AI god proposed using something called archunit to pin the behavior allowed for account-scoped pointer writes to code that calls specific methods.  Then if you write new code that doesn't call `CasCheckAbsent`, then the new code fails to pass tests and tells you why.

PR #452, opened 2026-08-12
feat(integration): add catalog integration API (stack 5/10)

mcusack-floe:
## Summary

Adds the catalog integration resource and gRPC API, including persistence, idempotent CRUD behavior, authorization, credential storage and cleanup, garbage-collection hooks, account-deletion participation, and service/storage documentation.

Catalog overlays and CLI commands are intentionally excluded into subsequent PRs.

## Type of Change

- [ ] fix (bug fix)

## Testing

Focused catalog integration API/service suite: 117 tests passed.

- [x] `make fmt`
- [ ] `make verify`

- [x] No breaking changes

## Checklist

## Additional Notes

Stack position: **5/10**. Base: `mc-account-deletion-fence`. Review the integration resource independently of overlays, CLI, and connector-independent access.

### DIFF
--- a/docs/storage-layout.md
+++ b/docs/storage-layout.md
@@ -26,6 +26,7 @@ Resource and transaction blobs:
 /accounts/{account_id}/tables/{table_id}/table/{sha}.pb
 /accounts/{account_id}/views/{view_id}/view/{sha}.pb
 /accounts/{account_id}/connectors/{connector_id}/connector/{sha}.pb
+/accounts/{account_id}/catalog-integrations/{integration_id}/integration/{sha}.pb
 /accounts/{account_id}/transactions/{tx_id}/transaction/{sha}.pb
 /accounts/{account_id}/transactions/{tx_id}/intent/{sha}.pb
 /accounts/{account_id}/transactions/{tx_id}/objects/{sha}.bin
@@ -89,11 +90,29 @@ Catalog hierarchy, lookup indexes, and maintenance markers:
 /accounts/{account_id}/catalogs/{catalog_id}/namespaces/{namespace_id}/relations/by-name/{relation_name}
 /accounts/{account_id}/connectors/by-id/{connector_id}
 /accounts/{account_id}/connectors/by-name/{connector_name}
+/accounts/{account_id}/catalog-integrations/by-id/{integration_id}
+/accounts/{account_id}/catalog-integrations/by-name/{integration_name}
+/accounts/{account_id}/catalog-integrations/overlays-marker/{integration_id}
+/accounts/{account_id}/catalog-integrations/deleting/{integration_id}
+/accounts/{account_id}/deleting
+/catalog-integration-credential-cleanup/{account_id}/{integration_id}/{generation}
 /accounts/{account_id}/catalogs/{catalog_id}/markers/children
 /accounts/{account_id}/namespaces/{namespace_id}/markers/children
 /accounts/{account_id}/gc/cas/generation-cursor
 /accounts/{account_id}/root-resyncs/by-table/{table_id}
 ```
+
+Catalog integration secret payloads are stored outside the resource blob through SecretsManager:
+
+```
+accounts/{account_id}/catalog-integrations/{integration_id}:{credential_generation}
+```
+
+The key is derived internally from the persisted integration identity and credential generation; it
+is not present in the resource payload. A new secret is written before publishing the matching
+generation; an old secret is deleted only after the resource mutation commits. Definite failures
+delete an unpublished generation. Durable cleanup records retain other candidates until pointer GC
+can prove that their generation is no longer referenced.
 
 Transaction and idempotency pointers:
 
--- a/docs/storage-layout.md
+++ b/docs/storage-layout.md
@@ -105,7 +105,7 @@ Catalog hierarchy, lookup indexes, and maintenance markers:
 Catalog integration secret payloads are stored outside the resource blob through SecretsManager:
 
 ```
-accounts/{account_id}/catalog-integrations/{integration_id}:{credential_generation}
+accounts/{account_id}/catalog-integrations/{integration_id}.credentials.{credential_generation}
 ```
 
 The key is derived internally from the persisted integration identity and credential generation; it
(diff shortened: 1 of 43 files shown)


service/src/main/java/ai/floedb/floecat/service/repo/impl/CatalogIntegrationRepository.java
```diff
@@ -0,0 +1,240 @@
...
+    requiredAbsent.add(newMarker);
+    requiredAbsent.add(oldDeletionFence);
+    requiredAbsent.add(newDeletionFence);
+    if (expectedOverlayMarkerVersion == 0L) requiredAbsent.add(oldMarker);
+    return repo.replaceIdentityWithMeta(
+        current,
+        expectedPointerVersion,
+        replacement,
+        new PointerConditions(Map.of(), Set.copyOf(requiredAbsent), Map.of()),
+        expectedOverlayMarkerVersion == 0L
```
mrouvroy-floe: replaceIdentityWithMeta only requires the old overlay marker to be absent when expectedOverlayMarkerVersion == 0; once the marker exists, it passes that marker as a companion pointer to delete. deleteWithPreconditionAndOverlayMarker repeats the same pattern for normal deletes.

This means CM_REPLACE and DeleteCatalogIntegration can succeed in exactly the state the design says must be rejected: an existing overlay dependency marker. The marker is the only dependency signal introduced in this stack, so deleting it here would let a replacement silently retarget or orphan follow-on overlays instead of failing while dependents exist.

Make public replace/delete require the overlay marker to be absent, or add an explicit overlay-existence check and reserve marker deletion for a true cascading delete path.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
```diff
@@ -0,0 +1,834 @@
...
+    boolean storesCredentials =
+        credential != CatalogIntegrationCredentials.CredentialCase.CREDENTIAL_NOT_SET;
+    var persisted =
+        requested.toBuilder()
+            .setCredentialsConfigured(storesCredentials)
+            .setCredentialGeneration(storesCredentials ? generation : 0L);
+    return new PreparedAuthentication(persisted.build(), credentials);
+  }
+
+  private static void validateAssumeRole(
```
mrouvroy-floe: validateAssumeRole only checks role_arn and (when present) role_session_name; the external_id field of AwsAssumeRoleAuthentication is never validated, in either the top-level AWS_ASSUME_ROLE case or the aws_sigv4 nested one.

Because prepareAuthentication persists the whole requested CatalogAuthentication verbatim, a client can set external_id to an explicitly-present-but-blank value and have it stored. external_id is exactly the field that defends against the confused-deputy problem on cross-account sts:AssumeRole, so silently persisting an empty or garbage value weakens that protection once overlays start consuming the config.

docs/storage-layout.md
```diff
@@ -88,12 +89,30 @@ Catalog hierarchy, lookup indexes, and maintenance markers:
...
 /accounts/{account_id}/catalogs/{catalog_id}/markers/children
 /accounts/{account_id}/namespaces/{namespace_id}/markers/children
 /accounts/{account_id}/gc/cas/generation-cursor
 /accounts/{account_id}/root-resyncs/by-table/{table_id}
 ```
 
+Catalog integration secret payloads are stored outside the resource blob through SecretsManager:
+
+```
+accounts/{account_id}/catalog-integrations/{integration_id}:{credential_generation}
```
mrouvroy-floe: The storage doc says catalog-integration secrets live at ```accounts/{account_id}/catalog-integrations/{integration_id}:{credential_generation}```.
The implementation derives the SecretsManager key from secretType = "catalog-integrations" and CatalogIntegrationCredentialStore.reference() = "{integration_id}.credentials.{generation}", so the actual key is ```accounts/{account_id}/catalog-integrations/{integration_id}.credentials.{generation}``` after segment encoding.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationCredentialStore.java
```diff
@@ -0,0 +1,147 @@
...
+        == CatalogIntegrationCredentials.CredentialCase.CREDENTIAL_NOT_SET) {
+      return;
+    }
+    byte[] payload = credentials.toByteArray();
+    String reference = reference(integrationId, credentialGeneration);
+    if (secretsManager.get(integrationId.getAccountId(), SECRET_TYPE, reference).isPresent()) {
+      return;
+    }
+    try {
+      secretsManager.putIfAbsent(integrationId.getAccountId(), SECRET_TYPE, reference, payload);
```
mrouvroy-floe: Should we do something specific on ```putifAbsent``` returning false ?

service/src/main/java/ai/floedb/floecat/service/repo/impl/CatalogIntegrationRepository.java
mcusack-floe: Fixed. Replace/delete requires overlay marker to be absent.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationCredentialStore.java
mcusack-floe: store() now re-reads after putIfAbsent(false). It accepts the generation only if the secret is visible, otherwise it throws AbortRetryableException.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationsImpl.java
mcusack-floe: An explicitly blank external_id is now rejected.

docs/storage-layout.md
mcusack-floe: Updated doc.

kwesterfeld-floe:
Two issues.  

1. No IT for the new entity type here.  Probably needs some similar IT added before the other PRs land and make this harder.
2. Logging.  There isn't any.  Add some like we have sprinkled through the ConnectorsImpl service, for example, which this is striving to replace.

mcusack-floe:
> Two issues.
> 
> 1. No IT for the new entity type here.  Probably needs some similar IT added before the other PRs land and make this harder.
> 2. Logging.  There isn't any.  Add some like we have sprinkled through the ConnectorsImpl service, for example, which this is striving to replace.

Added logging and a new IT that covers the CRUD Catalog Integration lifecycle.

kwesterfeld-floe (approved):

PR #458, opened 2026-08-18
reconciler: scope reconcile jobs to a deployment cohort so incompatible versions can share one durable queue

mcusack-floe:
## Summary

Two Floecat deployments that share a durable job queue currently lease each other's
reconcile jobs. That is fine while both understand the same job-tree contract, and
harmful when they do not: a job tree planned by one version gets continued by another
that interprets its payloads differently. This blocks two things we want - rolling
upgrades across a breaking job-tree change, and pointing a staging or CI deployment at
a shared Dynamo table.

This adds an optional deployment cohort. When `floecat.reconciler.worker-affinity` is
set, the job store stamps that cohort onto every job it enqueues and constrains every
lease to it, so a cohort only ever processes its own trees.

The cohort is a property of the deployment, not of a producer or an executor, so both
stamps live at the job-store boundary - `ReconcileJobEnqueuer` on the way in and
`leaseNext` on the way out. That is deliberate: an earlier shape had each producer stamp
its own execution policy, which meant any code path that built a policy from scratch
silently enqueued unstamped work that no worker in an affinity-configured deployment
would ever lease. Stamping at the single choke point each store already has makes that
class of bug unrepresentable, and `ReconcileControlImpl`, `ReconcilePlannerScheduler`
and `TransactionsServiceImpl` need no affinity awareness at all.

The cohort is carried in the execution policy's attributes, which are already persisted
and already propagate to every child enqueue, so a whole job tree inherits its cohort
with no extra plumbing and no schema change.

Ready-queue indexes are partitioned per cohort by qualifying the execution-class,
execution-lane and job-kind filter values. Those values are percent-encoded into a
single key segment, so this needs no key-format change, and each cohort keeps its own
slice of every filtered index rather than sharing one undifferentiated scan. Notably
the cohort is not modelled as an executor pin: `pinnedExecutorId` sends a record to
the pinned index alone, which would have collapsed every cohorted job into one slice
and discarded lane and class routing entirely.

Also folds the duplicated eligibility rules into one implementation each, since the
duplication is what made the above easy to get wrong:

- `ReconcileExecutorRegistry.pinExcludes` is now the only definition of executor-pin
  eligibility, used by both `executorFor` and the lease poller.
- `LeaseRequest.matches` is now the only definition of job-vs-request eligibility; the
  native store, its in-memory double and the in-memory job store all delegate to it.

## Type of Change

- [ ] fix (bug fix)
- [X] refactor (no behavior change)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes
- [X] Config/env var changes (describe below)

## Checklist

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/ReadyQueueKeys.java
+++ b/service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/ReadyQueueKeys.java
@@ -89,6 +89,8 @@ public final class ReadyQueueKeys {
       return List.of();
     }
     List<String> keys = new ArrayList<>();
+    ReconcileWorkerAffinity workerAffinity =
+        ReconcileWorkerAffinity.fromPolicy(record.executionPolicy());
     String pinnedExecutorId = record.pinnedExecutorId();
     if (!blank(pinnedExecutorId)) {
       String pinnedExecutorKey =
@@ -96,13 +98,11 @@ public final class ReadyQueueKeys {
               record,
               ReconcileReadyQueueStore.ReadyIndexType.PINNED_EXECUTOR,
               dueAtMs,
-              pinnedExecutorId);
+              workerAffinity.indexFilterValue(pinnedExecutorId));
       return pinnedExecutorKey.isBlank() ? List.of() : List.of(pinnedExecutorKey);
     }
     // A cohort keeps its own slice of each filtered index, so a shared queue does not force
     // every deployment's workers through one undifferentiated scan.
-    ReconcileWorkerAffinity workerAffinity =
-        ReconcileWorkerAffinity.fromPolicy(record.executionPolicy());
     keys.add(readyPointerKeyFor(record, dueAtMs));
     String executionClassKey =
         readyPointerKeyFor(
--- a/service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/ReadyQueueKeys.java
+++ b/service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/ReadyQueueKeys.java
@@ -16,6 +16,7 @@
 
 package ai.floedb.floecat.service.reconciler.jobs.durable.store;
 
+import ai.floedb.floecat.reconciler.jobs.ReconcileWorkerAffinity;
 import ai.floedb.floecat.service.reconciler.jobs.durable.model.StoredReconcileJob;
 import ai.floedb.floecat.service.repo.model.Keys;
 import java.util.ArrayList;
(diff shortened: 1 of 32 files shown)


mrouvroy-floe:
From what I understand, we want old/new reconcilers to be able to share the same durable queue without picking up job trees they don’t understand. This PR does that by stamping jobs with a configured cohort/affinity, then making lease requests only match jobs from the same cohort. It also qualifies the ready queue indexes so each cohort mostly scans its own slice.

That makes sense, but `worker affinity` looks like it describes the deployment, when the thing we’re trying to protect is the job-tree format.

During an upgrade, I would expect new code to keep draining old jobs while writing new-format jobs, and child jobs should stay on whatever format the parent tree already has.

But with the current exact-match model, that seems hard to express. In particular, if we default to `reconciler-v1`, existing unversioned jobs won’t be picked up after deploy. And since enqueue applies the current server affinity, a child job can get stamped based on which control-plane pod handled the parent completion, instead of inheriting from the parent.

Should we model this as an explicit job-tree format/version instead? Roots would get the current produced version, children would inherit from their parent, and workers would say which versions they can run. Then we can have an actual drain/migration phase for retiring old versions.

mcusack-floe:
> From what I understand, we want old/new reconcilers to be able to share the same durable queue without picking up job trees they don’t understand. This PR does that by stamping jobs with a configured cohort/affinity, then making lease requests only match jobs from the same cohort. It also qualifies the ready queue indexes so each cohort mostly scans its own slice.

That's exactly right.

> 
> That makes sense, but `worker affinity` looks like it describes the deployment, when the thing we’re trying to protect is the job-tree format.
> 
> During an upgrade, I would expect new code to keep draining old jobs while writing new-format jobs, and child jobs should stay on whatever format the parent tree already has.

This is corrected now. A versioned executor will only process versioned jobs from the durable queue that match version exactly. For backwards compat, unversioned workers will only pull unversioned jobs.
> 
> But with the current exact-match model, that seems hard to express. In particular, if we default to `reconciler-v1`, existing unversioned jobs won’t be picked up after deploy. And since enqueue applies the current server affinity, a child job can get stamped based on which control-plane pod handled the parent completion, instead of inheriting from the parent.

The idea is that unversioned queues and unversioned executors will continue to drain now. versioned executors/queues will only interact with each other.

> 
> Should we model this as an explicit job-tree format/version instead? Roots would get the current produced version, children would inherit from their parent, and workers would say which versions they can run. Then we can have an actual drain/migration phase for retiring old versions.

I think this implementation accompishes this but I might be missing something.

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/ReadyQueueKeys.java
mrouvroy-floe, line 95: Should this pinned index filter be qualified by the job-tree version/cohort too?

For class/lane/kind indexes we qualify the filter value by affinity, but the pinned path still seems to write the raw `pinnedExecutorId`. 

If an old/unversioned worker scans the same pinned executor id, I’m not seeing what keeps it out of a new versioned pinned job’s ready slice. I think we may want to version-qualify the pinned index value as well, while keeping `pinnedExecutorId` itself semantically unchanged.

mrouvroy-floe (commented):
Got it, I was assuming a more lenient rollout model.

If the contract is strict isolation, then this makes sense to me: unversioned workers drain unversioned jobs, versioned workers only pick up jobs with the exact same version, and before retiring a version we need to drain/cancel/replan whatever is left for it.

I do think we should write that explicitly somewhere though, and ideally have a test for it. The important bit is that once we default to `reconciler-v1`, any old unversioned jobs still rely on unversioned workers staying up. The new versioned deployment won’t pick them up, by design.

service/src/main/java/ai/floedb/floecat/service/reconciler/jobs/durable/store/ReadyQueueKeys.java
mcusack-floe: I've fixed this. pinned executors are now qualified by job version, alongside class, lane and kind. I've also updated the contract in the docs.

mrouvroy-floe (approved):

PR #467, opened 2026-08-27
delta: extract the Unity Catalog HTTP surface behind a typed client

kwesterfeld-floe:
## Summary

Extracts Unity Catalog HTTP access from the Delta connector into a thin, typed client module. URL construction, pagination, JSON normalization, and HTTP error mapping now sit behind `UnityCatalogClient`, so connector-domain code can use mocks and client behavior can be tested against a local HTTP server.

Changes:

- Add normalized client records and a transport-neutral exception with semantic failure kinds.
- Confine JDK HTTP types and status handling to `HttpUnityCatalogClient`.
- Cover pagination, table-property wire shapes, omitted collections, and failure classification.
- Delete the unused Databricks SQL statement client, warehouse option, and superseded HTTP helpers.
- Normalize missing names used as sort keys and close the client transport with the connector lifecycle.

The credentials operation is present on the client boundary but is not consumed until a later change.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [ ] No breaking changes
- [x] Breaking changes: Unity Catalog endpoints must use `https://`. Cleartext to a loopback host,
  and base URIs naming a private address literal, each require an explicit operator opt-in;
  link-local, wildcard and multicast literals are always rejected. Timeouts must be positive, and a
  base URI carrying userinfo is rejected.
- [x] Config/env var changes:
  - `floecat.security.allow-loopback-catalog-endpoints` /
    `FLOECAT_SECURITY_ALLOW_LOOPBACK_CATALOG_ENDPOINTS` — permit cleartext `http://` to a loopback
    host.
  - `floecat.security.allow-private-catalog-endpoints` /
    `FLOECAT_SECURITY_ALLOW_PRIVATE_CATALOG_ENDPOINTS` — permit a base URI naming a private address
    literal (10/8, 172.16/12, 192.168/16, IPv6 unique-local).
  - `unity.temporary-table-vend-path` connector option — vending route, defaulting to the Databricks
    2.0 path; set the OSS 2.1 path for an OSS Unity Catalog endpoint. Named to avoid the
    secret-bearing "credentials" phrase, which `PersistedSecretPropertyValidator` refuses to
    persist.
  - `floecat.unity.max-pages` system property — cap on pages fetched by one listing, default
    10,000; exceeding it is reported as `INVALID_RESPONSE`.
  - `floecat.unity.max-response-bytes` system property — cap on a single response body, default
    32 MiB; a larger body is refused rather than buffered.

  Private-CA HTTPS endpoints need no flag, only the JVM truststore. See `docs/connectors-delta.md`.
- **Intermediate state, folded in a follow-up stacked PR:** the compose smoke Unity scenario is
  disabled here (`COMPOSE_SMOKE_UPSTREAM_DELTA_UNITY_IMPORT=false`) and its
  `COMPOSE_SMOKE_UPSTREAM_DELTA_UNITY_URI` default still points at `http://unity:8080`. Neither is
  the intended end state: PR 474 fronts the endpoint with a TLS proxy, moves the default to
  `https://unity-proxy:8443`, and re-enables the scenario. No deployment consumes these values.

## Checklist

## Reviewers

- **The Databricks CLI auth path is out of bounds for this PR.** `DatabricksCliTokenProvider`
  predates this stack (#138). The only reason it appears in this diff is transport ownership: the
  extracted client takes a per-request header supplier, so the provider now receives an
  `HttpClient` instead of building one, exposes `close()` so the factory can release it, and
  resolves its cache path before that client is allocated so an unusable path cannot leak a started
  selector thread. Thirty-five lines, all mechanical.

  Nothing about how that provider *behaves* is changed here — not what it does with a failed
  refresh, not how a missing login or a revoked refresh token classifies, not what it does with an
  interrupt. An earlier revision of this PR did change some of that and it was removed as unrelated
  to extracting an HTTP client. `oauth.mode=cli` is a workstation auth mode; a production Unity
  connector authenticates with a resolved bearer token. Improvements there belong in their own PR.

  The one client-side change that does touch this area is deliberate and generic: an auth provider
  that throws while the thread's interrupt flag is set is reported as `INTERRUPTED` rather than
  `TRANSPORT`, because that classification lives in the client this PR extracts and the failure
  otherwise reads as retryable. It applies to any `AccessTokenProvider`, not just the CLI one.

- **Transport gates are deny-by-default, with two named opt-ins.** HTTPS is required. Cleartext to a
  loopback host needs `floecat.security.allow-loopback-catalog-endpoints`; a base URI naming a
  private address literal needs `floecat.security.allow-private-catalog-endpoints`. Both default to
  deny, mirroring `floecat.security.allow-loopback-token-endpoints` in `CredentialResolverSupport`,
  so a tenant-supplied connector URI cannot grant itself either. Link-local, wildcard and multicast
  literals are always rejected, including the cloud metadata address. **These checks are
  deliberately literal-only:** deciding them for a hostname would mean resolving it, and a resolver
  in the gate disagrees with the resolution `HttpClient` performs at connect time. A hostname
  pointing at a private address is therefore out of scope and belongs to network policy — this
  raises the cost of the obvious probe without claiming to be an SSRF boundary. Private-CA HTTPS
  needs no flag, only the JVM truststore. See `docs/connectors-delta.md`.
- **Failure classification runs in three passes, most authoritative first.** Statuses that answer
  the question themselves (401, 403, 429, 405, 408, 422, 423, 425) are classified from status alone,
  so no body code can present an auth failure as `NOT_FOUND` or make a rate limit terminal. Only the
  genuinely ambiguous 400, 404, and 409 consult `error_code`: a Databricks 400 whose real reason is
  in the envelope; a 404, where an explicit denial must not degrade to `Optional.empty()`; and a
  409, whose code distinguishes a permanent existing-object conflict from a retryable bare
  conflict. A bare 404 stays `NOT_FOUND`, because Unity Catalog OSS does not send
  Databricks' envelope. A bare 409 is `TRANSIENT`. A 5xx is retryable whatever its body says, and a 3xx is a base-URI
  misconfiguration rather than whatever its body claims.
- **The `Failure` enum's consumer is the next PR, not this one.** Read alone, this PR adds
  classification logic that nothing branches on. #475 switches on `failure()` to decide vending
  behavior. The retry consequences described in the enum and `httpFailure` Javadoc are the contract
  that PR relies on.
- **Credential redaction has three parts, and they have to stay aligned.** `AwsCredentials`
  overrides `toString()` so secrets cannot reach a log through the record; response bodies are kept
  out of exception messages on the vending path; and that suppression is keyed on the *configured*
  credentials route, not a constant, so selecting a non-default route does not quietly opt out of
  it.
- **The vending route is selectable.** Databricks serves it under 2.0 and OSS Unity Catalog 0.6.0+
  under 2.1; every other operation is 2.1 on both. The default is Databricks;
  `unity.temporary-table-vend-path` selects the OSS route or one a proxy exposes.
- **Column validation is call-path-specific.** `getTable` rejects malformed `columns` because its schema
  is authoritative. A list response still decodes valid columns for view descriptors, but malformed
  columns degrade only that entry to an empty column list so one bad table cannot block discovery
  of the rest of the namespace.
- **Base URI hygiene:** userinfo is rejected, and checked before every gate whose message
  interpolates the URI, so a rejection cannot echo an embedded credential. Pagination is bounded by
  both a repeated-token guard and a page cap, so a server minting fresh tokens fails as a
  classified response error rather than looping.
- **Body reads are not bounded by `http.read.ms`, and that is pre-existing — please do not block on
  it.** The JDK's request timeout bounds time-to-headers, not body completion. A server that sends
  headers and then stalls pins the calling thread regardless of the timeout. This is *not* a
  consequence of the bounded body handler added here: `BodyHandlers.ofString` behaves identically
  (measured: 8.0s elapsed against a 2s timeout, no exception), so the behaviour predates this PR and
  is unchanged by it. What this PR does change is the other half — a firehose response is now
  refused at `floecat.unity.max-response-bytes` instead of being buffered whole. Reverting the cap
  would reintroduce the heap exhaustion without addressing the stall.
  `HttpClient.close()` can consequently wait without a deadline for the same stalled operation.
  Fixing the body deadline and giving shutdown a bounded fallback belong together; they require a
  read deadline or bounded `BodySubscriber` plus shutdown tests, which is larger than this extract.
  A follow-up issue covering both request and close paths will be filed once this lands.
- **Listing size is bounded per request, not in aggregate — deliberately, and please do not block on
  it.** `MAX_PAGES` caps how many requests one listing makes and
  `floecat.unity.max-response-bytes` caps each response; there is no cumulative budget across
  pages, so the theoretical worst case is their product. The original code had neither limit, so
  this is strictly tighter than what it replaced. A cumulative budget was written and then reverted:
  it needed a second return value out of `send`, which was smuggled through a mutable field and
  promptly became a thread-safety bug, and its default was derived from the per-response cap rather
  than being independently bounded. Doing it properly means changing what `send` returns, which is
  more than this extract should carry. A follow-up issue will be filed once this lands.

- **Glue construction cleanup is separate hardening.** The Glue branch has a narrow, pre-existing
  cleanup window if `DeltaGlueConnector` construction throws after its refreshing client is made.
  This Unity extraction neither introduces nor widens that window, and the intervening constructor
  is assignment-only today; keep that cleanup out of this stack and address it with Glue ownership
  tests if the constructor gains fallible work.
- **Stacked Unity smoke coverage:** do not treat this PR's temporary
  `COMPOSE_SMOKE_UPSTREAM_DELTA_UNITY_IMPORT=false` or its interim `http://unity:8080` default as
  the final coverage or configuration state. This is an intentional intermediate layer: PR 474
  moves the endpoint to `https://unity-proxy:8443`, re-enables the scenario, and extends it through
  end-to-end credential vending. The full stack retains and expands Unity smoke coverage; no
  additional change is needed here.

## Additional Notes

The change is confined to the Delta connector and the new client module. The removed SQL and HTTP helpers had no callers.


tools/compose-smoke.sh
```diff
@@ -33,7 +33,10 @@ COMPOSE_SMOKE_UPSTREAM_ICEBERG_EXPECTED_TABLE=${COMPOSE_SMOKE_UPSTREAM_ICEBERG_E
 COMPOSE_SMOKE_UPSTREAM_ICEBERG_WAREHOUSE=${COMPOSE_SMOKE_UPSTREAM_ICEBERG_WAREHOUSE:-quickstart_catalog}
 COMPOSE_SMOKE_UPSTREAM_ICEBERG_TABLE=${COMPOSE_SMOKE_UPSTREAM_ICEBERG_TABLE:-trino_types}
 COMPOSE_SMOKE_UPSTREAM_ICEBERG_METADATA_PREFIX=${COMPOSE_SMOKE_UPSTREAM_ICEBERG_METADATA_PREFIX:-s3://floecat/sales/us/trino_types/metadata/}
-COMPOSE_SMOKE_UPSTREAM_DELTA_UNITY_IMPORT=${COMPOSE_SMOKE_UPSTREAM_DELTA_UNITY_IMPORT:-true}
+# Disabled until PR 474 fronts the compose Unity endpoint with the TLS-backed unity-proxy. Until
```
mcusack-floe, line 36: Only noting that this means we have no test coverage between this one and PR474. If these are merged in relatively quick succession, That's fine.

mcusack-floe (approved):

PR #468, opened 2026-08-27
unity-catalog: harden the HTTP transport

kwesterfeld-floe:
## Summary

Stack 3/7. Base: `kw-uc-vending-02-client-extract` (#467).

Hardens the extracted Unity Catalog HTTP client at its configuration and transport boundaries.

Changes:

- Require HTTPS except for genuine loopback hosts, and refuse credential-bearing cleartext requests regardless of auth scheme.
- Follow only bounded, same-origin redirects so authorization is never replayed to an unconfigured authority.
- Keep secret-bearing response bodies out of errors.
- Preserve redirect and non-JSON failure identity instead of collapsing them into generic parse errors.
- Classify Databricks error envelopes and remaining 4xx responses so permanent refusals do not retry forever while server, rate-limit, and transport failures remain transient.
- Raise a typed failure when request headers cannot be constructed.

## Type of Change

- [x] fix (bug fix)

## Testing

Unity client module: 28 tests, 0 failures. Formatting clean.

- [x] `make fmt`
- [ ] `make verify`

- [ ] No breaking changes

Non-loopback `http://` Unity workspace URIs are now rejected. Local loopback stub servers remain supported.

## Checklist

## Additional Notes

Known follow-ups:

- A malformed or scheme-less base URI can still escape as raw `IllegalArgumentException`.
- A null auth-header value can still surface as `NullPointerException`.
- `getTable` maps any 404 to absent without inspecting the error envelope.
- The loopback carve-out resolves numeric-looking hostnames through DNS before deciding whether cleartext is allowed.

### DIFF
--- a/connectors/clients/unity-catalog/src/main/java/ai/floedb/floecat/client/unity/HttpUnityCatalogClient.java
+++ b/connectors/clients/unity-catalog/src/main/java/ai/floedb/floecat/client/unity/HttpUnityCatalogClient.java
@@ -20,8 +20,10 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
+import java.net.InetAddress;
 import java.net.URI;
 import java.net.URLEncoder;
+import java.net.UnknownHostException;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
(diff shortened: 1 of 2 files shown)


kwesterfeld-floe:
Closing. The transport hardening now lives in #467 (`471fcf6ee`), which took a different and simpler approach: refuse redirects outright (`Redirect.NEVER`) and require HTTPS for a non-loopback base URI.

That obsoletes most of what is here. The same-origin comparison, port normalization, hop counter, per-request cleartext check, and five redirect tests all exist to make redirect *following* safe — none of it has a purpose once redirects are not followed. Following was never useful for the credentials call anyway: `Redirect.NORMAL` re-issues a 301/302/303 as a bodyless GET, so the one POST in this client cannot survive a redirect regardless.

The one thing worth keeping from this branch was the auth-header classification, and that has been carried into #467 as `0d4ceac24`, split more finely than it was here: a provider that throws while producing the header map stays retryable (a token endpoint can be briefly unreachable), while a header value the client cannot carry — including a null value, which this branch did not handle — is classified permanent.

#469 has been rebased off this branch onto #467.

PR #475, opened 2026-08-28
delta: vend storage credentials from Unity Catalog

kwesterfeld-floe:
## Summary

Extends the existing source-catalog credential-vending contract to Databricks Unity Catalog. When no storage authority covers a table, an explicitly opted-in Delta connector can request short-lived, table-scoped AWS credentials from the Unity temporary-table-credentials API.

Changes:

- Gate vending on `databricks.access-delegation=vended-credentials` and reject unrecognized values during create/update.
- Share the opt-in dispatcher between both service gates so they cannot disagree.
- Fall back to a storage authority when Unity returns no recognized credential shape.
- Classify permanent credential refusals as terminal while preserving retryability for transient failures.
- Carry non-authorization terminal refusals as their own structured reason.
- Add catalog scope to the shared vending record and centralize epoch-millis expiry parsing.

The scope/expiry changes also affect the existing Iceberg vending contract.

## Type of Change

- [ ] fix (bug fix)

## Testing

Full build, 0 failures. Formatting clean.

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

Unity credential vending is disabled unless the connector explicitly sets `databricks.access-delegation=vended-credentials` (accepted boolean aliases are also validated). Existing connectors retain storage-authority behavior.

## Checklist

## Reviewers

- **The vending SPI is deliberately not wired here — #476 does that, and please do not block on it.**
  `SourceCatalogVending` and `SourceCatalogAccessException` are the contract this PR defines; the
  production admission gate and the failure-status mapping that consume them land in the next PR of
  the stack. Read alone, this PR looks inert: `SourceCatalogCredentialVendor` still calls
  `IcebergAccessDelegation.declaresVendedCredentials` directly, and
  `SourceCatalogVendingGrpcStatus.sourceCatalogVendRefused` has no production caller. Both are
  addressed in #476, which routes the gate through `SourceCatalogVending.declaresVendedCredentials`,
  uses `clientAppliesVendedCredentials` at the reconcile boundary, and adds an
  `instanceof SourceCatalogAccessException` arm mapping `PERMISSION_DENIED` / `UNAUTHENTICATED` /
  `UNSUPPORTED` to terminal statuses, with tests. Nothing further is needed in this PR.
- **The `scopePrefix` contract is stated here; the consumer that honors it is in #476.** This PR
  documents `VendedStorageCredentials.scopePrefix` as a bound to intersect with the
  caller-authorized location, never a substitute for it. A non-null value is not necessarily
  catalog-issued — the Iceberg connector substitutes the table location when the catalog gives no
  prefix, and `credentialsFromVendChannel` picks the longest matching prefix, where an empty prefix
  matches everything — so a catalog that legitimately returns a warehouse-wide scope for one table
  must not widen what travels back. `SourceCatalogCredentialVendor.responsePrefix` in #476 is what
  enforces it, taking the vended scope only when it narrows the request. That method's own prose
  still described the scope as "catalog-issued"; it is corrected in #476, not here.
- **`SourceCatalogAccessException` carries no cause.** Its only constructor is
  `(Denial, String)`, so the originating `UnityCatalogException` — and its `statusCode()` and
  `errorCode()` — is not propagated. That is a real gap and is not fixed by #476; it is tracked as
  a follow-up rather than widening this PR's SPI surface.
- **An ambiguous table lookup stays retryable; only an enveloped refusal is terminal.**
  `HttpUnityCatalogClient.getTable` folds every `NOT_FOUND` into an empty `Optional`, so a table
  Unity genuinely dropped and an HTML 404 from a load balancer mid-deploy arrive identically. Both
  therefore fall back rather than failing the job, because terminalizing the pair would permanently
  fail a reconcile on a condition that recovers on its own. The same envelope requirement applies to
  `INVALID_REQUEST`, which the client derives from status alone for 405, 422 and any 3xx it declines
  to follow — an SSO proxy's login redirect among them. Terminalizing a genuine absence would need
  the client to preserve *why* a lookup came back empty (a typed lookup result rather than
  `Optional`), which is a change in #467; tracked as a follow-up.
- **Two delegation models coexist, and the kind boundary is resolved here but enforced in #476.**
  Iceberg signals vending per request with the `X-Iceberg-Access-Delegation` header (in properties
  or `auth.header_hints`); Unity signals it in configuration with `databricks.access-delegation`.
  `SourceCatalogVending` dispatches on `ConnectorConfig.kind()`, so a Delta connector routes to
  `DatabricksAccessDelegation` and its Iceberg header is ignored. `IcebergAccessDelegation`
  itself does not check kind, so it will answer "yes" for a Delta connector carrying that header;
  the boundary is held by the dispatcher rather than by the reader.

  That matters only between this PR and #476. This PR wires `vendStorageCredentials`, while the two
  live gates -- `SourceCatalogCredentialVendor` and `ServerSideStorageConfigResolver` -- still call
  `IcebergAccessDelegation.declaresVendedCredentials` directly; #476 points both at
  `SourceCatalogVending`. In between, a Delta connector with the Iceberg header would pass the gate
  and vend without the Databricks opt-in. The whole stack merges together, so that window does not
  occur in practice, and the same window is the only reason `s3.access-point` appears unconsumed:
  #476 adds the consumer, and nothing reaches Unity vending before it. Hardening
  `IcebergAccessDelegation` to reject a non-Iceberg kind on its own -- so the boundary does not
  depend on one caller -- is left to a later PR.

  On `s3.access-point` specifically: **at this PR nothing reads it.** `UnityDeltaConnector` puts the
  ARN in the property map and the only other mentions in the tree are the client that parses
  `access_point` and a comment in `IcebergConnector`. Every consumer arrives in #476 —
  `SourceCatalogCredentialVendor.VENDED_ROUTING_KEYS` becomes `("s3.region", "s3.endpoint",
  "s3.path-style-access", "s3.access-point")`, so the ARN reaches `client_safe_config` and survives a
  refresh via `getClientSafeConfigMap()`, and `ServerSideFileIoPropertiesResolver`,
  `DeltaConnectorFactory` and `DeltaManifestMaterializer` all read the flat key there. Iceberg is the
  exception rather than the convention: it has no such property and keys access points per bucket as
  `s3.access-points.<bucket>`, which is why `IcebergConnector`'s vended key list omits it. An earlier
  revision of this section claimed those readers existed here; they do not, and a reader of this PR
  alone should expect the key to be inert.

- **Vended credentials are not supported for `delta.source=glue` or `filesystem`, and the opt-in is
  not rejected there.** `DatabricksAccessDelegation.usesUnityCatalog` returns false for both, so
  `databricks.access-delegation=vended-credentials` reads as "not opted in" and vending never runs.
  This is scope, not a missing validation: Glue-backed vending is a different mechanism, requiring
  AWS Lake Formation and its own credential model, and belongs with the S3 Tables + Glue vending
  configuration story rather than here. A validator added now would have to be revisited when that
  lands. If one is added later, note that an absent `delta.source` reads as Unity — so the check
  must key on the option being explicitly present and not `unity`, not merely unequal to it.
- **A vend without `session_token` is rejected on the reconcile path and deliberately allowed on the
  query path.** `SourceCatalogCredentialVendor.requireUsableCredentials` requires
  `s3.access-key-id`, `s3.secret-access-key`, `s3.session-token` *and* an expiry for
  `CredentialUse.RECONCILE`, so an incomplete STS tuple fails at vend time with a named missing key.
  `CredentialUse.QUERY` requires only the key and secret, because it hands credentials straight to
  the scan engine's FileIO and registers no refresh provider, so a renewal requirement has no
  meaning there.

  Rejecting a missing `session_token` in `HttpUnityCatalogClient.awsCredentials` was considered and
  declined: a session token is absent exactly when the credentials are long-lived IAM keys rather
  than STS, which is how an OSS Unity Catalog deployment fronting static S3 credentials behaves, and
  `AwsBasicCredentials` is the correct choice for those. Treating absence as malformed would reject
  a working configuration to defend against a Unity response that is internally inconsistent.
- **`UNAUTHENTICATED` and `PERMISSION_DENIED` are terminal from status, without the envelope guard
  the other two arms carry.** `HttpUnityCatalogClient.httpFailure` derives them from 401 and 403
  alone, before the `error_code` step, so a 403 from a WAF, a corporate egress proxy, or an SSO
  gateway terminalizes a reconcile the same way a genuine Databricks denial does. The asymmetry with
  the `NOT_FOUND` / `INVALID_REQUEST` carve-out is real and deliberate rather than overlooked.

  Extending the guard was considered and declined for now. The reconciler retries a retryable
  failure indefinitely, so requiring an envelope would convert a genuine credential
  misconfiguration that arrives without one into an infinite loop — the failure mode this PR exists
  to remove. Databricks does envelope its own 401 and 403, which makes the change probably safe, and
  "probably" is not the basis on which to trade a visible terminal failure for an invisible loop. If
  it is taken later, `PERMISSION_DENIED` is the better half to gate: proxies return 403 far more
  often than 401.

- **The Databricks CLI auth path is deliberately untouched, and it has known gaps.**
  `DatabricksCliTokenProvider` and `OAuthRefreshSupport` predate this stack (#138). This PR moves
  neither and changes neither; earlier revisions did, and that work was backed out as out of scope.
  What that leaves, for `oauth.mode=cli` only — a workstation auth mode, not a service one:

  - `accessToken()` swallows every refresh failure and returns the cached token, so a token endpoint
    that is briefly unreachable yields an expired token, a 401, and — because of the classification
    this PR adds — a terminally failed reconcile. This is the one interaction the PR genuinely
    creates: the provider's behaviour is unchanged, but the consequence of it is new.
  - Permanence is not distinguished at the token endpoint. A revoked refresh token (`invalid_grant`)
    and a 5xx are both untyped, so neither can be classified differently downstream.
  - `InterruptedException` from the refresh is absorbed. #467 now reports an interrupted provider as
    `INTERRUPTED` rather than `TRANSPORT`, which stops a cancellation from burning a retry budget,
    but the provider still discards the interrupt before it gets there.
  - The cache is a read-modify-write over a file shared with other connector instances *and with the
    Databricks CLI itself*. Two writers lose one update, and with single-use refresh tokens the lost
    update restores a spent token to disk. Closing it needs a lock held across read, mutate and
    write — and the CLI would not take that lock, so even a lock is partial. Out of proportion to a
    dev auth path, and the reason the whole area is fenced rather than fixed here.
  - A missing login classifies as retryable. `readCacheForHost()` returning empty is wrapped as a
    `RuntimeException`, so "no token cached for this host" reads as `TRANSPORT` and is retried
    indefinitely rather than reported as something a human has to fix.
  - The refresh failure message carries the raw response body, unbounded.

  A production Unity connector authenticates with a resolved bearer token and none of this applies.
  Fixing the path belongs in its own PR against `DatabricksCliTokenProvider`.
- **A vend with no expiry still passes catalog-integration validation.** This PR's shared expiry
  parser reads absent, blank, non-positive and unparseable alike as "no expiry", so a catalog sending
  `"s3.session-token-expires-at-ms": "0"` now validates where it previously failed with
  `CIVI_CREDENTIAL_EXPIRED`. Failing on `expiresAt().isEmpty()` was tried and reverted: six existing
  `CatalogIntegrationDiscoveryTest` cases construct a vend with `Optional.empty()` and assert it
  passes, because most catalogs legitimately send no expiry. Separating "absent" from "present but
  zero" needs the raw property plumbed into the discovery path, which is more than the `"0"` shape
  warrants. The reconcile path **does** change, contrary to an earlier revision of this note: a
  `"0"` expiry previously parsed to `Instant.ofEpochMilli(0)`, passed `requireUsableCredentials`,
  and registered a refresh provider against a 1970 expiry; it now fails there with
  `vendedCredentialsNotRefreshable`. That is the intended reading — nothing can schedule a renewal
  against an expiry in 1970 — but it is a behavior change, not a no-op.

  **Follow-on in #476:** a *positive but already-past* expiry was still accepted by
  `requireUsableCredentials`, which only tested for null. That is pre-existing rather than something
  this PR introduces — main's `IcebergRestCatalogClient` parsed the field with a bare
  `Instant.ofEpochMilli(Long.parseLong(...))`, no freshness check either — but it interacts badly
  with the refresh registry: `computeRefreshSkew` returns `Duration.ZERO` for a past expiry and
  `shouldRefresh` is then true on every `resolveCredentials()` call, so a catalog that keeps
  returning the same stale timestamp is re-vended once per credential resolution and still hands
  back credentials S3 will refuse. #476 now treats an already-expired value as missing on the
  reconcile path, which routes it to the same terminal `vendedCredentialsNotRefreshable`. The query
  path is deliberately unchanged: it never renews, so a stale expiry there fails at S3 either way.
  The discovery path is also unchanged — that would need the plumbing declined above.

- **`VendedStorageCredentials` loses its two-argument constructor, and that is safe because the
  stack lands together.** Adding `scopePrefix` as a record component changes the canonical
  descriptor, and the `(Map, Instant)` overload was removed rather than kept: with the prefix in the
  middle, an overload that omits it is an ordering trap rather than a shortcut. A provider compiled
  against the previous SPI and loaded without recompiling would get a `NoSuchMethodError`.

  There is no such provider. Both call sites were inside this repository —
  `StorageAuthorityServiceImplTest` and `SourceCatalogCredentialVendorTest` — and both are updated
  in this stack, which merges as a unit rather than shipping the SPI ahead of its consumers. If an
  out-of-tree connector provider ever exists, this is the kind of change that needs a deprecation
  cycle instead.

- **A whitespace-padded vend key is rejected with a message that names no difference.** The guard
  enters on `separatorsFolded(key.trim())` matching the option while `key.equals(option)` is false,
  so `" databricks.access-delegation"` reaches the throw and the message interpolates `key.trim()`
  and the canonical spelling — byte-identical strings. Rejecting is right: `declaresVendedCredentials`
  does an exact `options().get(...)`, so a padded key would silently never opt in. What is missing is
  a message naming the whitespace. Left alone deliberately: it is message quality on a path that
  already rejects, the operator's own config shows the padding, and splitting the case needs a second
  branch and a second generated message key. `ConnectorsImplPropertyValidationTest.aMisspelledKeyIsNamedWithoutItsSurroundingWhitespace`
  uses `databricks.access_delegation`, which is also separator-misspelled, so the pure-whitespace
  shape is untested — a gap in the test rather than in the code.

- **`VendedStorageCredentialsParityTest` points `catalog-access-spi` at `connectors-spi`.** That test
  is what keeps the two ~50-line copies of the prefix-redaction rules from drifting, and the
  test-scope edge it needs runs against the layering: `core/catalog-access-spi` is documented as the
  module with no floecat dependencies, and this puts `floecat-proto` and `floecat-types` on its test
  classpath. The same guard works with the edge reversed — host the test in `core/connectors/spi`
  with a test-scope dependency on `floecat-catalog-access-spi`, which is the direction the layering
  already runs and can never become a cycle. Not done here because it is a move of a working test
  with no correctness gain. Worth taking the day anything in `connectors-spi` needs
  `catalog-access-spi` — `FloecatConnector.VendedStorageCredentials`'s own javadoc contemplates
  reconciling the two records — because that day the build fails with a Maven cycle rather than a
  compile error someone can reason about.

## Additional Notes

Not covered: `delta.source=glue`, `delta.source=filesystem`, or Azure/GCP/R2 credential consumption. Unsupported cloud credential shapes continue to fall back to storage authorities. The shared Iceberg expiry behavior changes slightly: non-positive epoch values now mean no expiry.


mcusack-floe (approved):
No blockers that I can see -- same comment as prior PR in the stack to note that smoke test is disabled for the very thing this PR implements.

PR #480, opened 2026-08-29
Shard account deletion fences to eliminate write contention

mcusack-floe:
## Summary

This removes the account-deletion fence hotspot while preserving deletion safety.

- Replace the single per-account deletion fence with 64 independently partitioned fence shards.
- Atomically install, repair, and clear all shards during account deletion.
- Route writers through a deterministic shard and remove deletion-marker pre-reads from successful write paths.
- Preserve marker reads only for classifying failed writes as account deletion conflicts.
- Add coverage for shard distribution, DynamoDB routing, fence lifecycle, and affected writer paths.

## Type of Change

- [X] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java
+++ b/service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java
@@ -395,12 +395,11 @@ public class GenericResourceRepository<T, K extends ResourceKey> extends BaseRes
         () -> {
           K resourceKey = schema.keyFromValue.apply(value);
           String accountDeletionMarker = Keys.accountDeletionMarker(resourceKey.accountId());
-          if (mutationPointerStore.get(accountDeletionMarker).isPresent()) {
-            throw accountDeletionInProgress(resourceKey);
-          }
           PreparedCreate prepared = prepareCreate(value);
           Set<String> effectiveRequiredAbsent = new HashSet<>(conditions.requiredAbsent());
-          effectiveRequiredAbsent.add(accountDeletionMarker);
+          effectiveRequiredAbsent.add(
+              Keys.accountDeletionFenceShard(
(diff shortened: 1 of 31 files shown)


service/src/main/java/ai/floedb/floecat/service/repo/util/GenericResourceRepository.java
```diff
@@ -1001,8 +993,10 @@ public Optional<MutationMeta> updateWithMetaWhilePointersMatchAndBumpMarkers(
           List<PointerStore.CasOp> ops = new ArrayList<>(prepared.ops);
           Pointer committedCanonical = prepared.committedCanonical;
           addPointerConditions(requiredPointerVersions, requiredAbsentPointers, batchedKeys, ops);
-          if (batchedKeys.add(accountDeletionMarker)) {
-            ops.add(new PointerStore.CasCheckAbsent(accountDeletionMarker));
+          String accountDeletionFence =
+              Keys.accountDeletionFenceShard(key.accountId(), canonicalPointer);
+          if (batchedKeys.add(accountDeletionFence)) {
+            ops.add(new PointerStore.CasCheckAbsent(accountDeletionFence));
```
jchauzi-floe, line 1022: Does this mean we are doing the check after the other ops? Is there some cleanup we are missing if the check fails?

mcusack-floe: Yes, there is a hole here that will leave any staged blobs orphaned, and because the account is deleted, GC has no way of cleaning that up. Now we do a best effort cleanup of a blob if the pointer transaction failed and the reason why was because of account deletion.

jchauzi-floe (approved):

PR #485, opened 2026-08-30
fix(idempotency): commit receipts atomically with keyed mutations

mcusack-floe:
## Summary

Make every Floecat operation carrying an idempotency key commit its business pointer mutation and exact success receipt atomically, or use a named mechanism whose effects are independently repeatable.

This fixes the reconcile failure where a DynamoDB transaction conflict aborted a constraints mutation but left its separately committed `PENDING` claim behind. Retries then observed `PENDING`, skipped the callback, and remained blocked for the full idempotency TTL.

The PR:

- propagates a confirmed transaction abort from DynamoDB instead of flattening it into an ordinary failed CAS;
- adds the update-shaped `runOnceCommitted` protocol and repository completion hooks;
- migrates every keyed create and update onto atomic receipt helpers;
- gives receipt-only, planner enqueue, and connector credential effects explicit repeatable mechanisms;
- removes the generic non-atomic `IdempotencyGuard.runOnce`, `MutationOps.create`, and `MutationOps.createProto` entry points;
- records derived constraints and statistics root-publication failures for durable `RootResyncQueue` re-drive while preserving their acknowledgement semantics.

## Type of Change

- [x] fix (bug fix)

## Testing

- [ ] `make fmt`
- [ ] `make verify`

- [ ] No breaking changes

The removed non-atomic helpers are an intentional internal source-compatibility break: new keyed operations must choose an atomic or explicitly repeatable mechanism. There are no RPC, persisted-format, configuration, or environment-variable changes.

## Checklist

## Additional Notes

Reviewer map:

1. `7e90abd3d` preserves the distinction between stale preconditions and confirmed transaction aborts.
2. `b6217d003` adds the atomic protocols, migrates all keyed operations, removes the unsafe APIs, and covers derived root convergence.

The regression deliberately injects a transaction conflict into parallel constraints reconcile and verifies that neither business state nor receipt commits, the idempotency key is not poisoned, retry succeeds, and replay returns the exact recorded `changed` response.

Known follow-ups, intentionally outside this PR:

- `CurrentSnapshotPointerService` still publishes a stats generation synchronously only when currency reports `UPDATED`; after a failed first attempt, an `UNCHANGED` retry may rely on the durable marker and GC cadence. Removing that guard is a separate measurable hot-path tradeoff.
- Query-side validation of `constraints_ref_version` remains a separate consistency task.

Successful stats re-drive markers may remain briefly after an immediate caller retry succeeds. They are keyed per table and idempotent, so queue depth can reflect contention as well as persistent divergence.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,7 +108,7 @@ jobs:
     if: needs.detect-core-changes.outputs.core_changed == 'true'
     name: ci-localstack
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 40
     concurrency:
       group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}-localstack
       cancel-in-progress: false
--- a/core/storage-spi/src/main/java/ai/floedb/floecat/storage/errors/StorageTransactionConflictException.java
+++ b/core/storage-spi/src/main/java/ai/floedb/floecat/storage/errors/StorageTransactionConflictException.java
@@ -16,12 +16,9 @@
 
 package ai.floedb.floecat.storage.errors;
 
-public class StorageAbortRetryableException extends StorageException {
-  public StorageAbortRetryableException(String message) {
-    super(message);
-  }
-
-  public StorageAbortRetryableException(String message, Throwable cause) {
+/** A storage transaction was definitively cancelled without applying any of its mutations. */
+public final class StorageTransactionConflictException extends StorageAbortRetryableException {
+  public StorageTransactionConflictException(String message, Throwable cause) {
     super(message, cause);
   }
 }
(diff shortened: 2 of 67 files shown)


mrouvroy-floe:
I think we have one leak still : 
`createWithMeta` and `updateWithMetaWhilePointersMatchAndBumpMarkers` invoke `completionFactory.apply(...)` (which calls `prepareSuccessOps` and writes a success blob)  before the companion `compareAndSetBatch`. 
When that batch returns false, throws StorageTransactionConflictException, or otherwise fails, neither path calls discardPreparedSuccess, orphaning the blob.

We already have the fix in the code with `discardPreparedSuccess` so I think we just need to add a `discardPreparedSuccess` call on every non-commit exit path in both `createInternal` and `updateWithMetaWhilePointersMatchAndBumpMarkers`.

mrouvroy-floe (approved):

PR #489, opened 2026-08-31
fix: fence namespace and catalog shape changes so concurrent writers cannot strand rows

mrouvroy-floe:
## Summary

Two key families in the KV schema are derived from mutable data: `namespaces/by-path`
(catalog id + ancestors' names + own name) and relation by-name keys (catalog id +
namespace id + name). A repository recomputes secondaries only for the row it writes, so
renaming `sales` leaves `sales/eu` keyed under a parent that no longer exists.

Worse, any write depending on another resource's shape reads it and writes in two round
trips. `DeleteNamespace` can count zero relations while `CreateTable` is landing one.
Both return OK; the table is orphaned.

This folds the condition into the writer's own CAS batch. Namespaces and catalogs carry
markers — version counters for their child and relation sets — and a writer asserts the
ones it depends on:

- **marker version**: require-then-advance, which excludes a writer racing the check
- **pointer version**: require-only, which refuses a write against a row whose delete
  already finished (a marker can't catch that — a later write matches the post-delete
  version)
- **deletes remove** their markers rather than advancing them

Two things worth checking in review:

1. **Participation is all-or-nothing.** Batches sharing no key can't lose to each other,
   so one non-participating writer voids the guard rather than weakening it. Requests,
   the overlay reconciler, bootstrap seeding and the transaction commit path are all in.
2. **Every fence is sampled before the check it guards.** Sampling after reads the
   version the racing writer already moved, so the CAS confirms that writer. Silent when
   wrong — it compiles, the CAS succeeds, the guard does nothing.

What can't be fenced at bounded cost is refused instead: a namespace with any descendant
can't be renamed, re-parented, or moved to another catalog.

Relation deletes assert nothing, deliberately. A racing namespace delete can only find
the namespace emptier than it counted, so nothing is orphaned — asserting would put every
`DeleteTable` on a hot key for no exclusion.

## Type of Change

- [x] fix (bug fix)

## Testing
- [x] `make fmt` 
- [x] `make verify`

- [x] No breaking changes

No API or schema changes, and no new config. Two behaviour changes worth knowing about:

- **Renaming, re-parenting or moving a namespace that has descendants is now refused**
  (`MC_CONFLICT`) where it previously succeeded and silently stranded them. Clients
  relying on the old behaviour were corrupting data, but the error is new.
- **Prepare accepts larger batches of table deletes than on the previous revision of
  this branch.** The per-intent op estimate is now charged per intent *shape*; charging
  every shape the worst case had cost clients roughly half their batch size.

Markers are created lazily on first write, so no migration or backfill is required.

## Checklist

## Additional Notes

**Known follow up:**
- The relations marker is one key per namespace, so sustained concurrent writes into a
  single namespace still contend. The commit path retries locally so this doesn't reach
  clients. Sharding it is the real fix, and should be done as a follow up, following pattern from #480 .

**Follow-up this unblocks.** This is a prerequisite for the caching effort — a cache
keyed on derived keys needs those keys not to go stale silently.


service/src/main/java/ai/floedb/floecat/service/transaction/impl/TransactionIntentApplierSupport.java
```diff
@@ -929,50 +946,120 @@ private ApplyOutcome buildRelationClaimUpsertOp(
...
+   * guarantee from the advance already queued, and two CAS ops on one key are rejected outright.
+   */
+  private ApplyOutcome addNamespaceJoin(NamespaceJoin join, PointerBatch batch) {
+    if (join.namespaceGone()) {
+      return ApplyOutcome.conflict(
+          "NAMESPACE_NOT_FOUND", "namespace no longer exists", null, null, null);
+    }
+    if (join.conditions() != null) {
+      for (PointerStore.CasOp condition : join.conditions().toCasOps()) {
+        if (!batch.addIdentical(condition)) {
```
mcusack-floe: Each table intent looks up the namespace's relation marker version separately. If another write bumps that marker in between two lookups, the two intents produce different guard conditions on the same key, addIdentical rejects them, and we fail with POINTER_TXN_DUPLICATE_KEY.

That error is treated as terminal, not retryable; a transaction touching two tables in the same namespace can fail for good just because something else committed at the wrong moment. The javadoc at TransactionsServiceImpl:2089 says this case should land in TS_APPLY_FAILED_RETRYABLE.

Either read the marker version once per namespace and reuse it for all intents in the batch, or treat a version mismatch here as retryable rather than a duplicate-key conflict.

mcusack-floe (approved):

service/src/main/java/ai/floedb/floecat/service/transaction/impl/TransactionIntentApplierSupport.java
mrouvroy-floe: Good catch — fixed in `16a7aeabc`.

The batch now retains the first guard sampled for a key. If the marker changes during planning, the final CAS fails through the existing retryable path. Guard/mutation collisions still return `POINTER_TXN_DUPLICATE_KEY`, since those represent genuinely invalid transactions.

I added a deterministic regression test that advances the marker between two intents joining the same namespace.

PR #490, opened 2026-09-01
feat: expose metadata store read observability and gate request costs

mrouvroy-floe:
## Summary

Adds production observability for physical metadata-store reads, together with integration tests that establish and gate the current storage cost of warm requests and `information_schema` scans.

The deployed `PointerStore` and `BlobStore` are now observed at their shared interfaces. This makes the instrumentation backend-independent and ensures it measures the same paths used by the running service.

The implementation reuses the existing `floecat.core.store.*` telemetry family and adds the missing distinction between requests and addressed items:

- `floecat.core.store.requests`: physical calls; one batch is one request
- `floecat.core.store.items`: keys or objects addressed; a batch of eight is eight items
- `floecat.core.store.bytes`: bytes returned when the operation has a byte unit
- `floecat.core.store.latency`: operation latency
- `floecat.core.store.errors`: failed operations

Metrics use low-cardinality `component`, `operation`, and `result` tags. Components distinguish `pointer_store`, `blob_store`, and the existing higher-level `repository` view.

Request-local operation summaries also expose call and item counts, such as `store_pointer_get_calls` and `store_pointer_get_items`, when an operation summary is active.

## Store-cost regression tests

The existing cost-test results remain the compatibility baseline:

| Operation | Cost |
| --- | --- |
| Warm request over `n` tables | `8n + 1` KV round trips, `7n + 1` S3 GETs, `3n` S3 HEADs |
| `information_schema` scan | 1 KV round trip and 1 S3 GET, flat at 1 and 8 tables |

Everything scaling with `n` in the warm-request row is a candidate for caching. The tests make those costs explicit so future work can demonstrate savings and prevent accidental regressions.

The `information_schema` result is also intentional: the scan resolves the account/catalog data it needs without introducing per-table storage reads. Any future scaling there will fail the test.

Round trips, addressed items, GETs, HEADs, listings, and prefix scans remain separate because they represent different costs. For example, eight individual gets and one batch of eight address the same number of items but require different numbers of round trips.

## DRY instrumentation

Production telemetry and test diagnostics now share one store-read observation seam:

- `ObservedPointerStore` and `ObservedBlobStore` decorate the deployed stores.
- `StoreReadObserver` defines the common operation lifecycle.
- `TelemetryStoreReadObserver` exports production telemetry.
- `RecordingStoreReadObserver` supplies the richer local diagnostics used by the cost tests.
- `StoreCostMeter` retains the existing measurement protocol, assertions, and reports.

## Cardinality and data safety

Pointer keys, blob URIs, prefixes, and stack traces are deliberately excluded from production metric tags and trace attributes.

Those details are captured only by the local test recorder, where they are useful for identifying unexpected reads without creating high-cardinality or sensitive production telemetry.

## Type of Change

- [ ] fix (bug fix)
- [x] docs (documentation)

## Testing

- [x] `make fmt`
- [x] Full `make verify`
- [x] Added test

- [x] No breaking changes
- [ ] Breaking changes
- [ ] Config/env var changes


service/src/main/java/ai/floedb/floecat/service/storage/TelemetryStoreReadObserver.java
```diff
@@ -0,0 +1,84 @@
...
+        succeed(OptionalLong.empty());
+      }
+
+      @Override
+      public void success(long bytes) {
+        succeed(OptionalLong.of(bytes));
+      }
+
+      @Override
+      public void failure(Throwable failure) {
```
mcusack-floe, line 54: On failure we are recording a raw failure message? The PR suggests that things that could be in spans like object keys are excluded. These could be in S3 exceptions passed through here.

service/src/test/java/ai/floedb/floecat/service/testsupport/RecordingStoreReadObserver.java
```diff
@@ -0,0 +1,262 @@
...
+        pointerThreads.record();
+        pointerScans.incrementAndGet();
+      }
+      case COUNT_PREFIX, COUNT_PREFIX_CONSISTENT -> {
+        pointerThreads.record();
+        pointerCounts.incrementAndGet();
+      }
+      // InMemoryPointerStore.isEmpty was never overridden by the old recorder. Production still
+      // observes it, but adding it to this test unit would change the established cost formulas.
+      case IS_EMPTY -> {}
```
mcusack-floe: Should we remove this legacy compatibility translation?

service/src/main/java/ai/floedb/floecat/service/storage/TelemetryStoreReadObserver.java
mrouvroy-floe: Fixed. We only keep exception.type in the span to avoid leaking data

service/src/test/java/ai/floedb/floecat/service/testsupport/RecordingStoreReadObserver.java
mrouvroy-floe: Done

mcusack-floe (approved):

PR #491, opened 2026-09-01
feat: vend Iceberg integration storage credentials

kwesterfeld-floe:
## Summary

Allow tables materialized through an Iceberg REST Catalog Integration/Overlay to obtain table-scoped storage credentials from their source catalog when no storage authority matches.

The shared source-catalog vendor now resolves either a Catalog Integration or a legacy Connector, opens integrations through the catalog-access SPI, preserves tenant and requested-location scoping, validates credential completeness and expiry, and maps deterministic vending failures to a structured terminal reason.

## Type of Change

- [ ] fix (bug fix)

## Testing

Validated with:

`mvn -q -pl service -am -Dtest=SourceCatalogCredentialVendorTest,ReconcileFailureClassifierTest -Dsurefire.failIfNoSpecifiedTests=false test`

- [ ] `make fmt`
- [ ] `make verify`

Formatting was applied with `mvn -q fmt:format`. The affected modules compile and the focused source-vending and failure-classification tests pass.

- [x] No breaking changes

Legacy Connector vending remains available and its storage-authority fall-back is unchanged. One connector-path behaviour does change: an already-past expiry on a reconcile vend used to be folded into the missing-fields check and raised terminally, and is now retryable on both paths — see the Reviewers note below for why, and #501 for the loop that motivated the old answer. Catalog Integration vending is attempted only after no storage authority matches, and an Integration that cannot vend now fails the read naming the cause rather than falling back.

## Checklist

## Additional Notes

- Catalog Integration/Overlay tables require no separate vending opt-in; the Iceberg REST provider requests delegated credentials and reports its capability.
- The returned credential prefix remains the already-authorized requested location rather than the catalog's potentially broader table scope.
- Ambient/default-chain and assume-role Catalog Integration authentication remain an existing follow-up; explicit SigV4 credentials are supported today.

## Reviewers

- This ports runtime Iceberg credential vending to the Catalog Integration/Overlay identity without depending on a reconstructed legacy Connector.
- Deterministically unusable vending responses carry a structured terminal reason so reconcile jobs do not retry indefinitely. An *expired* response is deliberately not one of them: a past expiry is temporal, the next vend mints a new credential, and the reconciler's attempt budget bounds the retrying. Terminal is reserved for what a retry cannot change — a tuple missing a field the catalog never sends.

  **This is a change from the base branch for the connector path, and it was made twice.** `requireUsableCredentials` previously folded an already-past expiry on `RECONCILE` into `missing` and raised `vendedCredentialsNotRefreshable`, citing an unbounded re-vend loop as the reason. That loop is real, but it does not reach the vend this classification governs: the swallow-and-retry lives in `RefreshingAwsCredentialsProviderRegistry.Entry.resolveCredentials` and only exists once a refresh provider is registered. The **first** vend has no snapshot behind it, no evidence that re-vending returns the same timestamp, and its failure *is* counted by the reconciler's attempt budget — while a terminal answer there sets the job to `JS_FAILED` and cancels its children on a single stale or cached response, and on any clock disagreement past the 60s tolerance would do that to every table at once. So the classification is retryable and the loop is tracked separately as #501, which bounds it for every reason rather than only this one. Making the refresh terminal instead was considered and rejected: `SOURCE_CATALOG_VEND_UNAVAILABLE` also carries `UNAVAILABLE`, `TIMEOUT` and `CREDENTIAL_UNAVAILABLE`, so it would make a brief catalog outage mid-capture fatal.

- **A partial vend is an error on this path, never a fall-back.** If the upstream answers with key material but not a usable pair, the provider raises `INVALID_CONFIGURATION` rather than returning "did not vend". There is no opt-in that lets a catalog integration fall back to a storage authority, so returning empty would surface a missing-authority error that reads as a configuration gap and hides the real cause — and a retry cannot change a half tuple anyway. The connector path still falls back in the equivalent situation; that is left alone deliberately, since the connector architecture is being retired in favour of integrations and overlays, and this PR only changes the integration path.

- **A token-less vend is refused rather than treated as a long-lived credential.** An integration vends only when the credential it holds is itself a temporary session, and an AWS temporary credential always carries a session token — so a pair arriving without one means the integration is not holding what it should. Accepting it would also publish a credential floecat cannot bound, which `StorageAuthorityResolver` refuses to do for its own client vends, and it would travel into worker payloads and `loadTable` responses where a consumer cannot distinguish an absent expiry from "never expires". The connector path accepts this shape on purpose, because Unity really does vend long-lived keys for some external locations; the two are separated by `VendSource`.

- **The integration vend path opens a fresh catalog client per call, with no cache.** Each call does a credential-store resolve, a REST config round trip, an OAuth exchange, and a `loadTable`, then discards the client — and vending runs once per file group, so a large capture mints a token per group against a rate-limited endpoint. Known and not addressed here: the connector path carries the same NOTE, and the cache wants a design (keying on integration and table, bounded by the shortest credential expiry) rather than being bolted onto this change.

- **`catalog-integration.use` in `READ_PERMS` is deliberately wider than the vend it admits.** The permission gates the credential vend, but `CatalogIntegrationsImpl` also requires it for `ValidateCatalogIntegration`, `ListUpstreamNamespaces` and `ListUpstreamObjects` — none of which are gated on table authorization — so granting it to the default reader role also admits upstream enumeration of namespaces and tables that were never materialized into Floecat. That is understood and accepted: the product owner is fine with `use` being broader than we will eventually want, on the basis that it is shorthand for what should later split into `catalog-integration.vend` and `catalog-integration.enumerate`. Until that split exists, the alternative is leaving overlay-materialized Iceberg tables unreadable by every role except administrator, which defeats the read-back path this PR delivers.

- **A client vend inherits whatever scope the upstream catalog minted.** `StorageAuthorityServiceImpl` reaches the source-catalog vend for `SCU_CLIENT` as well as `SCU_SERVER` (only `CredentialScope.forLeasedFile` opts out, and it says why), and the prefix stamped on the response is floecat's authorization boundary, not a restriction on the credential itself. A catalog fronting a warehouse-wide role with no session policy would therefore hand a client more reach than the table it asked for. Two things bound this. The vend refuses anything but the full session triad for an integration, so the credential is always short-lived and the exposure is one expiry wide. And a warehouse-wide role with no session policy is not a production configuration form — it cannot be run as a stable setup precisely because the token it mints is always expiring, so what the scenario actually describes is pointing floecat at a catalog to see what it does. Left as is deliberately: scoping down would mean an STS call using a credential we cannot assume is allowed to mint a narrower one, and restricting source-catalog vends to `SCU_SERVER` would remove client-side reads of overlay-materialized tables, which is a product decision rather than a fix.

- **Credential selection is by table root, not by the location the caller asked for.** `CatalogClient.vendStorageCredentials` takes only a `CatalogObjectName`, so a provider cannot know which location the caller wants and `IcebergRestCatalogClient` can only pick the longest credential covering `table.location()`. A catalog that returns one credential per prefix — the case Iceberg's own `credentials()` list and `S3FileIO`'s prefix-keyed clients exist for — therefore loses the tuple covering a file outside the table root, as `write.data.path`, `write.metadata.path` and imported files all produce. `responsePrefix` cannot compensate; it sets the advertised prefix, not the key material. Not fixed here on purpose: the fix is an SPI signature change, and the Unity provider in the branch stacked above this one implements the same interface, so it is held until that lands rather than breaking it mid-flight. The fuller answer — return every usable credential and let the consumer pick per path, which `storage_credentials` being `repeated` already allows — needs more than a signature: `VendedStorageCredentials` carries a single tuple, and the query path flattens the response into one FileIO property map in `mergeStorageAuthorityFileIoConfig`.

- **A lost call context on the vend path is terminal, not retryable.** The vend now requires `catalog-integration.use`, and `Authorizer.denied` deliberately reports the exact default `PrincipalContext` as `UNAUTHENTICATED` rather than `PERMISSION_DENIED`, so that a context-propagation miss stays diagnosable (eng-floe/floecat#361). `ReconcileFailureClassifier` treats a raw `UNAUTHENTICATED` as terminal, so a propagation miss during reconcile permanently fails the table rather than retrying. Not addressed here: both halves are shared behaviour outside this path — the classifier's rule covers every auth failure, and the `Authorizer` mapping is what makes #361-class bugs findable — so narrowing either belongs with them, not with the vend. The permission check is ordered after the missing-record and non-vending-type refusals, which is as far as this PR reduces the window.

- **A Catalog Integration that vends a bare key pair is refused at the provider.** `IcebergRestCatalogClient` now raises `INVALID_CONFIGURATION` when the selected credential has no session token or no parseable expiry, rather than returning it for the vend to reject later. An integration vends only when what it holds is itself a temporary session, and floecat does not scope a credential down with STS, so a pair that arrives is a pair that would travel to a reader. Raising it in the provider is what makes validation and the vend agree: both reach the catalog through this method, but only the vend applied the rule, so a catalog vending a long-lived pair passed every validation check — including the storage-access probe, which reads perfectly well with one — and then failed terminally on the first capture or query of every table an overlay had materialized from it. `requireUsableCredentials` keeps the same rule as a backstop, since it also covers the legacy Connector path.

- **Falling back to a storage authority is opt-in, and only a Connector can opt in.** A legacy Connector declares that it wants vended credentials (`IcebergAccessDelegation.declaresVendedCredentials`), so one that does not is answered with `null` and the caller uses the authority the operator configured for it — unchanged by this PR. A Catalog Integration has no such switch and no alternative to switch to: nothing on the record names an authority, and `ValidateCatalogIntegration` reports an Integration whose provider cannot vend as invalid rather than as configured some other way. Pairing an authority with an Integration is the split-brain this feature removes — authenticate to the catalog here, obtain storage credentials somewhere else — so every "cannot vend" condition on that path is a refusal naming the cause.

  Nothing was lost by dropping it. The vend is reached only once no authority covers the location, so a `null` from the Integration path landed on `StorageAuthorityResolver.buildResponse(null, ...)`, which raises no-matching-authority; the one consumer that absorbs that error gates on a `ConnectorConfig`, which an Integration does not have. The fall-back could only ever relabel a specific cause as "you configured no storage authority", which is neither true nor actionable. One case is deliberately not a refusal: a vended scope disjoint from the requested location is still stamped with the authorized prefix and logged, because the read may succeed and the object store enforces the real grant either way.

- **The Integration vend has no end-to-end coverage, and adding it is larger than it looks.** Every assertion on this path is a unit test against a stubbed `CatalogClient`, `CatalogIntegrationRepository` and `Authorizer`. The Polaris Integration/overlay smoke section stops at `integration validate`, `overlay reconcile` and `describe table` with the storage authority still in place, so a read there resolves through the authority and never reaches `vendForTable`. The delegated *connector* scenario does it correctly — `run_upstream_iceberg_scenario "polaris-vended-creds"` deletes the authority first so a pass cannot come from it — but nothing does the equivalent for an Integration. This matters most for the permission check: the vend requires `catalog-integration.use` read from the ambient call context, and a propagation miss on the `ScanBundleService` → `ServerSideFileIoPropertiesResolver` path would fail every read of every overlay-materialized table as `UNAUTHENTICATED`, terminally, and no unit test can catch it because each injects the principal directly.

  Mirroring the connector scenario is not sufficient, which is why this is a note rather than a change. There is no CLI verb that captures an Integration-materialized table — `overlay reconcile` is metadata-only and capture requires `connector trigger` — so `assert_table_stats_available` has nothing to assert on. The remaining route is a gateway `loadTable` against the overlay table with the authority dropped, which exercises the client vend, but the harness's only gateway call is hardcoded to one catalog and namespace and would need new plumbing plus assumptions about delegation headers and response shape that cannot be validated without running the compose stack. `tools/compose-smoke.sh` is also owned and extended by the branch stacked above this one, which already builds vend-path smoke for Unity — that is the natural place to land it.

- **Floecat cannot yet consume its own vended credentials through an Integration (#502).** The Iceberg REST gateway reports a vended credential's expiry as a top-level `expires-at` on the credential object, but Iceberg's `CredentialParser` reads only `prefix` and `config` and discards the rest, and the config map carries `s3.session-token` without `s3.session-token-expires-at-ms`. So an Integration pointed at another floecat endpoint receives a complete tuple with no parseable expiry — the one shape `missingSessionFields` refuses, since a session token floecat cannot renew is the case that lapses mid-capture. Inert for every other client, because nothing else consumes an expiry from a `StorageCredential`. Pre-existing from #287 and outside this PR's files; this PR only makes it reachable, by letting an Integration consume an upstream Iceberg REST catalog's vended credentials at all. The fix belongs in the gateway: write the expiry into the credential config under Iceberg's own `S3FileIOProperties.SESSION_TOKEN_EXPIRES_AT_MS`, which is where Polaris and Unity put it and where this PR's reader looks.
- **Rolling deployment: these reasons stay matchable against a pre-upgrade node, deliberately.** `LocalizeErrorsInterceptor` is a `@GlobalInterceptor` that rebuilds every outgoing status with `clearDetails().addDetails(the localized Error)`, so the `ErrorInfo` detail never survives an RPC and the floecat `Error` detail's params are the only discriminator a remote consumer — the reconciler included — ever sees. This change adds a `domain` param to those params so a reason cannot be impersonated by another subsystem's `params["reason"]`, which matters because `isSourceCatalogVendRefused` no longer narrows by status code. Requiring that param unconditionally would have made a new reconciler read an older service node's answers as "not this condition" for the length of a deploy: `isNoMatchingStorageAuthority` would stop being absorbed, so existing Connectors would lose their delegation fall-back, and terminal vending failures would retry until the attempt budget was spent.

  So `legacyErrorCode` accepts a domain-less detail under the `ErrorCode` its reason was historically raised with, and returns empty for anything newer — a reason introduced here has no legacy shape and must not fall into that branch. Which reasons qualify is a function of the stack's merge order rather than of `main` alone: `NO_MATCHING_STORAGE_AUTHORITY` and `VENDED_CREDENTIALS_NOT_REFRESHABLE` are in `main`, `SOURCE_CATALOG_VEND_REFUSED` is in the branch this one is stacked on and ships first, and `SOURCE_CATALOG_VEND_UNAVAILABLE` arrives here. The residual trade is worth stating: a domain-less detail carrying a qualifying reason and its historical `ErrorCode` is byte-identical to what an old node emits, so impersonation is only fully closed once this branch is deleted. It is safe to delete as soon as no deployed node emits the pre-domain shape, and nothing else depends on it.

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/storage/impl/SourceCatalogCredentialVendor.java
+++ b/service/src/main/java/ai/floedb/floecat/service/storage/impl/SourceCatalogCredentialVendor.java
@@ -16,8 +16,15 @@
 
 package ai.floedb.floecat.service.storage.impl;
 
+import ai.floedb.floecat.catalog.access.CatalogAccessException;
+import ai.floedb.floecat.catalog.access.CatalogCapability;
+import ai.floedb.floecat.catalog.access.CatalogClient;
+import ai.floedb.floecat.catalog.access.CatalogObjectName;
+import ai.floedb.floecat.catalog.access.NamespacePath;
+import ai.floedb.floecat.catalog.access.StorageLocations;
 import ai.floedb.floecat.catalog.rpc.Table;
 import ai.floedb.floecat.catalog.rpc.UpstreamRef;
+import ai.floedb.floecat.common.rpc.ErrorCode;
 import ai.floedb.floecat.common.rpc.ResourceId;
 import ai.floedb.floecat.connector.common.auth.CredentialResolverSupport;
 import ai.floedb.floecat.connector.rpc.AuthConfig;
@@ -31,20 +38,32 @@ import ai.floedb.floecat.connector.spi.FloecatConnector;
 import ai.floedb.floecat.connector.spi.LogSafeText;
 import ai.floedb.floecat.connector.spi.SourceCatalogAccessException;
 import ai.floedb.floecat.connector.spi.SourceCatalogVending;
+import ai.floedb.floecat.integration.rpc.CatalogIntegration;
+import ai.floedb.floecat.integration.rpc.CatalogIntegrationType;
 import ai.floedb.floecat.service.credentials.AuthResolutionContexts;
+import ai.floedb.floecat.service.integration.CatalogIntegrationAccess;
+import ai.floedb.floecat.service.integration.CatalogUpstreamBudget;
+import ai.floedb.floecat.service.repo.impl.CatalogIntegrationRepository;
 import ai.floedb.floecat.service.repo.impl.ConnectorRepository;
+import ai.floedb.floecat.service.security.RolePermissions;
+import ai.floedb.floecat.service.security.impl.Authorizer;
+import ai.floedb.floecat.service.security.impl.PrincipalProvider;
+import ai.floedb.floecat.storage.errors.SourceCatalogVendingGrpcStatus;
 import ai.floedb.floecat.storage.rpc.ResolveStorageAuthorityResponse;
 import ai.floedb.floecat.storage.rpc.VendedStorageCredential;
 import com.google.protobuf.util.Timestamps;
 import io.grpc.StatusRuntimeException;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
+import java.time.Clock;
+import java.time.Duration;
 import java.time.Instant;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
+import java.util.function.LongSupplier;
 import java.util.stream.Stream;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 
@@ -61,9 +80,29 @@ import org.eclipse.microprofile.config.inject.ConfigProperty;
  * what capture had just written.
  *
  * <p>Callers own admission: this class assumes the caller has already authorized access to the
- * table and decided that no storage authority covers the location. It returns {@code null} for
- * every "this catalog cannot or will not vend" condition so the caller can fall back, and throws
- * only when the catalog actively refused or answered unusably.
+ * table and decided that no storage authority covers the location.
+ *
+ * <p>The two upstream kinds answer differently, and deliberately. A legacy Connector opts in to
+ * vending -- {@code IcebergAccessDelegation.declaresVendedCredentials} is the switch -- so a
+ * connector that does not, or cannot, is answered with {@code null} and the caller falls back to
+ * the storage authority the operator configured instead. A Catalog Integration has no such switch
+ * and no such alternative: nothing on the record names an authority, and {@code
+ * ValidateCatalogIntegration} reports an Integration whose provider cannot vend as invalid rather
+ * than as configured some other way. Reaching for an authority beside an Integration is the
+ * split-brain this feature exists to remove, so every "cannot vend" condition on that path is a
+ * refusal naming the cause, not a fall-back.
+ *
+ * <p>Nothing was lost by that. This vend is reached only once no authority covers the location, so
+ * a {@code null} from the Integration path landed on {@code
+ * StorageAuthorityResolver.buildResponse(null, ...)}, which raises no-matching-authority -- the one
+ * consumer that absorbs that error gates on a {@code ConnectorConfig}. The fall-back could only
+ * ever relabel a specific cause as "you configured no storage authority".
+ *
+ * <p>What it throws is classified, because the reconcile path acts on it: only a condition a retry
+ * cannot change -- an authorization refusal, a vanished upstream table, an incomplete credential
+ * tuple -- travels as a terminal refusal reason. Anything a later attempt could clear stays
+ * retryable, since a wrongly terminal answer permanently fails a capture job while a wrongly
+ * retried one costs time.
  */
 @ApplicationScoped
 public class SourceCatalogCredentialVendor {
@@ -77,10 +116,10 @@ public class SourceCatalogCredentialVendor {
    * A catalog-supplied table identifier, flattened and bounded for a status or a log line.
    *
    * <p>Both halves come off the persisted {@code UpstreamRef}, so neither has a length or character
-   * bound. A status description becomes percent-encoded grpc-message trailer metadata that travels
-   * to every consumer and competes with HTTP/2 header limits, and {@code withReason} copies the
-   * same text into two {@code Any} details -- three carriages of whatever the catalog named its
-   * table. Flattening matters for the log line too: a name containing a newline would forge one.
+   * bound, and a refusal carries this text off the node twice: once in the {@code Error} detail's
+   * {@code detail} parameter, and again as the grpc-message {@code LocalizeErrorsInterceptor}
+   * renders from it -- percent-encoded trailer metadata competing with HTTP/2 header limits.
+   * Flattening matters for the log line too: a name containing a newline would forge one.
    */
   private static String boundedTable(String namespaceFq, String tableName) {
     return LogSafeText.bounded(namespaceFq + "." + tableName, MAX_LOGGED_PREFIX_CHARS);
@@ -116,6 +155,17 @@ public class SourceCatalogCredentialVendor {
       Stream.concat(REGION_KEYS.stream(), Stream.of("aws.region")).toList();
 
   /**
+   * How far a vended expiry may sit in the past before it stops looking like a race.
+   *
+   * <p>Wide enough to absorb the two things that legitimately produce a just-expired credential --
+   * transit time between the catalog issuing it and us reading it, and ordinary NTP-scale
+   * disagreement between floecat's clock and the catalog's -- and far narrower than any credential
+   * lifetime worth vending, so a timestamp that is genuinely stale still reads as stale. See {@link
+   * #requireLiveExpiry}, which is the only reader.
+   */
+  private static final Duration EXPIRY_SKEW_TOLERANCE = Duration.ofSeconds(60);
+
+  /**
    * What the caller will do with the credentials. Selects how strictly they are validated: only the
    * reconcile path renews them, so only it requires a renewable session tuple.
    */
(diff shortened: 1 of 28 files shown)


service/src/main/java/ai/floedb/floecat/service/storage/impl/SourceCatalogCredentialVendor.java
```diff
@@ -375,6 +487,58 @@ private static StatusRuntimeException catalogFailureStatus(
...
+            : accessFailure.code() + ": " + accessFailure.getMessage();
+    String detail =
+        String.format(
+            "source Catalog Integration %s refused credentials for %s.%s: %s",
+            integration.getResourceId().getId(), namespaceFq, tableName, safeCause);
+    if (accessFailure == null) {
+      return io.grpc.Status.INTERNAL.withDescription(detail).withCause(cause).asRuntimeException();
+    }
+    return switch (accessFailure.code()) {
+      case UNAUTHENTICATED ->
```
mrouvroy-floe, line 1147: catalogIntegrationFailureStatus maps UNAUTHENTICATED and PERMISSION_DENIED to plain gRPC statuses, while the other deterministic CatalogAccessException codes attach SourceCatalogVendingGrpcStatus.sourceCatalogVendRefused in their error details.

The reconciler currently stops retrying on raw UNAUTHENTICATED/PERMISSION_DENIED status codes, so the functional behaviour is correct today. However, diagnostics and any future consumer that matches on the structured source-catalog vending reason cannot distinguish an upstream catalog vending refusal from an unrelated service-level auth failure. Terminal retry policy is also split between raw status-code checks and structured reason codes, making future changes error-prone.

Attach the SOURCE_CATALOG_VEND_REFUSED structured status to both UNAUTHENTICATED and PERMISSION_DENIED refusals, and centralise terminal-status detection on the structured reason rather than raw gRPC codes.

service/src/main/java/ai/floedb/floecat/service/storage/impl/SourceCatalogCredentialVendor.java
```diff
@@ -171,38 +196,117 @@ ResolveStorageAuthorityResponse vendForTable(
...
-    // is missing" after planning has already succeeded. Region is resolved here so both paths see
-    // the same answer, and written under every alias the authority path emits -- consumers read
-    // different ones, and an authority-backed response has always carried all three.
-    Map<String, String> routing =
-        routingProperties(vended.get().properties(), resolvedConfig.options());
+    var credentials = vended.get();
+    // The provider uses scopePrefix to select a credential that covers the upstream table. Keep
+    // the response prefix at the already-authorized requested location: returning the catalog's
+    // potentially broader table prefix would widen the caller's service-level grant.
+    return credentialResponse(
```
mrouvroy-floe, line 455: vendFromCatalogIntegration stamps the response with responseLocationPrefix (the caller's already-authorised location) and silently ignores credentials.scopePrefix(). The response therefore advertises the credential as covering the requested prefix without verifying that the requested location actually falls within the scope the upstream catalog vended.

This is safe today only because of two implicit invariants: vending is gated to CIT_ICEBERG_REST (line 221), and the Iceberg REST provider's vendStorageCredentials self-selects a credential via coversLocation(tableLocation, prefix) and raises CREDENTIAL_SCOPE_INVALID otherwise. Neither guarantee is expressed at this call site, and the CatalogClient SPI contract does not require providers to validate coverage — so a future provider or a change to the gating condition could silently return a mis-scoped credential.

Add an explicit assertion (or a guard that raises an appropriate CatalogAccessException) that requestedLocation is covered by credentials.scopePrefix() before building the response, and document in the SPI why providers must either validate coverage themselves or surface CREDENTIAL_SCOPE_INVALID.

mrouvroy-floe (approved):
Small hardenings available but otherwise, LGTM

service/src/main/java/ai/floedb/floecat/service/storage/impl/SourceCatalogCredentialVendor.java
kwesterfeld-floe: Fixed in 618ea6ed5.

`vendFromCatalogIntegration` now checks `credentials.covers(responseLocationPrefix)` before building the response, and the SPI contract is written down: `CatalogClient.vendStorageCredentials` documents that a provider must return a scope covering the table location or raise `CREDENTIAL_SCOPE_INVALID`, and must never widen the scope it reports. The prefix comparison itself moved to `StorageLocations` in the catalog-access SPI (with `VendedStorageCredentials.covers`), which the Iceberg provider now shares instead of keeping its own `coversLocation`/`normalizeS3Scheme` — one implementation, so the two ends cannot disagree about what a credential covers.

One deviation from the suggestion: the guard falls back (returns `null`) rather than raising. An under-reaching scope is not always a provider bug — an Iceberg table can legitimately hold data outside its own table prefix — and raising `CREDENTIAL_SCOPE_INVALID` maps to a terminal refusal, which would permanently fail capture for a table that storage-authority handling could still read. The condition is logged. Happy to make it strict if you would rather it be loud.

service/src/main/java/ai/floedb/floecat/service/storage/impl/SourceCatalogCredentialVendor.java
kwesterfeld-floe: Fixed in 618ea6ed5.

`SourceCatalogVendingGrpcStatus` gained a `sourceCatalogVendRefused(code, errorCode, description, cause)` overload that keeps the status code the refusal deserves while attaching `SOURCE_CATALOG_VEND_REFUSED`, and `isSourceCatalogVendRefused` now matches on the domain-scoped reason alone rather than reason-plus-`FAILED_PRECONDITION`. Both vending paths use it — `catalogIntegrationFailureStatus` and the connector path's `NotAuthorizedException`/`ForbiddenException` mapping — so an upstream 401/403 is now distinguishable from floecat's own auth failing. The overload also keeps the originating exception as the status cause, which the existing structured helpers were dropping.

On centralising terminal detection: I left `ReconcileFailureClassifier`'s raw `UNAUTHENTICATED`/`PERMISSION_DENIED` rule in place. It covers auth failures from anywhere, floecat's own services included, so folding it into the structured check would quietly make those retryable — a wider policy change than this PR should carry. The vending path is now identified structurally, so the two no longer overlap by accident, and there is a comment at the branch saying so. New `SourceCatalogVendingGrpcStatusTest` covers the refusal being recognised under every code it can carry, a service-level auth failure *not* matching, and the cause surviving while the wire description stays safe.

PR #495, opened 2026-09-02
Add the shared memory-cache contract and telemetry

mrouvroy-floe:
## Summary

Introduces the cache foundation for the warm-read path:

- Adds `MemoryCache<K, V>` with single and batch loading, publication, eviction, and partition eviction.
- Prevents stale in-flight loads from overwriting concurrent mutations.
- Derives per-cache budgets from the container memory limit.
- Adds shared hit, miss, load, eviction, and size telemetry.
- Documents the boundaries between memory, pointer, and disk caches.

Pointer ordering and authoritative listings will be owned by the pointer layer. 
The disk blob cache will use a lifecycle-oriented interface while sharing the same telemetry conventions.

## Validation

- `make fmt`
- Full unit-test suite
- Cache contract tests
- Documentation tests

**Base:** `mr-remove-pin-validation`
## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist

## Additional Notes


core/cache/src/main/java/ai/floedb/floecat/cache/CaffeineMemoryCache.java
```diff
@@ -0,0 +1,360 @@
...
+    Set<K> distinctKeys = new LinkedHashSet<>(keys);
+    if (distinctKeys.isEmpty()) {
+      return Map.of();
+    }
+
+    Map<K, V> result = new LinkedHashMap<>();
+    Map<K, ReadStamp> misses = new LinkedHashMap<>();
+    for (K key : distinctKeys) {
+      long startNanos = System.nanoTime();
+      CachedValue<V> value = entries.getIfPresent(key);
```
mcusack-floe, line 201: An eviction can be missed here. getIfPresent() checks the cache before the fence is sampled. If evict() completes between those operations, the subsequent bulk-loaded value can be installed and retained. Sample the fence before checking the cache, as get() does.

mcusack-floe (changes requested):

core/cache/src/main/java/ai/floedb/floecat/cache/CaffeineMemoryCache.java
mcusack-floe: This also applies to evictPartition()

mrouvroy-floe: getAll() now samples the mutation fence before probing the cache, so concurrent evict() and evictPartition() cannot leave a bulk-loaded value installed after invalidation.

mcusack-floe (approved):

PR #503, opened 2026-09-07
Add assembled relation metadata object cache

mrouvroy-floe:
## Summary

Add the object-cache layer for assembled relation metadata.

## Changes

- Cache assembled relation and schema POJOs through the shared cache module.
- Cache constraints and snapshot facts using immutable identities.
- Resolve pinned schemas through cached objects before falling back to storage.
- Keep historical snapshot facts separate from current relation metadata.
- Remove duplicate metadata/memoization paths that are now covered by the object cache.
- Preserve invalidation and refresh behavior when relation metadata changes.

## Result

Warm requests reuse already assembled metadata instead of rebuilding POJOs or rereading the same relation, schema, constraint, and snapshot records.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist


core/cache/src/main/java/ai/floedb/floecat/cache/CaffeineMemoryCache.java
```diff
@@ -145,54 +150,64 @@ public void maximumBytes(long maxBytes) {
...
+      return cached.value();
+    }
+
+    // Caffeine's atomic mapping function cannot load another key from this same cache: colliding
+    // keys can throw "Recursive update". Per-key ownership keeps single-flight while running the
+    // loader outside Caffeine's map computation, so one cached object may safely be assembled from
+    // another without introducing lock ordering between unrelated keys.
+    LoadingValue<V> mine = new LoadingValue<>(Thread.currentThread());
+    LoadingValue<V> existingLoad = loads.putIfAbsent(key, mine);
+    if (existingLoad != null) {
```
mcusack-floe: A caller arriving after evict() can still join a load started before invalidation and receive stale data. Retire or version in-flight loads during eviction.

service/src/main/java/ai/floedb/floecat/service/query/catalog/RelationBundleBuilder.java
```diff
@@ -204,63 +208,33 @@ private RelationInfo buildRelation(
...
-        relation.node() instanceof ViewNode view
-            ? view.outputColumns()
-            : relation.node() instanceof UserTableNode userTable
-                ? SchemaColumns.topLevelOnly(
-                    logicalSchemaForRelation(
-                            correlationId, relation.relationId(), userTable, queryContext)
-                        .getColumnsList())
-                : Optional.ofNullable(graphView.tableSchema(relation.node().id()))
-                    .orElseGet(List::of);
+        SchemaColumns.topLevelOnly(template.schema().getColumnsList());
```
mcusack-floe: topLevelOnly drops direct-view output columns whose physical_path differs from their name. Preserve view output columns or enforce this invariant when views are written.

service/src/main/java/ai/floedb/floecat/service/statistics/StatsOrchestrator.java
```diff
@@ -182,6 +259,57 @@ public StatsResolutionResult resolveInGeneration(
...
+   * Historical pins read through so time travel cannot displace the resident current catalog.
+   */
+  public Optional<ObjectCache.SnapshotFacts> resolveTableFactsInGeneration(
+      StatsCaptureRequest request,
+      Optional<String> pinnedGenerationToken,
+      boolean retainCurrentFacts) {
+    Optional<String> pinned =
+        pinnedGenerationToken.filter(token -> !token.isBlank()).map(String::trim);
+    if (pinned.isPresent()) {
+      java.util.function.Supplier<Optional<ObjectCache.SnapshotFacts>> loadExact =
```
mcusack-floe, line 224: This direct pinned-generation read bypasses resolveSingleFromStore’s failure fallback. An unreadable frozen manifest now produces no stats instead of falling through to newest.

service/src/main/java/ai/floedb/floecat/service/statistics/StatsOrchestrator.java
```diff
@@ -182,6 +259,57 @@ public StatsResolutionResult resolveInGeneration(
...
+          .filter(TargetStatsRecord::hasTable)
+          .map(StatsOrchestrator::snapshotFacts);
+    }
+    java.util.function.Supplier<Optional<ObjectCache.SnapshotFacts>> loadLive =
+        () ->
+            resolveInGeneration(request, Optional.empty())
+                .stats()
+                .filter(TargetStatsRecord::hasTable)
+                .map(StatsOrchestrator::snapshotFacts);
+    return retainCurrentFacts
```
mcusack-floe: This caches mutable live facts under a blank identity without expiry. Since invalidation is replica-local, other replicas can serve stale facts indefinitely.

mcusack-floe (changes requested):

core/cache/src/main/java/ai/floedb/floecat/cache/CaffeineMemoryCache.java
mrouvroy-floe: Fixed. evict() and evictPartition() now retire in-flight owners, and miss registration is gated while a partition sweep snapshots and removes loads. A post-eviction caller can no longer join a retired load. getAll() also samples the fence before probing. Added regression coverage.

service/src/main/java/ai/floedb/floecat/service/query/catalog/RelationBundleBuilder.java
mrouvroy-floe: Fixed. View schemas now bypass the table-only topLevelOnly filtering, so output columns whose physical_path differs from their name are preserved in both the cached template and the emitted relation info. Added a regression test.

service/src/main/java/ai/floedb/floecat/service/statistics/StatsOrchestrator.java
mrouvroy-floe: Fixed. Pinned reads now use the shared PlannerStatsResolver.resolvePinnedFromStore(...) path. An unreadable frozen manifest follows the existing newest/live fallback, while retryable aborts are still propagated. Added test coverage.

service/src/main/java/ai/floedb/floecat/service/statistics/StatsOrchestrator.java
mrouvroy-floe: Fixed. Live facts are no longer cached under an empty generation identity. Only immutable generation-specific facts are retained in ObjectCache; live facts are read through to avoid indefinite replica-local staleness. Added a regression test.

mcusack-floe (approved):

mcusack-floe:
Is the metric floecat.core.cache.loads.discarded used anywhere? I did a search of the code and couldn't find any producers. Probably remove and update docs.

mrouvroy-floe:
> Is the metric floecat.core.cache.loads.discarded used anywhere? I did a search of the code and couldn't find any producers. Probably remove and update docs.

Removed. The fix is on #532 rather than here, because that's where the producer died: moving the immutable caches onto native Caffeine loading deleted LoadCoordinator and the retained check that raised it, and a later commit restored the CacheEvents declaration without the call site. Caffeine's atomic compute either retains the loaded value or caches nothing, so there's no discard left to count. Took floecat.core.cache.write.through with it for the same reason — its producer was PointerCache, which #532 replaces.

PR #504, opened 2026-09-07
Add NVMe cache for immutable Floecat blobs

mrouvroy-floe:
## Summary

Add the on-disk blob-cache layer for large immutable payloads.

## Changes

- Add a scoped NVMe cache API for blob reads, writes, eviction, and fallback.
- Cache immutable blob ranges and other large content-addressed payloads locally.
- Route snapshot and statistics reads through the blob cache.
- Preserve correct behavior for current snapshots and missing/reclaimed blobs.
- Add cache telemetry and store-cost coverage.
- Fall back to S3 on an NVMe miss, then populate the local cache.

## Result

Repeated warm reads of large snapshot and statistics payloads are served from local 
NVMe instead of issuing repeated S3 GETs.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist

## Additional Notes

No compression of the data on dis for now


mcusack-floe (approved):
No blockers.

core/cache/src/main/java/ai/floedb/floecat/cache/DiskBlobCache.java
```diff
@@ -0,0 +1,926 @@
...
+      } catch (IOException ignored) {
+        // Best effort: account deletion remains authoritative in the durable stores.
+      }
+    } finally {
+      state.lock.writeLock().unlock();
+    }
+  }
+
+  @Override
+  public synchronized SweepResult sweep() {
```
mcusack-floe: sweep, publish and deleteIfIdle and retire are all synchronized. fill is serialized to one blob at a time process-wide and stops entirely during each sweep.

mcusack-floe (changes requested):

mcusack-floe:
It appears that N concurrent readers of the same cold blob now issue N S3 GETs and N duplicate publishes?

core/cache/src/main/java/ai/floedb/floecat/cache/DiskBlobCache.java
mrouvroy-floe: Fixed. Entries take a striped lock now, so a fill no longer fsyncs while excluding every other fill, and sweeps don't block fills.

mrouvroy-floe:
> It appears that N concurrent readers of the same cold blob now issue N S3 GETs and N duplicate publishes?

Two different answers here:

- The duplicate publishes don't happen: publish returns early when the file exists and the read path never passes replace=true, and even without a lock the CREATE_NEW plus atomic move means the loser just discards its staging file. 

- The duplicate GETs are real and expected: every key in this cache is content-addressed, so N concurrent loads return identical bytes and the cost is bandwidth, never a wrong answer. Fixing it would require putting back in-flight merge mechanism, which Neil ruled out for the first drop as un-needed optimization.

mcusack-floe (approved):

PR #505, opened 2026-09-07
Add per-engine relation hint caching

mrouvroy-floe:
## Summary

Move engine-specific hints into their own cached resources.

## Changes

- Store hints independently from the relation blob.
- Identify hints by relation, engine kind, and engine version.
- Persist immutable hint payloads through the blob/cache path.
- Resolve and attach hints through the shared pointer and object-cache layers.
- Reuse complete hints on warm requests instead of recomputing them.
- Remove the obsolete hint-provider cache path.
- Avoid rewriting the relation resource when one engine's hints change.

## Result

Hints for different engines no longer collide, and updating one engine's hints does not 
invalidate unrelated relation metadata or cached derivatives.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist


service/src/main/java/ai/floedb/floecat/service/cache/HintCache.java
```diff
@@ -0,0 +1,501 @@
...
+        repository.getThrough(
+            node.id(),
+            engineKind,
+            engineVersion,
+            blobUri -> readBody(node.id(), engineKind, engineVersion, blobUri));
+    Hints hints =
+        stored
+            .filter(cached -> node.cacheIdentity().equals(cached.relationIdentity()))
+            .map(CachedHints::hints)
+            .orElseGet(() -> readLegacy(node, engineKind, engineVersion));
```
mcusack-floe, line 94: Do we need a legacy fallback?

mcusack-floe (approved):

service/src/main/java/ai/floedb/floecat/service/cache/HintCache.java
mrouvroy-floe: Removed it. It was the read path for the old storage format: plugins persisted hints as engine.hint.* relation properties, and this PR moves writes to the per-engine resource. Nothing writes properties any more (EngineHintMetadata.encodeValue is test-only), so it only served data written before the change. 
But that doesn't need serving: a missing hint is a cold path, the decorator will just re-derives it.

PR #506, opened 2026-09-07
Serve the accounts Core assigns, fenced in the store

mrouvroy-floe:
## Summary

Make Floecat account aware: it serves the accounts Core assigns it, and the key-value store
refuses a write from a pod that is no longer the owner.

## What Floecat does

Floecat is an assignment receiver. Core pushes it a complete account set through
`AccountAssignmentControl` — `ApplyAssignment` and `GetAssignmentStatus`, both answered, never
initiated. For an account it serves, it admits pins and mutations under a permit and answers
from its planning index. For one it does not, catalog reads fall through to the store and
`BeginQuery`, pin resolution and mutations fail with `FAILED_PRECONDITION` carrying
`floecat.not_assigned`, which Core reads as "refresh the route and restart".

## The store fence

One item per account, `/assignment-fence/<account>`. Taking an account is a compare-and-set on
it; every account-scoped write then carries that version as a condition inside the same
transaction. This is the part that actually holds the invariant: a pod that is partitioned,
paused, or already replaced cannot commit, and it finds out because its write is refused rather
than because a message reached it. A pod that gives an account up keeps the remembered version
until its in-flight writes finish, so work already admitted fails its condition instead of
committing unconditioned.

A second record, `/assignments/<member>`, lets a restarted process recover the accounts whose
fence still names it without waiting for Core.

## Changes

- Defines the assignment contract and serves the control RPC (managed mode only).
- Serves only the accounts Core assigns, with reads falling through for the rest.
- Admits pins and account mutations on the owner only.
- Fences every account-scoped write on the assignment version.
- Collects garbage only for assigned, Core-permitted accounts, revalidating the permit before
  each irreversible delete.
- Recovers the assignment from the store at startup.
- Drains in-flight work before termination, with a shutdown observer as the fallback when the
  lifecycle hook never arrives.
- Re-verifies ownership against the store on a background sweep, failing closed when the store
  cannot be reached.

## Notes

- **Ownership state is written to KV**: the fence and the member index. What is not written is
  anything on the query path — queries take no writes to maintain ownership.
- No lease and no TTL. Ownership ends on a Core push or a store-proven fence move, never on a
  timer, so Core's absence cannot fence a serving pod.
- No Floecat-to-Floecat communication, and no Floecat-to-Core call of any kind.
- Stale assignments are rejected by epoch and by process incarnation.
- Standalone stays the OSS default and owns every account; `none` serves reads only.
- The drain endpoint is registered only in managed mode and authenticates no one — the preStop
  hook is a kubelet `httpGet` from the node address, which no in-process check can distinguish
  from any other caller, so access is the mesh policy's job
  (https://github.com/eng-floe/floe-infrastructure/pull/182).

## Compatibility with FloeDB

Works with changes in:
- core: https://github.com/eng-floe/core/pull/2419
- floe-infrastructure: https://github.com/eng-floe/floe-infrastructure/pull/182
- application-base: https://github.com/eng-floe/application-base/pull/54

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [ ] No breaking changes


service/src/main/java/ai/floedb/floecat/service/account/FloecatDrainEndpoint.java
```diff
@@ -0,0 +1,147 @@
...
+public class FloecatDrainEndpoint {
+  private static final String PATH = "/internal/drain";
+  private static final long POLL_MILLIS = 25L;
+
+  @Inject AccountGcAuthority authority;
+
+  @ConfigProperty(name = "floecat.account-ownership.drain-timeout-ms", defaultValue = "180000")
+  long defaultTimeoutMs;
+
+  void routes(@Observes Router router) {
```
mcusack-floe, line 47: This is unauthenticated so isn't safe in a stand-alone env.

service/src/main/java/ai/floedb/floecat/service/query/resolver/QueryInputResolver.java
```diff
@@ -454,6 +483,73 @@ private ResolutionResult resolveInputsCore(
...
+    private ResolutionAdmission(AccountGcAuthority authority, String accountId) {
+      this.authority = authority;
+      this.accountId = accountId;
+    }
+
+    private static ResolutionAdmission unrestricted() {
+      return new ResolutionAdmission(null, "");
+    }
+
+    private synchronized void acquire() {
```
mcusack-floe, line 528: ResolutionAdmission.close() only nulls the permit; an abandoned fan-out sibling calling acquire() after close leaks a permit, so activeResolutions never reaches 0 and drain never completes.

service/src/main/java/ai/floedb/floecat/service/account/FloecatDrainEndpoint.java
```diff
@@ -0,0 +1,147 @@
...
+        .onSuccess(result -> respond(context, result))
+        .onFailure(context::fail);
+  }
+
+  private void status(RoutingContext context) {
+    respond(context, authority.processStatus());
+  }
+
+  private AccountGcAuthority.ProcessStatus awaitDrained(long timeoutMs) {
+    long deadline = System.nanoTime() + Duration.ofMillis(timeoutMs).toNanos();
```
mcusack-floe, line 96: unvalidated timeoutMs overflows the nanoTime deadline (instant "not drained") or throws ArithmeticException.

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
mcusack-floe, line 1: GcPermitRevokedException is swallowed by the chain-walk catch (RuntimeException) (also catch (Exception) ~1968), so revocation mid-mark degrades to a poisoned sweep and the pass continues under a revoked permit.

mcusack-floe (changes requested):

service/src/main/java/ai/floedb/floecat/service/account/FloecatDrainEndpoint.java
mrouvroy-floe: Standalone is covered: the route is only registered in managed mode, so in standalone the endpoint doesn't exist.
```java
void routes(@Observes Router router) {
  if (!assignment.managed()) return;
  router.get(PATH).handler(this::handleHttp);
```
In FloeDB managed mode, it is still unauthenticated, deliberately. 
I tried an in-process pod-local check and reverted it: the preStop hook is a kubelet httpGet on the pod address, so the caller arrives from the node IP with no identity an in-process check can recognise. The check 403'd the only real caller and dropped every drain onto the termination grace period. 
Instead, access is validated through the mesh policy's job, [eng-floe/floe-infrastructure#182], which DENYs /internal/drain for anything that isn't floe-control or floe-account. 
Every other service here serves the same path with no in-process check

service/src/main/java/ai/floedb/floecat/service/gc/CasBlobGc.java
mrouvroy-floe: Fixed. Rethrown ahead of both catch-alls, at the chain walk (:1753) and the artifact-index rooting (:1954), alongside DeadlineReached and CapacityExceededException. Revocation mid-mark now aborts the pass instead of poisoning the sweep.

service/src/main/java/ai/floedb/floecat/service/account/FloecatDrainEndpoint.java
mrouvroy-floe: Fixed. Parsed, with negative and malformed both returning 400, and clamped to [0, MAX_TIMEOUT_MS] (1 hour) before it reaches the deadline: Math.max(0L, Math.min(MAX_TIMEOUT_MS, timeoutMs)).

service/src/main/java/ai/floedb/floecat/service/query/resolver/QueryInputResolver.java
mrouvroy-floe: Fixed. close() now takes the monitor, sets closed, drains the map and releases outside the monitor; admit() throws CancellationException once closed, so a late fan-out sibling can't add one. activeResolutions reaches zero and the drain completes.

service/src/main/java/ai/floedb/floecat/service/account/AssignmentFence.java
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  // -----------------------------------------------------------------------------------------
+
+  private record Check(String accountId, long version) {}
+
+  private boolean fenced(List<Check> checks, List<CasOp> ops) {
+    List<CasOp> fencedOps = new ArrayList<>(ops.size() + checks.size());
+    for (Check check : checks) {
+      fencedOps.add(new CasCheck(Keys.accountAssignmentFence(check.accountId()), check.version()));
```
mcusack-floe, line 194: Looks like we're overflowing the 100 record batch size in dynamo by adding a cas check to every batch of 100?

service/src/main/java/ai/floedb/floecat/service/account/AssignmentFence.java
```diff
@@ -0,0 +1,243 @@
...
+/**
+ * The store fence: every account-scoped write by an owning process carries {@code
+ * CasCheck(assignment fence, remembered version)}, so a process that lost the account cannot
+ * commit. Sits once beneath {@code IndexedPointerStore}, leaving repositories unchanged.
+ *
+ * <p>A single-key write becomes a two-item transaction; a batch gains one check per account it
+ * writes; a prefix delete, which cannot be transactional, is preceded by one consistent read.
+ * Outside managed mode, and for accounts this process does not own, every call passes through.
+ */
+public final class AssignmentFence implements PointerStore {
```
mcusack-floe, line 42: We're not going through pointer store here, so no ownership cas check is attached?

service/src/main/java/ai/floedb/floecat/service/account/AccountAssignment.java
```diff
@@ -0,0 +1,1323 @@
...
+    processDraining = true;
+    return status();
+  }
+
+  // ---------------------------------------------------------------------------------------------
+  // Store fence
+  // ---------------------------------------------------------------------------------------------
+
+  /** Takes an account's fence and starts serving it. Runs only on {@link #background}. */
+  private void fenceAndServe(String accountId, long epoch) {
```
mcusack-floe, line 631: fenceAndServe() doesn't check processDraining. A fence take queued by an earlier apply() can bump an account's fence and move it to SERVING after the pod has irreversibly begun draining and already reported itself drained.

mcusack-floe (changes requested):

service/src/main/java/ai/floedb/floecat/service/account/AssignmentFence.java
mrouvroy-floe: Good catch. Ownership fencing adds one CAS check per affected account, so a batch that was valid at 100 operations could exceed DynamoDB’s 100-action transaction limit.

The fix now validates the final decorated batch inside AssignmentFence before delegating to DynamoDB. The known batch builders also reserve capacity for the possible ownership/deletion checks by using 98 operations instead of 100. Oversized batches fail before submission, so there is no partial commit or transaction splitting.

service/src/main/java/ai/floedb/floecat/service/account/AccountAssignment.java
mrouvroy-floe: Fixed. A queued fenceAndServe() now has to pass the process-drain admission gate before it performs the durable read/CAS. The operation remains counted as active until the whole fence acquisition finishes, and it cannot publish SERVING after draining has started.

A deterministic regression test now covers the case where an account leaves the assignment while its fence acquisition is still in flight.

service/src/main/java/ai/floedb/floecat/service/account/AssignmentFence.java
mrouvroy-floe: This path is intentional. fenceAndServe() writes the assignment-fence record itself; that record is the ownership authority and is the key used to fence all subsequent account-scoped mutations. It cannot carry a CAS check against the same key it is creating or advancing.

All ordinary account-scoped pointer mutations go through AssignmentFence and receive the ownership CAS check. The authority’s own fence acquisition is the deliberate exception.

mrouvroy-floe:
Superseded by https://github.com/eng-floe/floecat/pull/549, implementing the simplified lifecycle-drain design.

PR #408, opened 2026-07-24
fix: rearchitect durable reconciliation and snapshot artifact publication

mcusack-floe:
## Summary

This PR improves the scalability and reliability of durable reconciliation, particularly for large snapshot captures and DynamoDB-backed deployments.
- Replaces chunked gRPC result transfer with executor-uploaded immutable stats, index, and file-group payloads.
- Introduces fenced snapshot capture manifests and manifest-backed artifact resolution using whole-object reads, plus GC protection. Floecat does not perform range lookups against these manifests; the unused blob-store `getRange` API and its associated implementations and tests have been removed.
- Adds remote snapshot finalization for file-group captures while retaining local finalization for direct-stats and empty snapshots.
- Reduces durable queue contention through atomic lifecycle/index updates, empty-queue polling backoff, and streamlined lease maintenance.
- Makes canonical child state the source of truth for parent rollups, with coalesced projection maintenance for list and tree views.
- Bounds reconciliation GC using retention and quarantine indexes, stored cleanup manifests, and tighter per-tick work limits instead of recurring scans.
- Adds abandoned-stats cleanup and GC-derived per-account storage estimates without incremental accounting writes or a counter-row fallback.
- Expands reconciliation timing, storage-estimate, GC telemetry, documentation, and test coverage.
- Query-driven stats collection is disabled by default. Cold query misses do not enqueue stats capture unless operators opt in with `floecat.stats.sync.enabled=true` (`FLOECAT_STATS_SYNC_ENABLED=true`).

## Type of Change

- [ ] feat (new functionality)
- [X] fix (bug fix)
- [ ] docs (documentation only)
- [ ] refactor (no behavior change)
- [ ] test (adds/updates tests)
- [ ] chore (build/tooling/maintenance)

## Testing

Describe how this was validated.

- [X] `make fmt`
- [X] `make verify`
- [X] Added/updated tests for changed behavior

## Deployment and Compatibility

- [ ] No breaking changes
- [X] Breaking changes (described below)
- [X] Config/env var changes (described below)

### Breaking remote-executor protocol change

This PR intentionally removes the legacy chunk-streaming SubmitReconcileJobResultChunk RPC and does not retain a compatibility shim. The replacement completion flow uses executor-written immutable stats, index, and file-group artifacts, then submits digest-bound descriptors and a prepared marker through SubmitLeasedReconcileJobResult.

[...]

PR #510, opened 2026-09-07
test(smoke): prove Catalog Integration vending

kwesterfeld-floe:
## Summary

Fixes a credential-resolution bug that stopped Iceberg connectors from ever asking their catalog to vend, and gives all four credential-vending paths a smoke oracle that cannot pass without vending.

The bug and the missing coverage are the same story. `polaris-vended-creds` was meant to prove floecat can read using credentials Polaris hands out, but it removed only the authority the smoke creates and left the seeded `fixture-floecat` one covering the same location, so it passed by reading through an authority and hid the fact that vending was broken. Making the oracle honest is what surfaced the bug.

Changes:

- Keep operator header hints across a credential token exchange. Every exchange branch replaced them with an empty map alongside auth props, which removed `header.X-Iceberg-Access-Delegation` from the catalog properties the Iceberg connector builds.
- Give the Catalog Integration vend end-to-end coverage for both providers, Polaris and Unity, through a gateway `loadTable` that must return a complete session tuple.
- Make the delegated Connector oracle take away every authority covering the table location, computed with the resolver's own matching rule, and assert none still does before reading.
- Run the delegated Connector scenario in both `localstack` variants rather than only under `-remote`.

## Type of Change

- [x] fix (bug fix)

## Testing

Full unit suite locally: 2336 service tests, 0 failures. `fmt:check` clean across all modules.

Each of the four exchange branches the fix touches is pinned by its own assertion: reverting the initialiser to an empty map fails five tests, one per branch plus the secret-handling one.

**The compose smoke ran in CI and all four vend paths pass**, in both `localstack` and `localstack-remote`, with zero `[FAIL]` lines in either job:

```
[PASS] upstream iceberg Catalog Integration credential vending
[PASS] upstream delta unity Catalog Integration credential vending
[PASS] Polaris credential vending          [PASS] Unity credential vending
gateway loadTable http=200                 unity integration gateway loadTable http=200
no storage authority covers s3://floecat/sales/us/trino_types
no storage authority covers s3://floecat-delta-vended/call_center
```

Two assumptions that could not be settled by reading were settled by that run: the gateway response shape is as assumed, and the Iceberg REST gateway does serve a Delta-sourced overlay table through `loadTable`. `polaris-vended-creds`, which failed with `catalog does not delegate` before the fix, now passes with every covering authority provably disabled, which it can only do by vending.

A passing smoke job carries no container logs, since the harness dumps them only on failure. So the evidence for the connector vend is the scenario passing with the authority path removed, not a log line, which is the stronger form and the reason these assertions were moved off log scraping.

**One change is verified only by CI.** Names now reach `set_authorities_enabled` through a here-string rather than a pipe, because as a pipeline sink it runs in a subshell that inherits the ERR trap under `set -E`. That mechanism does not reproduce on bash 3.2, which is what this was developed against, so the local checks say nothing about it. Everything else was exercised offline.

- [x] No breaking changes

**This PR changes production code.** `CredentialResolverSupport` is on the shared connector auth path, and the change affects the client-credentials, RFC8693, Azure OBO and GCP DWD exchanges.

The effect is to stop discarding `auth.header_hints` after an exchange. Behaviour that depended on hints being dropped would change; nothing in the tree does, and the drop was silent rather than deliberate. Catalog Integrations are unaffected: they never call `apply()`, and the Iceberg REST provider sets the delegation header itself rather than relying on a hint surviving.

Auth props are still replaced, so the client id and secret do not travel past the exchange.

## Checklist

## Reviewers

- **Preserving header hints cannot leak a secret past the exchange, and that is enforced rather than assumed.** `ConnectorsImpl` runs `validateSecretBearingMap(auth.getHeaderHintsMap(), corr, "auth.header_hints")` on every write, so a secret-bearing key cannot be persisted there in the first place. Credential headers are a separate map (`credential.getHeadersMap()`), which the exchange consumes and does not forward. `sensitiveBootstrapSecretsAreNotPropagatedAfterExchange` now starts from a non-empty hint map, so it asserts both halves at once: the bootstrap `Authorization` does not reach the connector, and the operator hint does.

- **`headerHints` is `final`.** The seven in-switch assignments were byte-identical re-copies of the initialiser, so each branch asserted what the variable already held, which is the shape of the bug being fixed. Stating the invariant once lets the compiler keep it, and a future change that populates hints before the switch cannot be silently discarded by a branch.

- **The declaration gate and the factory read different configs, which is why this failed silently.** `IcebergAccessDelegation.declaresVendedCredentials` reads the stored record, so it reported the connector as asking for delegation; `IcebergConnectorFactory` receives the *resolved* config, whose hints had been dropped. The vend therefore ran, asked the catalog nothing, came back empty, and fell through to a storage authority, invisible wherever an authority happened to cover the table. The CI diagnostic that pinned it was `accessDelegationHeader=null headerKeys=[]`.

- **The gateway is used rather than a capture, and that is forced.** `overlay reconcile` is metadata-only and capture needs `connector trigger`, which an Overlay-materialized table has no Connector for, as `describe table` asserting `connector_id: -` establishes a few lines earlier. So `assert_table_stats_available` has nothing to assert against on this path. A gateway `loadTable` is the one surface that reaches `vendForTable` for an Integration table, and it exercises `CredentialUse.QUERY`, the path a reader takes.

- **Coverage is computed with `matchesLocationPrefix`'s own rule, on the table location, and keyed on the authority id.** A substring test is wrong in both directions: it misses `s3://floecat` covering `s3://floecat/sales/...` and wrongly counts `s3://floecat-delta` as covering `s3://floecat-delta-vended/...`. The location matters as much as the rule, since a bucket root cannot see an authority scoped beneath it, so both callers derive the table root from the fixture's registered metadata URI. Rows are identified by the enabled flag plus a UUID rather than by field count, because neither the display name's position nor the field count is fixed: a multi-word name adds fields in the middle and a blank `TYPE` collapses to four. `storage-authority update` takes an id directly, since `resolveId` short-circuits on `looksLikeUuid`.

- **Failing to parse is not the same as finding nothing.** Recognising zero rows exits non-zero rather than reading as "no authority covers this", which every caller would otherwise treat as proof of absence. A missing covering authority likewise fails rather than skipping, since the `loadTable` proof matters most in exactly that state.

- **Authorities are disabled, not deleted, and restored from the mode handlers.** `resolveStorageAuthority` skips a disabled authority, and re-enabling restores the exact record, which a seeded fixture cannot be recreated from since the harness never chose its configuration. The restore also runs from `on_mode_error` and the failure branch of `on_mode_return`, gated on the stack surviving: a failure inside the window would otherwise leave the stack `COMPOSE_SMOKE_KEEP_ON_FAIL` preserves with `fixture-floecat` disabled, covering the whole `floecat` bucket including the seeded `examples` catalog, and producing the same credential-resolution failure whoever kept the stack is investigating. Against a stack about to be `down -v`'d it only costs its own timeout, so it is skipped there.

- **One log-scraping assertion remains and is deliberately untouched.** `assert_remote_file_group_worker_activity` predates this work, added by #280 and modified by #408, and whether it has a non-log alternative was not investigated here.

## Additional Notes

Gateway response bodies are never echoed, on success or failure, because they carry live temporary storage credentials, the same reason the existing Polaris `loadTable` precondition omits its own. The JSON checks are guarded so a non-JSON body exits quietly rather than printing itself in a traceback.

`docs/iceberg-rest-gateway.md` needed a correction too: it scoped the delegated Polaris scenario to `localstack-remote`, which the mode-gating change makes false.

Two follow-ups deliberately not taken. The two gateway `loadTable` checks are a near-copy that should fold into one helper, and they have already drifted in formatting; that was attempted and reverted because the shared version grew the embedded Python well beyond what this file otherwise carries, and `jq` is not used anywhere in it. Separately, `polaris-vended-creds` runs only where `smoke_scope = full`, so the `polaris-integration` mode exercises the Integration vend but not the Connector one.

### DIFF
--- a/tools/compose-smoke.sh
+++ b/tools/compose-smoke.sh
@@ -337,6 +337,171 @@ assert_remote_file_group_worker_activity() {
     "commitLeasedFileGroupResult"
 }
 
+# Ids of enabled storage authorities covering a location, one per line, on stdout only.
+#
+# Mirrors StorageAuthorityResolver.matchesLocationPrefix, which strips a trailing slash and then
+# requires a path boundary. A substring test is wrong both ways: it misses s3://floecat covering
+# s3://floecat/sales/..., and counts s3://floecat-delta as covering s3://floecat-delta-vended/...
+authorities_covering() {
+  local compose_cmd="$1"
+  local location="$2"
+
+  run_cli_script "$compose_cmd" "account t-0001
+storage-authority list
+quit" | COVERED_LOCATION="$location" python3 -c '
+import os
+import re
+import sys
+
+location = os.environ["COVERED_LOCATION"].strip()
+
+
+def covers(prefix):
+    prefix = prefix.strip().rstrip("/")
+    if not prefix or not location.startswith(prefix):
+        return False
+    return len(location) == len(prefix) or location[len(prefix)] == "/"
+
+
+UUID = re.compile(r"^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$")
+
+rows = 0
+for line in sys.stdin:
+    parts = line.split()
+    # Only the last two fields are at a fixed offset, and the enabled flag reading true or false is
+    # what identifies a row. Keyed on the id rather than the name because the field count is not
+    # fixed either: a blank TYPE collapses to four fields and a multi-word name adds fields in the
+    # middle. storage-authority update takes an id directly.
+    if len(parts) < 4 or parts[-2] not in ("true", "false"):
+        continue
+    identifier = None
+    for part in parts:
+        if UUID.match(part):
+            identifier = part
+            break
+    if identifier is None:
+        continue
+    rows += 1
+    if parts[-2] == "true" and covers(parts[-1]):
+        print(identifier)
+
+# Recognising nothing is a broken parse, not an empty catalog: every caller reads no output as
+# "no authority covers this", so a column change would turn each of them green asserting nothing.
+if rows == 0:
+    print("parsed no storage-authority rows; the list columns may have changed", file=sys.stderr)
+    sys.exit(1)
+'
+}
+
+# Flips the enabled flag on each authority named on stdin.
+#
+# Disabled rather than deleted, so re-enabling restores the exact record; a seeded fixture cannot
+# be recreated from arguments guessed at the call site. Callers use a here-string, not a pipe: as a
+# pipeline sink this runs in a subshell that inherits the ERR trap under set -E, so a failed update
+# would run the mode handler there, destroy the stack, then run it again in the parent.
+set_authorities_enabled() {
+  local compose_cmd="$1"
+  local label="$2"
+  local enabled="$3"
+  local authority_name
+  local -a authority_names=()
+
+  # Drained before the first update runs. run_cli_script pipes its own script into the container
+  # and so leaves this stdin alone, but a reader should not have to confirm that to trust the loop.
+  while IFS= read -r authority_name; do
+    [ -n "$authority_name" ] || continue
+    authority_names+=("$authority_name")
+  done
+
+  if [ "${#authority_names[@]}" -eq 0 ]; then
+    return 0
+  fi
+
+  # One shell invocation for the whole set. The CLI takes a multi-command script, as the setup
+  # blocks in this file already rely on, and a container start per authority adds up: two covering
+  # authorities across four disable and re-enable rounds is eight of them per localstack mode.
+  local script="account t-0001"
+  for authority_name in "${authority_names[@]}"; do
+    script="$script
+storage-authority update $authority_name --enabled $enabled"
+  done
+  script="$script
+quit"
+
+  local out
+  out=$(run_cli_script "$compose_cmd" "$script")
+  echo "$out"
+  # The combined output names whichever update failed, so one assertion over all of them keeps the
+  # signal without a label per authority.
+  assert_not_contains "$label storage authority enabled=$enabled" "$out" "! "
+}
+
+# Fails unless no enabled storage authority covers the location.
+assert_no_authority_covers() {
+  local compose_cmd="$1"
+  local label="$2"
+  local location="$3"
+  local still_covering
+
+  still_covering=$(authorities_covering "$compose_cmd" "$location")
+  if [ -n "$still_covering" ]; then
+    echo "[FAIL] $label storage authorities still cover $location"
+    printf '%s\n' "$still_covering"
+    return 1
+  fi
+  echo "[PASS] $label no storage authority covers $location"
+}
+
+# Fails unless a gateway loadTable returns a complete vended session tuple.
+#
+# Shared by the Polaris and Unity checks, which differ only in label and URL. The body is never
+# echoed: it carries live temporary storage credentials, the same reason the Polaris loadTable
+# precondition omits its own. The status is safe and is reported.
+assert_gateway_vends_session_tuple() {
+  local compose_project="$1"
+  local label="$2"
+  local url="$3"
+
+  local resp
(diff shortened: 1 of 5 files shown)


tools/compose-smoke.sh
```diff
@@ -337,6 +337,110 @@ assert_remote_file_group_worker_activity() {
...
+    if not prefix or not location.startswith(prefix):
+        return False
+    return len(location) == len(prefix) or location[len(prefix)] == "/"
+
+
+rows = 0
+for line in sys.stdin:
+    parts = line.split()
+    # AUTHORITY_ID DISPLAY_NAME TYPE ON PREFIX, with the CLI prompt possibly ahead of the header.
+    # The prefix is last and the enabled flag second to last, which is also what identifies a row:
```
mcusack-floe: tools/compose-smoke.sh:378 - authorities_covering drops list rows with fewer than 5 whitespace fields, but an authority with a blank TYPE prints only 4. 

StorageAuthorityResolver.isSupportedAuthorityType treats blank as "s3", so that authority genuinely covers the location. It never gets disabled, assert_no_authority_covers still prints [PASS], and the import resolves through it without vending - precisely the false pass this PR exists to remove. Reachable because StorageAuthorityServiceImpl:957 normalizes blank type on create but the update merge at :1036 does not.

tools/compose-smoke.sh
```diff
@@ -337,6 +337,110 @@ assert_remote_file_group_worker_activity() {
...
+    # anything else in the output, banner or prompt or header, fails the true/false test.
+    if len(parts) < 5 or parts[-2] not in ("true", "false"):
+        continue
+    rows += 1
+    if parts[-2] == "true" and covers(parts[-1]):
```
mcusack-floe, line 385: tools/compose-smoke.sh:383 - print(parts[-4]) yields the display name only for single-token names. storage-authority create "my auth" is accepted, and then set_authorities_enabled issues update auth --enabled false, resolving nothing. Loud failure rather than a silent pass, but it still breaks the test.

mcusack-floe (changes requested):

tools/compose-smoke.sh
kwesterfeld-floe: Fixed in 32058235b. The parser is keyed on the AUTHORITY_ID now: a row is identified by the enabled flag reading true or false plus a UUID present anywhere in the line, and the UUID is what gets reported. That drops the dependency on the field count and on the middle columns entirely, so a blank TYPE collapsing to four fields no longer skips the row.

One correction to the reachability, which I checked after you raised it. The mask merge at `case "type"` does clear and conditionally set as you describe, but the merged spec then passes through `buildAuthority` (the `MutationOps.updateWithPreconditions` call), which normalizes `trimToNull(spec.getType()) == null ? "s3"` on the update path as well as on create. So no persisted record prints four fields today, and this is hardening rather than a live false pass. Still worth closing for the reason you give: `isSupportedAuthorityType` reads blank as s3, so such a record would genuinely cover while being reported as absent.

The sibling issue on the display name was live, and the same change fixes it.

tools/compose-smoke.sh
kwesterfeld-floe: Fixed in 32058235b, by the same change as the sibling thread: the parser reports the AUTHORITY_ID rather than the display name, so a multi-word name no longer matters. `resolveId` short-circuits on `looksLikeUuid` before it lists, so `set_authorities_enabled` passes the id straight through and needed no change.

This one was live, unlike the blank-TYPE case: the server requires only `trimToNull(spec.getDisplayName()) != null`, so `storage-authority create "my auth"` is accepted. Your read on the symptom was right too, and it is the part that made it worth fixing rather than tolerating: it failed loudly, but against a legal configuration and while naming a check that was not the one at fault.

Exercised offline across seven shapes: typed row, blank type, multi-word name, blank type with a multi-word name, a disabled row (parsed but not reported), a non-covering sibling, and renamed columns still exiting non-zero.

mcusack-floe (approved):

PR #513, opened 2026-09-07
feat(delta): resolve source column identities

mcusack-floe:
## Summary

Adds a format-neutral resolved-schema model and a Delta resolver that interprets column-mapping modes and records logical paths, physical paths, and native field IDs. It validates malformed or conflicting metadata before downstream code can rely on it.

## Type of Change

- [] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist


mrouvroy-floe:
I like the added code but it looks like a parallel implementation of the existing Delta schema mapping/traversal logic instead of a replacement. I believe we need a single Delta schema adapter that all consumers use: stats, footer resolution, page indexes, and canonical identity. Otherwise we’ll keep fixing schema-mapping bugs in two places.

Likewise, do we also intend to extend this to support  Iceberg and generic format ? I think the adapter should expose Iceberg field IDs instead of Delta column mapping metadata in such cases for instance. 
And that would leave it the only authority on such topics, making the debugging and logic much simpler IMO

mcusack-floe:
> I like the added code but it looks like a parallel implementation of the existing Delta schema mapping/traversal logic instead of a replacement. I believe we need a single Delta schema adapter that all consumers use: stats, footer resolution, page indexes, and canonical identity. Otherwise we’ll keep fixing schema-mapping bugs in two places.

Agreed on the single-adapter point, and the resolver now works that way. I've pushed a
follow-up commit that moves the Delta producer into `connectors/catalogs/delta` and walks
Delta Kernel's `StructType` instead of re-reading the schema JSON with Jackson

On the "two places" concern more broadly: this PR is deliberately additive so the resolver
lands reviewable on its own. #517 does the replacement.

> 
> Likewise, do we also intend to extend this to support Iceberg and generic format ? I think the adapter should expose Iceberg field IDs instead of Delta column mapping metadata in such cases for instance. And that would leave it the only authority on such topics, making the debugging and logic much simpler IMO

Yes, this will be a follow-on PR stack to unify with Iceberg.

mrouvroy-floe (approved):
Looks much better now to me. The foundations are now generic enough that we can build iceberg on to later and clean up delta mapper in future PRs. 
Thanks !

PR #515, opened 2026-09-07
feat(identity): Add canonical schema identity reconciliation and wire contracts

mcusack-floe:
## Summary

Adds canonical IDs that persist across schema versions, including deterministic fingerprints and serialized identity state. Mapped schemas reconcile by native field ID, while unmapped schemas use exact structured paths with safeguards against skipped versions and ID reuse.

## Type of Change

- [] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist


core/schema-identity/src/main/java/ai/floedb/floecat/schema/identity/SchemaIdentityReconciler.java
```diff
@@ -0,0 +1,233 @@
...
+      updateString(digest, fingerprint);
+      digest.update(ByteBuffer.allocate(Long.BYTES).putLong(highWaterMark).array());
+      digest.update(ByteBuffer.allocate(Long.BYTES).putLong(sourceVersion).array());
+      return "sha256:" + HexFormat.of().formatHex(digest.digest());
+    } catch (NoSuchAlgorithmException e) {
+      throw new IllegalStateException("SHA-256 is unavailable", e);
+    }
+  }
+
+  public static String structuredPath(ColumnPath path) {
```
mrouvroy-floe: structuredPath` should be with `ColumnPath` IMO. 
Keeping the canonical, unambiguous (kind + length-prefixed name) encoding on ColumnPath (e.g. as stableKey()) co-locates all three renderings and their documentation in one place. The reconciler currently owns path-format knowledge it only needs for fingerprinting, which makes the stability contract harder to discover and audit.

core/schema-identity/src/main/java/ai/floedb/floecat/schema/identity/SchemaIdentityReconciler.java
```diff
@@ -0,0 +1,233 @@
...
+
+  private static void validateMode(IdentityMode mode, SchemaIdentityState previous) {
+    Objects.requireNonNull(mode, "mode");
+    if (mode != previous.mode()) {
+      throw new IllegalArgumentException(
+          "Identity mode changed; a clean identity reset is required");
+    }
+  }
+
+  private static void validateHighWaterMark(long highWaterMark) {
```
mrouvroy-floe: validateHighWaterMark here and the SchemaIdentityState constructor (SchemaIdentityState.java:47-50) both enforce the identical 0 <= x <= MAX_ALLOCATED_ID bound with the same error message.

Consolidating into a single helper (e.g. a package-private CanonicalColumnId.checkAllocatedRange) keeps the two in step if the ID-space bounds ever change.

core/proto/src/main/proto/floecat/catalog/snapshot.proto
```diff
@@ -39,6 +39,49 @@ message Snapshot {
   int32 schema_id = 11;
   optional string metadata_location = 12;
   SnapshotReuseManifestRef reuse_manifest_ref = 13;
+  // Authoritative column identities for this source schema version. The fingerprint binds schema,
+  // capture, and reuse consumers to the exact same mapping.
+  ColumnIdentityMap column_identity_map = 14;
+  string column_identity_fingerprint = 15;
```
mrouvroy-floe: Snapshot and SnapshotSpec (line 168) each carry both column_identity_map — whose ColumnIdentityMap.fingerprint (line 82) is the authoritative digest — and a separate top-level column_identity_fingerprint, creating two sources of truth for the same value.

Nothing in the contract enforces column_identity_fingerprint == column_identity_map.fingerprint. A future producer that sets one but not the other, or updates the map without copying the fingerprint, will silently mis-gate stats reuse. Other messages (SnapshotManifestEntry, TablePin, RelationPinIdentity, stats messages) correctly embed only the bare fingerprint since they don't carry the map.

If the redundancy is a deliberate optimization (checking compatibility without deserializing the full map ?), document the invariant explicitly and add a validation step on read. Otherwise, drop column_identity_fingerprint and derive it from column_identity_map.fingerprint at the call sites.

core/proto/src/main/proto/floecat/catalog/table.proto
```diff
@@ -35,6 +35,8 @@ enum ColumnIdAlgorithm {
   CID_UNSPECIFIED = 0;
   CID_FIELD_ID = 1;
   CID_PATH_ORDINAL = 2;
+  // IDs are supplied by the snapshot's authoritative column identity map.
+  CID_CANONICAL_MAP = 3;
```
mrouvroy-floe: CID_CANONICAL_MAP is now a valid ColumnIdAlgorithm in the public proto contract, but no consumer implements it.  ColumnIdComputer.compute(...) falls through to default -> 0L, so tables using this algorithm silently get SchemaColumn rows with no computed id. 
Is that expected ?

PR #517, opened 2026-09-07
Cut Delta over to canonical column identities

mcusack-floe:
## Summary

Makes persisted canonical maps authoritative for Delta schemas, statistics, constraints, and page indexes. Statistics are fingerprint-gated, mixed-identity rollups and sketch reuse are rejected, missing maps fail fast, and the superseded manual Delta JSON parser fallbacks are removed.

## Type of Change

- [X] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [ ] No breaking changes

## Checklist

## Additional Notes

This is contract-breaking with core repo. Needs coordinated changes there to merge.

### DIFF
--- a/reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/GrpcReconcilerBackend.java
+++ b/reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/GrpcReconcilerBackend.java
@@ -16,7 +16,6 @@
 
 package ai.floedb.floecat.reconciler.impl;
 
-import ai.floedb.floecat.catalog.rpc.ColumnIdentityMap;
 import ai.floedb.floecat.catalog.rpc.CreateNamespaceRequest;
 import ai.floedb.floecat.catalog.rpc.CreateSnapshotRequest;
 import ai.floedb.floecat.catalog.rpc.CreateTableRequest;
--- a/reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/GrpcReconcilerBackend.java
+++ b/reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/GrpcReconcilerBackend.java
@@ -555,6 +555,10 @@ public class GrpcReconcilerBackend implements ReconcilerBackend {
     if (snapshotId < 0) {
       return Optional.empty();
     }
+    ai.floedb.floecat.catalog.rpc.ColumnIdentityMap identityMap =
+        fetchSnapshot(ctx, tableId, snapshotId)
+            .map(Snapshot::getColumnIdentityMap)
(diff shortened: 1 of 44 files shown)


reconciler/src/main/java/ai/floedb/floecat/reconciler/impl/GrpcReconcilerBackend.java
```diff
@@ -583,17 +593,30 @@ public Optional<FloecatConnector.SnapshotFileDelta> fetchSnapshotFileDelta(
...
+                    tableId,
+                    baseSnapshotId,
+                    targetSnapshotId)
+                .map(
+                    delta ->
+                        new FloecatConnector.SnapshotFileDelta(
+                            delta.addedDataFiles(),
+                            delta.removedDataFilePaths(),
+                            delta.deleteArtifactsChanged(),
+                            ColumnIdentityExecutionSchema.attach(
```
mrouvroy-floe, line 618: `fetchSnapshotFileDelta` (and `fetchSnapshotFilePlan`) now run the connector's executionSchemaJson through `ColumnIdentityExecutionSchema.attach`, which parses the schema JSON and adds two top-level keys (`_floecat_column_identity_map`, `_floecat_column_identity_fingerprint`) whenever the identity map is non-default — i.e. for every canonical Delta table.

That mutated JSON flows into `RemoteSnapshotPlanningReconcileExecutor.planAppendOnlyDelta`, which decides whether the append-only fast path applies via:

```java
String baseSchema = historical.snapshot().getSchemaJson();      // raw, from the snapshot store
if (!schemasEquivalent(baseSchema, delta.executionSchemaJson())) // target now carries injected keys
```
`schemasEquivalent` returns `parsedBase.equals(parsedTarget)`. 

The base side is the stored (un-injected) schema, the target side is the injected schema, so the two objects differ by two extra keys and the comparison is now always false for canonical Delta tables. Append-only deltas are silently rejected and every snapshot falls back to a full recapture — a real cost/latency regression affecting precisely the tables this PR cuts over.

The injected JSON is also fed to `partitionFiles`, `RemoteFileArtifactReusePlanner.enrichFromBundles`, and `SnapshotPlanManifestIds` digests, so the schema string is doing double duty as both a schema and an identity-transport envelope.

Consider carrying the identity map as a first-class field on `SnapshotFilePlan/SnapshotFileDelta` (and the remote payload) rather than embedding it in executionSchemaJson; or, at minimum, make schemasEquivalent compare the schema after stripping the injected identity keys so append-only detection keeps working. Either way this path needs a regression test, since neither `RemoteSnapshotPlanningReconcileExecutor` nor its test was touched here.

PR #519, opened 2026-09-07
feat(catalog-access): add Delta Sharing Catalog Integration

kwesterfeld-floe:
## Summary

Adds Delta Sharing as a Catalog Integration type, so a share a partner grants you can be reconciled into a Floecat catalog and read through vended storage credentials.

Integration type only, and directory access only. The protocol offers two ways to read a shared table: `dir` mode issues temporary cloud credentials over the table's location, and `url` mode returns presigned per-file URLs from a query endpoint. Only the first is expressible as a storage credential, so a share offering `url` alone is unsupported here rather than partially supported: such a table is discoverable and describable, and refused when the overlay reconciles rather than materialized as a table nothing can open.

Changes:

- `CIT_DELTA_SHARING` and its `CatalogProtocol.DELTA_SHARING` mapping. Bearer only, because the sharing protocol defines no other scheme.
- A recipient client under `connectors/clients/delta-sharing`: shares, schemas, tables, metadata and temporary table credentials, with paging, NDJSON, both metadata response shapes and three cloud credential shapes decoded in the transport. Deliberately no query method.
- A provider under `catalog-access/delta-sharing`. `share.schema.table` maps onto `NamespacePath` without translation, and the vend publishes the temporary-table-credentials session triad scoped to the location the server named.
- Two shared modules, both extracted from Unity because a second Delta-backed provider needs the same thing: `core/http-endpoint-guards` for the catalog and storage endpoint gates, and `catalog-access/delta-common` for the Delta log storage probe.
- A validation fix that is not Delta Sharing specific: a provider answering `UNSUPPORTED` now reports the `_UNSUPPORTED` issue rather than `_FAILED`, because a capability can be per table rather than per integration.
- A compose smoke scenario covering the vend end to end.

## Type of Change

- [x] fix (bug fix)

## Testing

- [x] `make fmt`
- [ ] `make verify`

Full offline unit suite green across the repo, zero failures. Every behaviour this PR fixes has a test that fails without the fix.

**The compose smoke ran in CI and the Delta Sharing scenario passed**, with zero `[FAIL]` lines in the job and all four smoke modes green:

```
[PASS] Delta Sharing integration no storage authority covers s3://floecat-sharing-vended/call_center
[PASS] Delta Sharing integration validation        (valid: true)
[PASS] Delta Sharing credential vending validation
[PASS] Delta Sharing storage access validation
[PASS] Delta Sharing share vended temporary credentials for floecat_smoke.sharing.call_center
[PASS] Delta Sharing overlay reconciliation        tables_created: 1, branches_skipped: 0, objects_skipped: 0
[PASS] Delta Sharing Catalog Integration gateway credential vending
[PASS] Delta Sharing rejects a wrong recipient token
```

The fixture lives in a bucket absent from `COMPOSE_SMOKE_LOCALSTACK_BUCKETS` and the scenario asserts no authority covers it *before* creating anything; `CIVCT_STORAGE_ACCESS` is a real `ListObjectsV2` plus a ranged `GetObject` using only what the share vended; and the stub reports its own request log, so the credential call is named rather than inferred from a check that passed.

The scenario has two further variants, both green in CI: `bare-shape`, which recreates the stub in the reference server's shape and sends neither `accessModes` nor a location on any listing, and `legacy-format`, which answers in the flat `parquet` response format. Those are the end-to-end exercise of the three-surface location chain and the legacy shape argued for below:

```
[PASS] Delta Sharing bare-shape validation
[PASS] Delta Sharing bare-shape reconciliation
[PASS] Delta Sharing bare-shape storage location
[PASS] Delta Sharing bare-shape storage access
[PASS] Delta Sharing bare-shape completeness
[PASS] Delta Sharing legacy-format validation
```

What the smoke does not cover is a Delta scan through `S3V2FileSystemClient`: no fixture has a space in its object key, which is the shape the canonicalized location addresses.

The unstated-modes half of that variant did nothing until this round. It sets the stub's `SHARING_ACCESS_MODES` to empty, and the compose default was written `${VAR:-url,dir}`, which substitutes the default for an empty value as well as an unset one -- so the stub restarted advertising `url,dir` and kept sending the field. It is `${VAR-url,dir}` now, which `docker compose config` confirms: unset resolves to `url,dir`, explicitly empty resolves to empty. Its two neighbours keep `:-` deliberately, since an empty response format is not a shape the stub can serve.

- [x] No breaking changes

Additive. A new enum value, two new connection properties, and two new modules; no existing behaviour changes except the validation issue mapping below.

New connection properties, both optional:

- `delta.sharing.strict-access-modes` -- reads a table stating no access modes as `url` only, refusing without a round trip. Off by default; see the Reviewers note.
- `delta.sharing.reader-features` -- comma-separated Delta reader features this deployment can process. Empty by default.

`FLOECAT_SECURITY_ALLOW_CLEARTEXT_S3_ENDPOINTS` now governs the Delta Sharing `s3.endpoint` as well as Unity's, for the same reason: this vend also publishes an AWS session token.

Three system properties bound what one recipient client will pull: `floecat.delta-sharing.max-pages`, `floecat.delta-sharing.max-response-bytes` and `floecat.delta-sharing.max-listing-bytes`. All three are documented in `docs/operations.md` alongside the connection properties, matching how the Unity equivalents are already documented.

`s3.region` is now resolved for a Delta Sharing Integration the way it already was for Unity, from the stated region aliases or the deployment default. Both run the same Delta log probe, whose `us-east-1` fallback would otherwise answer PermanentRedirect for a share whose table lives elsewhere -- an Integration permanently reporting a storage-access failure that does not exist in practice.

**The `UNSUPPORTED` remapping also changed one signal it should not have, and that is corrected here. The storage probe's "no Delta log object" branch is not a capability boundary: an empty log prefix means the location is not a table root -- a wrong region, a wrong prefix, a broadly scoped credential -- all of which an operator can correct. It now reports `INVALID_CONFIGURATION`, so validation shows `CIVI_STORAGE_ACCESS_FAILED` rather than a terminal `..._UNSUPPORTED`, which is the signal that branch carried before this PR. `docs/operations.md` also lists the refusal shapes completely now: auxiliary locations and an unservable cloud were missing, and the no-location entry conflated a skip with a classified failure.

One behaviour change reaches Unity today.** `CatalogIntegrationDiscovery` now maps an upstream `UNSUPPORTED` to `CIVI_CREDENTIAL_VENDING_UNSUPPORTED` / `CIVI_STORAGE_ACCESS_UNSUPPORTED` instead of the `_FAILED` issue. `UnityCatalogAccessClient.vendStorageCredentials` already answers `UNSUPPORTED` on several paths -- a non-AWS cloud, a table the provider cannot vend for, an Integration that does not delegate -- and so does the storage probe when a table has no addressable S3 location or no Delta log object. A Unity Integration whose sampled tables all refuse that way previously validated as `CIVI_CREDENTIAL_VENDING_FAILED` and now validates as `..._UNSUPPORTED`. That is the intent of the fix rather than a side effect, but a UI keying on the issue enum will see the changed value. `valid: false` either way; the check status and summary are unchanged.

## Checklist

## Reviewers

- **A table stating no access modes is asked rather than refused, and the default is deliberate.** The protocol reads an absent `accessModes` as `url` only. The reference server implements `temporary-table-credentials` against a real `AwsCredentialVendor` and never sends the field -- a repository-wide search finds `accessModes` only in `PROTOCOL.md`, and PR delta-io/delta-sharing#845, which added directory access to the protocol, changed exactly that one file. Holding to the protocol reading refuses every table on a server that would have answered. So the vend asks and reports what the server says: `NOT_FOUND` (no such endpoint) and `INVALID_REQUEST` (the protocol's own refusal for a table not offering `dir`) become `UNSUPPORTED`, while a rejected token or an unreachable server is reported as itself. That mapping is not gated on whether the table stated its modes. The vend applied it only to an unstated table, so one HTTP 400 was a capability boundary on the load path and a configuration error on the vend path; the server's refusal settles it either way, since a table whose listing claims `dir` and whose credential endpoint refuses does not have directory access. Both paths now reach the server through one classifier, which is also what makes the two answers impossible to drift apart. A table that explicitly says `["url"]` is refused either way -- the knob only governs the ambiguous case. The modes are read from the `metaData` action as well as the listing, because the delta response format carries them there and a server may use either. A refusal on either surface is decisive, and that is a change: the load preferred the listing while the vend reads the metaData action alone, so a server saying `dir` on one and `url` on the other reconciled at the load and was refused at the vend on every read for the life of the table. The load cannot be more permissive than the vend, and the vend sees one surface, so both are consulted here and either one refusing settles it. The opposite disagreement is refused too, which costs only a table that nothing could have read through this path. The load path prefers the listing; the vend reads the metadata action and, under the strict setting only, falls back to the listing before refusing -- strict refuses without asking the server, so it must not refuse on a half-consulted answer.

- **The vend checks the credential covers the table, and validation probes it.** Only validation calls `validateStorageAccess`, so a check made solely there left every runtime resolve unguarded against a server returning a credential scoped somewhere unrelated or far broader; the vend now makes the same check the Unity provider makes in its own -- and so does the load, for the surface the vend cannot see. Where a surface states a location and the credential endpoint has also answered, the two are compared: reconciling a stated location the answered credential does not reach materialises a table every read of which is scoped elsewhere, and validation is optional, so nothing else need catch it. A credential scoped *above* the stated location still reaches it and is left alone. It is not quite the same check on every server shape, and the difference is worth stating: the vend reads the location from the metadata action alone, while the load resolves the listing first, so a server stating a location only on its listing reconciles and then has nothing for the vend to compare a scope against. The check is skipped rather than failed there, because the alternative is a schema listing per read -- the cost this path was changed to avoid. Validation consults both surfaces, so an operator sees a mis-scoped credential then; what is not caught is a server that begins mis-scoping afterwards. Closing it means carrying the load's location to the vend, and the vendor's fresh-client-per-resolve shape gives nowhere to put it. On the validation path, scope first, then probe. `covers()` answers without a network call and a credential scoped elsewhere is a provider bug rather than a permission an operator can grant. The probe then does the real read, because scope alone passes on a grant that cannot read -- the failure an operator otherwise finds at query time. Where the listing carries no location, which is exactly the unstated-access-modes case, it falls back to the metadata endpoint.

- **The smoke uses a stub, and the reason is upstream packaging rather than preference.** The OSS server does implement directory access -- PR delta-io/delta-sharing#854, merged March 2026 -- but no published image contains it: `deltaio/delta-sharing-server` tags stop at 0.7.8 from April 2024. Running the real server means building Scala from source in the harness. The stub answers in the delta response format by default, so the smoke exercises the shape a conforming server sends rather than the flat one that happened to parse. Separately, its STS client is AWS SDK v1 with no endpoint override, though the harness already solves that class of problem for Unity with a `sts.us-east-1.amazonaws.com` network alias.

- **The vend reads one table's metadata; it does not list the schema.** A credential is vended on every storage-authority resolve, so finding the table by listing paged the whole schema before every credential POST -- unbounded in the schema's table count, and quadratic across a reconcile that loads each listed table. The delta response format carries the access modes, location and auxiliary locations on the metadata action, which is everything the listing was consulted for, so `findTable` survives only in `loadTable`, where the share id is needed and no other source has it. That listing is memoised per schema for the client's lifetime, which is one reconcile pass: the reconciler lists a schema once and then loads every table in it, so looking each one up by paging the listing again made the pass quadratic. The vend pages nothing at all -- `SourceCatalogCredentialVendor` opens a fresh client per resolve and closes it in the same method, so a listing there is never absorbed by that memo. Tests assert the listing is never called on the vend or probe paths, and called once across three loads.

- **A table id is scoped to its share, so the identity is too.** The protocol makes a table id unique within its share and `shareId` unique across the server, so a bare table id collides between two shares -- including the same underlying table shared twice -- and the reconciler abandons an overlay over a duplicate stable identity. The identity is `shareId:tableId`, falling back to the metadata action's id before falling back to a name, and reported unstable when the share component is a name rather than an id.

- **A table with auxiliary locations is refused before it reconciles.** The protocol wants `temporary-table-credentials` called once per location, root and auxiliary alike, and tells a client that cannot read from several to fall back to url access or raise. The storage contract carries one credential over one scope prefix, so there is nothing to carry the rest; vending only the root would reconcile and validate cleanly and then fail at query time on the auxiliary files.

- **`Failure.OTHER` maps to `UNAVAILABLE`, not `UNSUPPORTED`.** A review suggested a new SPI code for unclassified upstream statuses. I used the existing one instead: a new enum value forces a decision at four consumer switches that have no distinct answer for it, and `INTERNAL` would have been worse because `GrpcErrors` hides an `INTERNAL` message, losing the HTTP status that makes the failure diagnosable. `UNAVAILABLE` already means the upstream stalled or fell over, keeps the status in the message, and reports `CIVI_CREDENTIAL_VENDING_FAILED` rather than the unsupported issue. The cost is that an unclassified 451 is treated as possibly-transient and ends the validation walk rather than skipping one table, which is the safer reading for a status likely to affect the whole share.

- **An envelope this client cannot use is a classified failure, not an IllegalArgumentException.** A credentials response missing its location, its expiry, a recognised cloud or part of its AWS session, and a metaData action missing its schema, all escaped unclassified -- most through the record's own `requireText` as an `IllegalArgumentException`, and the incomplete AWS session as `UNSUPPORTED`, which is a per-table skip meaning the provider will never do this -- past this transport's classification and past the catalog client's `translate`, arriving at the reconciler naming neither the table nor the server. The expiry is the one that mattered most: `SourceCatalogCredentialVendor` refuses a Catalog Integration session without one, while validation does not read an absent expiry as expired, so a share omitting it would have reported the integration valid and then failed every read. All three int64 fields -- `expirationTime`, `minReaderVersion` and `version` -- take one rule, and it parses with `Long`: an expiry in epoch milliseconds overflows an int, so a server rendering it as a JSON string, which is the standard protobuf encoding for int64, has to be accepted. It is type-checked as well, the way `version` and `minReaderVersion` are: `asLong` answers 1 for `true`, which clears both bounds and yields an expiry one millisecond after the epoch -- a session the vendor then refuses as expired, reported to an operator as an expired credential rather than a malformed response. It is bounded above as well, which it was not: a value past the last instant a proto `Timestamp` carries is a unit mismatch rather than a date -- microseconds where milliseconds were meant, which Unity deployments have been seen to send -- and without the bound it passes every downstream expiry check by looking far in the future and then throws inside `Timestamps.fromMillis` while the gRPC response is built, as an unclassified `RuntimeException` in a handler, on a share that validated clean, for every read. `MAX_EXPIRY_EPOCH_MILLIS` already existed in `FloecatConnector` and in the Unity client; this parser had only the lower half of the check.

- **The response cap cannot be the largest int.** The bounded reader asks for the cap plus one byte to spot an overrun, and that addition overflows at `Integer.MAX_VALUE` -- so a deployment setting `floecat.delta-sharing.max-response-bytes` to its largest allowed value got a negative length and every response failed as a transport error. The Unity client already refuses that value on the same property; this one copied the pattern without the guard.

- **A table reconciles with a storage location or not at all.** Five shapes are refused at the load and counted in `objects_skipped`: `url` access alone; no stated modes under `delta.sharing.strict-access-modes`; a server that will not vend for the table, which is `NOT_FOUND` or the protocol's own `INVALID_REQUEST` and is the shape a table with no obtainable location actually takes; a credential scoped to a whole bucket; and a location on a cloud this provider cannot vend for -- an `abfss://` or `gs://` table would otherwise load, reconcile, and fail at every scan when the vend refused its cloud. Strict mode applies only where the modes were genuinely absent, which is the ambiguous case it governs; an earlier revision refused a table stating `dir` with no location and told the operator it stated no modes. They were not always treated alike -- an earlier revision refused the third and let the first materialize, which is the same unreadable outcome by a different route -- and collapsing them into one rule is what made that visible. Strict mode consults the listing when the metaData action states nothing, matching the vend's own fallback, so it does not spend a credential request per table against a server stating its modes where the protocol defines them -- both surfaces are already in hand, so that costs nothing. It refuses without a request, since asking where an unstated table lives is the round trip the knob turns off, and answering would reconcile a table whose every vend then fails. It also refuses whether or not a location was found, which it did not before: the check sat below the load's early return on a stated location, so an unstated table whose listing named a location loaded and was then refused by the vend on the rule the load had skipped. That is the reconcile-then-fail-every-read outcome, arrived at by the two paths disagreeing rather than by either being wrong on its own. Knowing where a table lives says nothing about whether it may be read that way, which is the only question the knob answers, and a test now asserts both paths refuse the same shape.

- **A space in an object key is a location, and the location that reconciles has to be readable.** `java.net.URI` rejects a space and S3 permits it in a key, so `s3://bucket/db/my table` answered no parse at all: the shared probe read it as having no addressable bucket and this provider reported it as living on no addressable scheme -- pointing an operator at cloud support for a table that was fine, and counting it in `objects_skipped`. The space is percent-encoded before parsing, and `getPath` decodes it again so the key the probe asks S3 for is the key that was written. Other characters `URI` rejects and S3 permits still answer no parse; a space is the one that occurs, and the test says which is which. Listed paths are built by encoding the path alone, with no host for `URI` to validate: passing the bucket as a host makes `URI` reject exactly the names below, and the fallback then emits the raw key, losing the encoding for the one bucket shape this supports. The reader also derives a bucket the way the probe does, falling back to the authority where `getHost` answers null -- S3 permitted names `java.net.URI` will not parse as a host, an underscore being the one that survives in older regions. A name the probe accepts and the reader cannot address is a table that reconciles and then fails every scan. It also required a change to the reader, without which none of it worked. Encoding the root made the *table* parseable and not a scan of it: `S3V2FileSystemClient.listFrom` rebuilt every listed path as `"s3://" + bucket + "/" + S3Object.key()`, from the raw key, so the second Delta log entry arrived with the space restored and `URI.create` rejected it -- a table that reconciled, validated, and failed at scan, which is the outcome the canonicalisation exists to prevent. Listed paths are now built a segment at a time, and `getPath` decodes, so S3 is asked for the key that was written. Encoding the key as one path was wrong in exactly the shape `stripLeadingSlash` preserves: an S3 key may begin with a slash, so the string handed to `URI` opened `//`, which it read as an authority -- `getRawPath` answered without the first segment and raised nothing, so the fallback never ran and every entry listed under `s3://bucket//table` addressed a key missing its leading name. Splitting on the separator keeps the empty leading segment that slash stands for, and no single segment can hold a slash to re-parse. For a key holding nothing `URI` objects to the result is byte-identical to the concatenation it replaces, so only the paths that used to throw change at all. Refusing such tables instead was the alternative, and the wrong one: the consumers here are query workers and scan ingest, and they have to be able to read them. This reaches Unity too, through the same probe -- and for a while it reached Unity with only the probe half, which was worse than not fixing it. Unity kept publishing the raw string, so a table it used to refuse outright became one that validates clean and fails at scan. The rule now lives in `DeltaLogStorageProbe.canonicalLocation`, beside the probe whose encoding created the need for it, and both providers call it: Unity canonicalizes at its publish, its vend scope, its coverage comparison and its validation probe, with a regression test on the publish path. One rule, two callers, which is what this module is for.

  It preserves an `s3a://` scheme, folding only what the reader genuinely cannot parse -- `s3n` and any uppercase form, since `resolvePath` matches a literal lowercase `s3://` or `s3a://`. Folding `s3a` too would have put the stored location in a different scheme namespace from the prefix an operator registers: `StorageAuthorityResolver.matchesLocationPrefix` is a scheme-sensitive `startsWith` and `S3Location.parse` accepts an `s3a://` authority prefix, so a catalog reporting `s3a`, an authority registered as `s3a`, and a location stored as `s3` would never match. It also keeps a trailing space, since an S3 key may end in one and the probe does not trim either -- trimming published a different prefix from the one validation had just read. And it canonicalizes the scheme as well as the space, which it did not at first, and the gap had the same shape as the one it was written to close. `s3Serves` accepts what `StorageLocations.normalizeScheme` folds -- `s3`, `s3a`, `s3n` in any case -- while the reader's `resolvePath` matches a literal lowercase `s3://` or `s3a://` and throws on anything else. So `s3n://bucket/table` and `S3://bucket/table` validated, reconciled, and failed every scan. A helper named canonical was canonicalizing one of the two things that vary; a test now asserts everything `s3Serves` accepts is published as lowercase `s3://`.

  Encoding it only inside the probe was half a fix, and the half that was missing turned a clean refusal into a worse outcome. `requireServable` returned the string the server sent, so the table materialised as `s3://bucket/db/my table` -- and the read path calls `URI.create` on the resolved path at every site in `S3V2FileSystemClient`, which is what rejects a raw space. Before the encoding change such a table was refused at load and counted in `objects_skipped`; after it, the table reconciled, validated clean, and failed every scan. The canonical form is now what this provider publishes, vends and compares, on both sides of `covers()` -- a prefix test with one side encoded would make every such table look scoped elsewhere. Encoded rather than refused because the read path derives its key with `URI.getPath`, which decodes, so S3 is asked for the key that was written; and idempotent, since only a literal space is replaced.

- **A location does not establish directory access, and the default setting now asks even when one is stated.** The protocol permits a url-only server to state a location, so a legacy server that omits its access modes, names a location and refuses `temporary-table-credentials` is exactly the shape the ask exists to catch -- and the load's early return on a stated location was skipping it. That table reconciled and every read of it failed. An unstated table is now asked about whether or not a location is known, and the server's refusal arrives as `UNSUPPORTED` and skips it. This costs one credential request per unstated table at the load, which is the round trip the non-strict default is, and `delta.sharing.strict-access-modes` is still the way to refuse without it. A location a surface stated is what the table reconciles with; the credential's scope is only stood in for where no surface named one.

- **The credential request advertises the response type the protocol specifies, and reads both framings it offers. The whole body is parsed first, since a single-action ndjson body is itself valid JSON and that covers a server formatting its JSON across lines; line-by-line parsing is the fallback for a body genuinely carrying more than one document, which is exactly what a whole-body parse rejects. Offering ndjson while parsing a single document would fail a server that took the offer, since a second line is a trailing token -- and that arrives as `INVALID_RESPONSE`, which reaches the client as `INTERNAL` and ends the whole reconcile rather than skipping a table. A single JSON body is one line of ndjson, so both framings go through the same reader.** It sent `Accept: application/json` alone, and that response is specified as `application/x-ndjson` carrying a single JSON action -- so a server honouring content negotiation could answer 406 to every credential request, which is every reconcile and every read. It now advertises both. The parser is unaffected either way, since nothing here inspects a response content type and a one-action ndjson body is valid JSON; and the stub answers `application/json` while enforcing no negotiation, so nothing in the smoke would have caught this.

- **A location that carries a secret is not a location this serves.** `bucketOf` refuses an authority holding userinfo, but only on its fallback branch: `getHost()` answers `bucket` for `s3://user:password@bucket/table`, so that check never ran and the location was servable. A provider then stores it as the reconciled table's `storage_location` -- a password, or a signed query, in catalog metadata that anything able to read the table can read. Userinfo, query, fragment and a port are all refused now, the same set the endpoint gates already refuse on a catalog URI and none of which belong in an S3 object location.

- **The probe removes the URI separator, not every slash it finds.** An S3 object key may begin with one, so `s3://bucket//table` names the key `/table`. The read path strips exactly one and addresses that; this stripped all of them and probed a different prefix, so validation either rejected a readable table or read whatever Delta log sat at the stripped prefix and reported success for a location no scan would reach.

- **A table's location comes from three surfaces, and a table without one is refused.** The listing, then the metaData action, then the credential response. The reference server names it on neither of the first two -- both predate directory access, as its absent `accessModes` shows -- and only the credential response carries it. Without that third source the reconciler stores no `storage_location` and `setUri(source.metadataLocation().orElse(integration.getCatalogUri()))` leaves the Integration's own HTTPS catalog URI as the table's upstream reference, so the table materializes and cannot be opened by anything reading object storage. The credential call is made only when the first two are silent and neither earlier refusal applies. Validation consults the same three surfaces, ending at the scope the credential named, since on a server that states a location nowhere else that is the only one there is; without it validation refused every table before attempting a read. Taking the credential's scope as the table's root is sound because the protocol says a request naming no location answers for the table's main location, but nothing holds a server to that, so a bare bucket is refused outright. That is a narrower rule than the vend's scope check below, deliberately: tolerating a broad scope concerns a location some other surface already established, while here this is the only surface there is. Anything subtler than a whole bucket is left to validation, whose probe looks for a Delta log under the location and fails when the location is not a table -- a net rather than a proof, and named as one. The smoke stub can now withhold the location the way the reference server does, and both that and the response format are reachable from the harness, since always sending it is what hid this.

- **Nothing that carries a secret is attached as a cause or quoted in a message, and an echoed header is redacted by its name rather than by its value's shape. Every pass that had to recognise the credential was defeated in turn by a rendering nobody enumerated -- a colon, a JSON-escaped solidus, a backslash-u escape, percent-encoding -- so the redaction now matches the `Authorization` header *name* and discards everything after it to the first delimiter, whatever encoding the value arrived in. That covers a `Basic` credential too, which none of the token-shaped passes ever did. The token68 pattern stays behind it for a token quoted with no header name beside it, and the caller's own value stays as a third pass.** The token is validated when the client is constructed, because `HttpRequest.Builder.header` quotes the value it rejects and the send classifies anything it raises as `TRANSPORT` with the cause attached -- so a recipient token carrying a newline reached the cause chain the vendor logs, verbatim, on a request that never left the process. It is now refused by describing the character class and the index, never the value. That check stopped at the ASCII controls and the JDK's stops at `U+00FF`, so a token carrying an emoji passed construction and was refused by `HttpRequest.Builder.header` instead -- the same leak through the same channel, one character outside the range being checked. The range is now printable ASCII, narrower than the JDK's own `U+00FF` and deliberately so. A JSON serializer that escapes non-ASCII -- Python's default among them -- renders a token as `Bearer abc\u00e9SECRET` in an echoed body, where neither redaction pass reaches it: the literal is not present in that form and the pattern stops at the backslash. Closing that by unescaping, redacting and re-escaping would put more machinery in the one control that must not be subtly wrong, so the range is refused instead and the case disappears. A token outside printable ASCII is not a shape any deployment sends; the punctuation that does occur, a colon among it, still passes. Within that range the accepted set is the JDK's, and rather than restate another component's private rules construction also asks it directly: it builds the header it would send, discards it, and reports a refusal with no cause and no value. `0xFF` is pinned as still accepted, since that is where the JDK actually stops.

- **The shared storage-endpoint gate had it too.** `URI.create` quotes its whole input in its message and again in the nested `URISyntaxException`, and that exception was chained -- so a malformed endpoint carrying userinfo or a signed query put the secret into the chain a validation failure logs, on the one path where the value was never parseable enough to reach the userinfo check that exists to refuse it. The parse position is kept, the cause is dropped, and a test walks the whole chain.

- **A parser exception is never chained, on any path.** Jackson quotes the offending input in its message, so chaining a parse failure carried whatever the server echoed into the cause chain that `SourceCatalogCredentialVendor` logs. Withholding it only where the body was withheld was not enough: the recipient token that authorises the credential endpoint authorises the listing and metadata endpoints too, and those ask for the body. A redacted snippet is appended where the caller allows a body -- and the comment here claimed that already happened through `snippet()` while nothing called it, so a malformed 200 reached an operator as a parse position and nothing else, with the message hidden again at the gRPC boundary. The NDJSON path does the same, which it did not at first -- and that is the path a proxy error page arriving with a 200 actually lands on. It is withheld on the credential route, the one route that suppresses bodies: redaction covers the token this client sent, not the keys a credential response carries.

- **The redacted snippet names the token this client sent rather than guessing its shape.** The shared helper matched `Bearer` followed by the RFC token68 alphabet, which is what a bearer token is supposed to be. A recipient token need not be one: a header value permits characters the grammar does not, so a token carrying a colon matched only as far as it and the remainder reached the message -- in exactly the echoed-header case the control exists for. The client now passes the value it holds, and the literal is removed before the pattern runs, because the pattern rewrites the part that does match and so destroys the literal that would then be searched for. The token's JSON-escaped form is redacted alongside its raw one, because that is the form a body carries: a server echoing the header into JSON escapes the quotes and backslashes in it, both legal in a header value, so such a token survived the literal search -- and the pattern too, a quote being outside the token68 alphabet. The generic pattern stays as the backstop for a token the caller does not name -- and the Unity client, which is where this control started, is wired to the by-value form too. It had been left on the pattern alone: `bounded("Bearer abc:SECRET", 100)` answers `Bearer <redacted>:SECRET`, and a Unity bearer token is operator-supplied and under no obligation to match the grammar. Reaching it meant a `default Optional<String>` on `UnityCatalogAuthentication` for a fixed secret worth redacting, an override in the `bearer(token, ...)` factory, and nine methods in the client losing `static` so they can read it. The OAuth factory deliberately returns empty: reading a rotating token there means asking the provider for a value on a failure path, and a rotating access token is issued in the token68 shape the pattern already covers. The escaped renderings are covered too, the solidus most of all: `/` is inside the token68 alphabet and common in real tokens, so `abc/SECRET` echoed as `abc\/SECRET` was missed by the literal search and cut at the backslash by the pattern, leaving most of the token in the message. Its comparison de-escapes both sides the same way, which it did not at first: stripping backslashes from the body alone destroyed the evidence for a secret that contains one, so a token `\SECRET` echoed as a backslash-u escape stripped to `SECRET`, was compared against the raw `\SECRET`, matched nothing, and stayed reconstructable in the snippet -- while the pattern missed it too, the token beginning with a backslash. Behind those named forms is a check that fails closed -- if any de-escaping of the body would still reveal the secret, the body is withheld rather than published half-redacted. It de-escapes repeatedly rather than once, up to four layers, stopping as soon as a pass changes nothing: a doubly-escaped rendering survived a single pass, and a check that stops short of what the eventual reader will decode publishes the thing it was added to withhold. Reaching the layer bound withholds the body as well, so the bound fails closed too. It runs before the generic pattern, because that pattern rewrites the matching part of a token and would destroy the literal being looked for: the same ordering mistake as bounding before redacting, one layer further in. Redaction also had to move ahead of the character bound. Bounding first cut a token in half before anything looked for it, and the halves were covered by neither pass: the literal no longer matched, and the pattern stops at the first character outside the grammar, so a colon-bearing token echoed near the 2000-character bound left everything between the colon and the cut in the message. The whole body is scanned rather than a window around the cut, because a replacement is shorter than what it replaces and so pulls later text forward into the bounded range -- text a window would not have reached, and that can hold another occurrence.

- **The parser is strict where its defaults undid the guards.** Every shape check assumes the tree it inspects is what the server sent, and two Jackson defaults broke that before any check ran. `readTree` reads the first document and discards what follows, so `{"items":[]}{"items":[...]}` parsed as an empty page; and the last duplicate key wins, so a body naming tables and then restating `items` as empty parsed as empty. Both arrive looking well formed, and both land on the destructive side, since an empty inventory with no recorded skip is what the reconciler reads as the share having dropped its tables. `FAIL_ON_TRAILING_TOKENS` and `FAIL_ON_READING_DUP_TREE_KEY` are enabled, checked against the databind version this build resolves. The NDJSON path reads a line at a time, so a line carrying two concatenated actions is refused rather than half read.

- **One protocol action and one metaData action.** A second action of either kind silently overwrote the first, so a concatenated or malformed response materialised the later schema and location on a table that reconciled clean -- wrong data on a live table rather than a skipped one. That is the last-write-wins trap `FAIL_ON_READING_DUP_TREE_KEY` closes for duplicate keys, one level up at the action level, where multi-line framing is legal and so that setting does not reach it. Order is deliberately not enforced: the wire format states protocol then metaData, but accepting the reverse costs nothing and refusing it would fail a server that is otherwise correct.

- **The metaData action's scalars are held to the same rule as its arrays.** `version` is published as the `delta.sharing.version` property and `asLong` answers zero for a string or a container, so an unchecked malformed version is a table definition nobody sent rather than a failure; a `configuration` is a map of strings in this protocol: a stated value that is not an object leaves no properties at all, and a number, a boolean or a null coerces through `asText`, so the table carries metadata the server did not send. An object with string values, or nothing. The two `sendJson` overloads went with them -- nothing called either once every route read its body through `sendForBody`, `sendForBodyAndVersion` or `sendNdjson`, and a fourth body-reading path in a class whose argument is that framing and redaction must not drift between paths is surface worth removing.

- **An unrecognised access mode is a stated mode.** A value that is neither `url` nor `dir` was dropped, so `["dir2"]` decoded to the empty list -- the same value that means the server said nothing. Under `delta.sharing.strict-access-modes` such a table was then refused with a message saying it had stated no access modes, which is the misdiagnosis this PR removed for the dir-with-no-location case. An element that is not text is refused outright rather than mapped to `OTHER`, because `OTHER` beside a readable value runs the unsafe way: `[{"a":1}, "dir"]` decoded to `[OTHER, DIR]`, `contains(DIR)` answered true, and the table was treated as directory-accessible on a response this client could not read. An unrecognised *string* decodes to `OTHER`, which keeps the original intent, since a list holding `dir` and something unknown still holds `dir`, and the refusal for a table stating only unknown modes says that rather than claiming it offers url access. Both paths say it: validation's sampler sends names straight from `listTables` to the vend without loading them, so the vend's message is the one an operator running a validation actually sees.

- **A cursor has to be a scalar, an action has to be an action, a format provider has to be a name, an access mode has to be a mode, a stated field has to be a value, and only an empty cursor ends a listing.** `nextPageToken` was read with `asText("")`, which answers the empty string for a stated object or array, and an empty cursor reads as "no more pages" -- so a malformed cursor after a legitimate first page reported one page as the complete inventory, which is again what makes the reconciler retire the rest. A container is refused. Absent and null still mean the listing is done, since null is the ordinary idiom for it, and a numeric cursor is still accepted. `isBlank` was also the wrong test for the end of pagination: a cursor is opaque, the protocol ends on an absent, null or empty token and nothing else, so a whitespace token returned the first page as the complete authoritative inventory -- the retirement shape again. It needed its own encoder too, because `encode` refuses a blank string as an unusable path segment and would have raised an unclassified `IllegalArgumentException` out of a listing rather than send the token back. The metadata actions had the same hole in a third spelling: `has()` answers true for an explicit null and `parseProtocol` tolerates a null node, so `{"protocol":null}` satisfied the both-actions guard with a protocol action this client had invented -- a defaulted `minReaderVersion` and no reader features, for a body that stated neither. `hasNonNull` now, which is the test `parseCredentials` already used. `readerFeatures` was the fifth array field in that parser and the only one not held to the shape rule, which was easy to miss because nothing reads what it produces -- the `Protocol` record is written and never consulted, so a feature-negotiation reader added later would have treated an unsupported feature as absent. `minReaderVersion` is checked as a version too, accepting a number or a numeric string, since refusing `"2"` would fail a working share over a field no consumer reads. The credential envelope is also held to exactly one cloud rather than at least one: the decode takes the first branch that matches, so a response naming two silently became whichever is tested first, making the answer a property of the decoder's ordering rather than of the response. A stated field has to be a value, but an explicit null is not stated: `requireArrayWhereStated` refused a null array, and that refusal is `INVALID_RESPONSE`, which ends the whole reconcile pass rather than skipping one table -- over the rendering a serializer emits routinely for an absent list, and one every other optional field in this parser already reads as absent. Null counts as unstated at all four array sites now. The page's own `items` field still refuses it, because there an empty inventory is what the reconciler retires against.

- **A redirect is a wrong base URI, not a table without directory access.** The transport folds a refused 3xx into `INVALID_REQUEST`, which is also the protocol's own per-table refusal, and the credential path read every `INVALID_REQUEST` as the latter. An auth proxy or a trailing-slash redirect therefore reported each table as offering no directory access: a configuration error arriving table by table as a capability limit, leaving the integration looking partly healthy rather than misconfigured. The status is already on the exception, so only a 400 carries that meaning. The transport folds 400, 405 and 422 into one failure, and a 405 is a GET-only proxy or a gateway rule while a 422 is neither -- read as capability limits, they reported every table as lacking directory access while the real fault went unnamed. All of them stay a per-table skip, since `INVALID_CONFIGURATION` also describes one branch; what changes is the issue an operator is shown.

- **One location, whichever path asks.** The load resolved it listing-first while validation and the vend read the metaData action, so a server stating a location on both surfaces that differ reconciled the table at one, probed at another, and compared the credential's scope against that other -- a passing `CIVCT_STORAGE_ACCESS` check saying nothing about where a read would go. All three now agree on the metaData action. Aligning there rather than on the listing is what makes it three paths rather than two: the vend cannot see a listing without paging the schema on every read, the cost that path was changed to remove, and validation is asserted not to page it either.

- **Access modes are read from both surfaces, and a refusal on either is decisive.** The load preferred the listing and the vend reads the metaData action alone, so a server saying `dir` on one and `url` on the other loaded here and was refused there, on every read, for the life of the table. The load cannot be more permissive than the vend, and that had one surface pairing left: `statedModes` prefers the listing, so a table stating `dir` only there took the load's early return unprobed while the vend, reading the metaData action alone, saw nothing stated and asked -- so a server advertising `dir` on its listing and then refusing the credential endpoint reconciled and failed every read. Whether to probe is now decided on the metaData action, the surface the vend actually reads, rather than by giving the vend a listing it cannot afford. What both paths still share is trusting a `dir` the metaData action does state, without probing it: a stated `dir` followed by a refusal is a server contradicting itself, where an unstated table refusing is the ordinary case the probe exists for, and probing regardless would cost a credential request per table on the common shape. That is a shared assumption rather than a disagreement, and it is written down as one. The opposite disagreement is refused too, which costs only a table nothing could have read this way. Separately, an unstated table is now asked about whether or not a location is known: a location does not establish directory access -- the protocol lets a url-only server state one -- so a legacy server omitting its modes, naming a location and refusing the credential endpoint used to reconcile a table whose every read then failed.

- **The listing bounds are now a memory bound, which they were not.** A page cap of ten thousand and a per-response cap of thirty-two mebibytes multiply out past any heap, and the endpoint is named by tenant configuration, so a server answering enormous pages took the process down rather than failing its own reconcile. `floecat.delta-sharing.max-listing-bytes` bounds what one client pulls in total, defaulting to 32 MiB. Per client rather than per listing: a client's life is one reconcile pass, and a pass memoises each schema's tables for the whole of it, so a share of two hundred schemas would otherwise retain two hundred times the cap. The size is set from what a listing costs -- a table entry is around three hundred bytes of JSON, nearer six hundred with a deep prefix -- so it admits above sixty thousand tables across a share where a large real share holds thousands, and a pass peaks near twice that with the decoded records. Exceeding it is `INVALID_RESPONSE`, which reaches the catalog client as `INTERNAL`: outside `describesOneBranch`, so it ends the pass instead of being recorded as one skipped schema, and not shaped like something to retry. Items are also decoded as each page arrives rather than after the last one, so a listing no longer retains raw pages including the fields this client does not read, and a malformed item fails on the page it is on rather than after fetching every remaining one.

- **A malformed listing is a failure, not an empty share.** `paginate` read any non-array `items` as though the field were absent, so a malformed 200 from `/shares`, `/schemas` or `/tables` became a successful empty inventory. `retireStaleRelations` reads an empty inventory with no recorded skip as the share having dropped its tables and retires the overlay's, so the shape that should have failed loudly would instead have deleted what a classified failure preserves. A page must be an object, and `items` an array where it appears. A required string must be a string, not merely coercible: `schemaString` is persisted as the table's Delta schema without being parsed here, and the credential's location and its three AWS fields are published as a tuple, so `17` reaching any of them is a table or a credential that fails at first use rather than at this boundary. A name must be a string, too: `asText` answers `"17"` for a number and `"true"` for a boolean, so `{"name": 17}` entered the inventory as a table named `17` -- and an inventory the reconciler treats as authoritative is one it retires against, so the tables that malformed page omitted go with it. An explicit `null` is present and is not an array, so it is refused with the rest; exempting it left one malformed shape still reading as an authoritative empty inventory. Absent and empty both stay legal and mean the same thing. The fields inside an item were not held to that rule, and iterating a scalar node yields nothing, so `"auxiliaryLocations": "s3://other/part"` read as absent -- the one shape the auxiliary refusal exists to catch, which would have reconciled with a credential covering the root alone and failed at scan time on the files it does not reach. A stated `accessModes` scalar read as unstated the same way. `partitionColumns` is held to it too, and that is the one that persists: it becomes `CatalogTable.partitionKeys` and then `UpstreamRef.partition_keys`, so a scalar read as absent reconciled a partitioned table as unpartitioned -- a wrong table definition that validates clean rather than a failure. A non-text element is refused with it, since `asText` answers an object with the empty string and an empty partition key is worse than none; absent still means unpartitioned. All of them now go through the guard the page shape already used, and so do array elements: `asText` answers an object with the empty string, so `auxiliaryLocations` stated as an array of objects read as an empty list, `hasAuxiliaryLocations()` answered false, and the refusal at the load never fired, and that guard names the listing it refused. `INVALID_RESPONSE` reaches the boundary as `INTERNAL`, whose message `GrpcErrors` hides, so the log line is the only place an operator learns which schema held the malformed entry -- and it was the one refusal in this transport that said only what was wrong, not where.

- **Auxiliary locations are refused from either surface.** The protocol states them on the table listing as well as the metaData action, and says a client that cannot read one should fall back to url access or fail the request. Reading only the metadata action let a listing-only shape reconcile and then fail at scan time on a file the root credential does not reach.

- **The shared gates are tested where they live.** `core/http-endpoint-guards` shipped with no tests of its own, so every invariant in it was covered incidentally from whichever consumer happened to exercise the branch, and `mvn -pl core/http-endpoint-guards test` ran nothing. That is the argument the module was extracted on, applied to its evidence rather than its code. It now has a test class covering the zone-scoped literal, the all-numeric host, the HTTPS gate, the userinfo refusal and the redaction ordering. The ordering one matters most: that control was broken and repaired twice while this PR was in review, and both times the failing test lived in a different module.

- **`delta-common` publishes two members, not ten.** Interface members are implicitly public, so making the probe a public interface published its S3 client, its listing bound and its location helpers as module API. The contract is `validate` plus the `s3(subject)` factory; everything else is package-private in `S3DeltaLogProbe`, which the same-package test reaches without any of it being public.

- **Reader features are validated before they reach the header.** The list is interpolated into `delta-sharing-capabilities`, so a value carrying a semicolon appends a capability field of its own: `x;responseformat=parquet` makes a server answer in the flat shape, which this client then parses without complaint. Letters, digits, hyphens and underscores only.

- **Identity uses the listing's table id or none, and its components are escaped.** The protocol treats a share id and a table id as opaque strings, so joining them on a bare delimiter made `a:b` plus `c` indistinguishable from `a` plus `b:c`. The metaData action's id is not a substitute for a missing listing id: `PROTOCOL.md` defines it as the unique identifier for the table itself, so every share entry pointing at one physical table reports the same value, and a share exposing a table under two names would have produced one stable identity twice. Both shapes end at `requireUniqueStableIdentities`, which abandons the whole overlay on a duplicate -- the failure the composite exists to prevent. Without a listing id the identity is an honest unstable name.

- **Reader-feature enforcement is the server's, and that is documented rather than implemented.** `delta.sharing.reader-features` reaches the server in `delta-sharing-capabilities` on the metadata call and is not compared against the `readerFeatures` the protocol action reports back. Under directory access the data read goes straight to object storage with the vended credential, so there is no later point at which the server could refuse -- which means a server ignoring the header is not caught here. Checking it locally would mean refusing tables on a knob whose purpose is to be conservative, and is a policy this does not take.

- **A response this client cannot read fails the reconcile rather than skipping a table.** `INVALID_RESPONSE` mapped alongside `INVALID_REQUEST` onto `INVALID_CONFIGURATION`, which `CatalogTraversalFailures.describesOneBranch` treats as per-branch, so a proxy error page or a version mismatch reshaping every table would be absorbed as per-table skips and leave a broken share looking partially healthy across repeated cycles. It now maps to `INTERNAL`, which is outside that predicate, matching the Unity provider. The cost is that `GrpcErrors` hides an `INTERNAL` message, so some detail is lost at the boundary; the protocol's own per-table refusal, `INVALID_REQUEST`, stays per-branch.

- **The load takes the location a credential names as the table's root, whatever shape its path has: the protocol answers a request naming no location with the table's main location, and a Delta table may sit at a bucket root, which the probe supports on purpose. A refusal based on the shape of the path would skip tables validation reads happily -- it did, and the two paths disagreed. Whether the scope matches where the table lives is settled by reading the Delta log under it. The vend also refuses a credential scoped somewhere it cannot read: `requireServable` ran only at the load, so on the reference-server shape -- no location on the metaData action, so the coverage check is skipped -- whatever the credential endpoint returned was published unexamined, and a table reconciled from an earlier good credential could start receiving a tuple scoped to a `gs://` location. And the vend refuses a credential that does not reach the table, and leaves a broader one alone.** A scope above the table -- a share root, a bucket, an external-location prefix -- is published as the server named it. `StorageLocations` states the rule its own `covers` sits under: never be stricter than the component that will actually use the credential, because this path is reached only once no storage authority matched, so a refusal here fails a read that the catalog's own client would have attempted and that S3 would have answered on the real grant. An earlier revision of this branch did reject an ancestor; that would have failed every share whose server scopes credentials to a prefix rather than a table, which is how they are commonly scoped.

- **A malformed credential response ends the pass, and the guard that said otherwise could not fire.** The load also carried a check for a blank location coming back from the credential call, refusing it as a per-table skip. It was unreachable: the transport already refuses a credentials response naming no location, as `INVALID_RESPONSE`, before any caller sees it. Dead either way, but the shape of it was the problem -- it named a graceful per-table refusal for a case that in fact ends the whole reconcile, since `INVALID_RESPONSE` reaches the client as `INTERNAL` and `INTERNAL` is outside `describesOneBranch`. That is the intended rule and it is not being changed here: a response shape this client cannot read is not a property of one table, and absorbing it per table leaves a broken share looking partly healthy. The check is gone and the reasoning is stated where it was. Nothing tested it, which is what unreachable code looks like.

- **Auxiliary locations stated only on the listing are caught at the load, not the vend.** The vend would need a schema listing per resolve to see them, which is what the location change just removed for the same reason. The table never reconciles, so nothing materializes with a root-only credential; what remains is that validation's sampler sends table names straight from `listTables` to the vend without loading them, so an Integration can report valid while that table is later skipped at reconcile. That is a sampler shape rather than a provider one, and `objects_skipped` shows it.

- **Share identity falls back to the share name, and that is deliberate.** When a table listing omits `shareId` the identity is `shareName:tableId`, reported unstable. `listShares` decodes a share `id` that the protocol makes server-unique and immutable, so caching `/shares` could recover stability -- but only for a server that omits `shareId` on tables while populating `id` on shares, a shape there is no evidence of, at the cost of an extra listing on the load path. Left as the unstable fallback rather than built on a guess.

- **One certificate now covers both TLS endpoints in the smoke stack**, so `prepare_unity_tls` became `prepare_smoke_tls`. The rename is mechanical; the SAN gains `delta-sharing-stub` and the gate widens to either scenario being enabled.

## Additional Notes

Not in scope, deliberately:

- **`url` mode.** It needs a per-file URL path that the storage contract cannot express, and a refresh story on top (`minUrlExpirationTimestamp`, `refreshToken`).
- **Per-table access mode in `ListUpstreamObjects`.** That RPC is documented as a lightweight name listing and `CatalogClient.listTables` returns names only, so surfacing it means a new proto field and a new SPI method for one protocol's benefit. Validation reports the boundary instead, naming the table and the reason. The modes do not reach the reconciled table: `CatalogOverlayReconciler` builds a table's properties from the Integration and Overlay identities and the storage location, and does not carry the upstream properties a provider reports. That is true of every provider rather than something this work introduces, so it is a reconciler change and not this PR's business; `docs/operations.md` states it.
- **Change data feed**, and any reader feature beyond what an operator opts into.

`catalog-access/delta-common` is named for what it should accumulate rather than the one class it holds. The substance still sits in `core/connectors/common` -- `DeltaSchemaMapper` and `DeltaSchemaNormalizer`, which the Unity provider already depends on and which a shared table's `schemaString` feeds into downstream. Moving those is gated on the connector retirement, not on this work.

### DIFF
--- a/catalog-access/delta-common/pom.xml
+++ b/catalog-access/delta-common/pom.xml
@@ -16,7 +16,7 @@
     <relativePath>../..</relativePath>
   </parent>
 
-  <artifactId>floecat-catalog-access-unity</artifactId>
+  <artifactId>floecat-catalog-access-delta-common</artifactId>
 
   <dependencies>
     <dependency>
@@ -25,19 +25,8 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
-      <groupId>ai.floedb.floecat</groupId>
-      <artifactId>floecat-http-endpoint-guards</artifactId>
-      <version>${project.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>ai.floedb.floecat</groupId>
-      <artifactId>floecat-unity-catalog-client</artifactId>
-      <version>${project.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>ai.floedb.floecat</groupId>
-      <artifactId>floecat-catalog-access-delta-common</artifactId>
-      <version>${project.version}</version>
+      <groupId>software.amazon.awssdk</groupId>
+      <artifactId>s3</artifactId>
     </dependency>
     <dependency>
       <groupId>org.junit.jupiter</groupId>
--- a/catalog-access/delta-common/src/main/java/ai/floedb/floecat/catalog/delta/DeltaLogStorageProbe.java
+++ b/catalog-access/delta-common/src/main/java/ai/floedb/floecat/catalog/delta/DeltaLogStorageProbe.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2026 Yellowbrick Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ */
+
+package ai.floedb.floecat.catalog.delta;
+
+import ai.floedb.floecat.catalog.access.VendedStorageCredentials;
+import java.util.Locale;
+import java.util.Objects;
+
+@FunctionalInterface
+public interface DeltaLogStorageProbe {
+
+  void validate(String tableLocation, VendedStorageCredentials credentials);
+
+  /**
+   * Whether the S3 probe can address this location at all.
+   *
+   * <p>Part of the contract rather than an internal: a provider that vends AWS credentials needs to
+   * know before it reconciles a table whether the location it was given is one this can read. An
+   * {@code abfss://} or {@code gs://} table would otherwise load, reconcile, and fail at every
+   * scan.
+   */
+  static boolean s3Serves(String tableLocation) {
+    return S3DeltaLogProbe.s3Location(tableLocation) != null;
+  }
+
+  /**
+   * The form of a location a provider should publish, vend and compare.
+   *
+   * <p>An S3 object key may hold a space and {@code java.net.URI} may not. This probe encodes one
+   * before parsing, so it answers for {@code s3://bucket/db/my table} -- and a provider that then
+   * publishes the raw string reconciles a table the read path cannot open, because that path calls
+   * {@code URI.create} on the location at every site. Before the probe accepted such a location it
+   * was refused outright, which was at least a recorded failure; accepting it on one side without
+   * canonicalising on the other turns that into a table that validates clean and fails every scan.
+   *
+   * <p>Here rather than in either provider because both need the same answer, which is the reason
+   * this module exists. Encoded rather than refused because the read path derives its key with
+   * {@code URI.getPath}, which decodes, so S3 is asked for the key that was written. Idempotent,
+   * since only a literal space is replaced: a location that already carries {@code %20} is
+   * unchanged and nothing is double-encoded.
+   *
+   * <p>Other characters {@code URI} rejects and S3 permits are still refused by {@link #s3Serves}.
+   * A space is the one that occurs.
+   */
+  static String canonicalLocation(String tableLocation) {
+    if (tableLocation == null) {
+      return null;
+    }
+    // The scheme too, not only the space. s3Serves accepts s3, s3a and s3n in any case, while the
+    // reader's resolvePath matches a literal lowercase s3:// or s3a:// and throws on anything
+    // else -- so publishing the spelling the server used meant s3n://bucket/table and
+    // S3://bucket/table validated, reconciled, and failed every scan on an unsupported path.
+    // Servability and publication have to agree about what a location is, and the reader decides.
+    // No trim. An S3 object key may end in a space, and the probe does not trim either -- so
+    // trimming here published a different prefix from the one validation had just read, which is
+    // the divergence this method exists to prevent. Whatever the server sent is the key; the only
+    // changes made to it are the ones the reader requires.
+    return publishableScheme(tableLocation).replace(" ", "%20");
+  }
+
+  /**
+   * The scheme folded only where the reader cannot parse it.
+   *
+   * <p>{@code StorageLocations.normalizeScheme} folds s3, s3a and s3n alike, and folding everything
+   * to s3 put the stored location in a different scheme namespace from the prefix an operator
+   * registers: {@code StorageAuthorityResolver.matchesLocationPrefix} is a scheme-sensitive {@code
+   * startsWith}, and {@code S3Location.parse} explicitly accepts an {@code s3a://} authority prefix
+   * -- so a catalog reporting s3a, an authority registered as s3a, and a location stored as s3
+   * never match.
+   *
+   * <p>So only the spellings the reader genuinely cannot read are changed: {@code s3n} and any
+   * uppercase form, since {@code S3V2FileSystemClient.resolvePath} matches a literal lowercase
+   * {@code s3://} or {@code s3a://}. What the source reported is otherwise what gets stored.
+   */
+  private static String publishableScheme(String location) {
+    int schemeEnd = location.indexOf("://");
+    if (schemeEnd < 0) {
+      return location;
+    }
+    String scheme = location.substring(0, schemeEnd).toLowerCase(Locale.ROOT);
+    if ("s3a".equals(scheme)) {
+      return scheme + location.substring(schemeEnd);
+    }
+    return "s3".equals(scheme) || "s3n".equals(scheme)
+        ? "s3" + location.substring(schemeEnd)
+        : location;
+  }
+
+  /**
+   * A probe that lists and reads one object under the table's Delta log with the vended
+   * credentials.
+   *
+   * <p>Listing alone would pass on a grant that cannot read, so the probe reads a byte. What it
+   * proves is that the credential the provider vended actually reaches the storage it named, which
+   * is the question a validation run is asking on the operator's behalf.
+   *
+   * @param subject how the catalog is named in a refusal, such as {@code "Unity Catalog"}
+   */
+  static DeltaLogStorageProbe s3(String subject) {
+    Objects.requireNonNull(subject, "subject");
+    return (tableLocation, credentials) ->
+        S3DeltaLogProbe.validate(subject, tableLocation, credentials);
+  }
+}
--- a/catalog-access/delta-common/src/main/java/ai/floedb/floecat/catalog/delta/S3DeltaLogProbe.java
+++ b/catalog-access/delta-common/src/main/java/ai/floedb/floecat/catalog/delta/S3DeltaLogProbe.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  */
 
-package ai.floedb.floecat.catalog.unity;
+package ai.floedb.floecat.catalog.delta;
 
 import ai.floedb.floecat.catalog.access.CatalogAccessException;
 import ai.floedb.floecat.catalog.access.StorageLocations;
@@ -33,35 +33,30 @@ import software.amazon.awssdk.services.s3.model.GetObjectRequest;
 import software.amazon.awssdk.services.s3.model.ListObjectsV2Request;
 import software.amazon.awssdk.services.s3.model.S3Exception;
 
-@FunctionalInterface
-interface UnityStorageAccessValidator {
+final class S3DeltaLogProbe {
 
-  /** Implicitly public static final: this interface is package-private, so the constant is too. */
-  String DELTA_LOG_DIR = "_delta_log/";
+  private S3DeltaLogProbe() {}
 
-  void validate(String tableLocation, VendedStorageCredentials credentials);
+  private static final String DELTA_LOG_DIR = "_delta_log/";
 
-  static UnityStorageAccessValidator s3() {
-    return UnityStorageAccessValidator::validateS3;
-  }
-
-  private static void validateS3(String tableLocation, VendedStorageCredentials credentials) {
+  static void validate(String subject, String tableLocation, VendedStorageCredentials credentials) {
     // Parsed once, and inside the guard. Doing it here unguarded put an IllegalArgumentException
     // ahead of the check meant to catch it: an object key holding a character URI rejects -- a
-    // space is the common one -- threw before s3Location could answer, and UnityCatalogErrors
-    // mapped it to "storage access validation configuration is invalid", which describes neither
-    // the location nor anything an operator configured.
+    // space is the common one -- threw before s3Location could answer, and the caller mapped it
+    // to "storage access validation configuration is invalid", which describes neither the
+    // location nor anything an operator configured.
     URI location = s3Location(tableLocation);
     if (location == null) {
       throw new CatalogAccessException(
           CatalogAccessException.Code.UNSUPPORTED,
-          "Unity Catalog storage validation needs an addressable S3 location, and could not read"
-              + " a bucket from this one");
+          subject
+              + " storage validation needs an addressable S3 location, and could not read a"
+              + " bucket from this one");
     }
     String bucketName = bucketOf(location);
     Map<String, String> properties = credentials.properties();
-    String accessKey = required(properties, "s3.access-key-id");
-    String secretKey = required(properties, "s3.secret-access-key");
+    String accessKey = required(subject, properties, "s3.access-key-id");
+    String secretKey = required(subject, properties, "s3.secret-access-key");
     String sessionToken = nonBlank(properties.get("s3.session-token"));
     AwsCredentials awsCredentials =
         sessionToken == null
@@ -83,10 +78,10 @@ interface UnityStorageAccessValidator {
       builder.endpointOverride(URI.create(endpoint));
     }
 
-    // The bucket named in the object URI, never s3.access-point. SourceCatalogCredentialVendor
-    // strips that key before any reader sees it -- noteIgnoredAccessPoint records why -- so probing
-    // the access point would validate an endpoint no scan ever addresses, and a grant scoped to it
-    // alone would report success against reads that all get 403.
+    // The bucket named in the object URI, never s3.access-point. The vend strips that key before
+    // any reader sees it, so probing the access point would validate an endpoint no scan ever
+    // addresses, and a grant scoped to it alone would report success against reads that all get
+    // 403.
     String bucket = bucketName;
     String logPrefix = deltaLogPrefix(location.getPath());
     try (S3Client s3 = builder.build()) {
@@ -95,8 +90,8 @@ interface UnityStorageAccessValidator {
               ListObjectsV2Request.builder().bucket(bucket).prefix(logPrefix).maxKeys(1).build());
       if (listing.contents().isEmpty()) {
         throw new CatalogAccessException(
-            CatalogAccessException.Code.UNSUPPORTED,
-            "Unity Catalog table storage contains no Delta log object to validate");
+            CatalogAccessException.Code.INVALID_CONFIGURATION,
+            subject + " table storage contains no Delta log object to validate");
       }
       try {
         // Streamed and aborted, not buffered. Range is a request hint: an S3-compatible endpoint
@@ -136,29 +131,27 @@ interface UnityStorageAccessValidator {
           failure.awsErrorDetails() == null ? null : failure.awsErrorDetails().errorCode();
       throw new CatalogAccessException(
           storageFailureCode(errorCode, failure.statusCode()),
-          "Unity Catalog storage validation failed",
+          subject + " storage validation failed",
           failure);
     } catch (java.io.IOException failure) {
       // Reading or releasing the probe body: a fact about reaching the store, like the transport
       // failures below it, not about the credential.
       throw new CatalogAccessException(
           CatalogAccessException.Code.UNAVAILABLE,
-          "Unity Catalog storage validation could not read the probe object",
+          subject + " storage validation could not read the probe object",
           failure);
     } catch (RuntimeException failure) {
       throw new CatalogAccessException(
-          CatalogAccessException.Code.UNAVAILABLE,
-          "Unity Catalog storage validation failed",
-          failure);
+          CatalogAccessException.Code.UNAVAILABLE, subject + " storage validation failed", failure);
     }
   }
 
-  private static String required(Map<String, String> properties, String key) {
+  private static String required(String subject, Map<String, String> properties, String key) {
     String value = nonBlank(properties.get(key));
     if (value == null) {
       throw new CatalogAccessException(
           CatalogAccessException.Code.INVALID_CONFIGURATION,
-          "Unity Catalog storage credentials omitted " + key);
+          subject + " storage credentials omitted " + key);
     }
     return value;
   }
@@ -182,7 +175,7 @@ interface UnityStorageAccessValidator {
    * <p>Scoped to the listing on purpose. The ranged read streams a body of unknown size and aborts
    * after one byte, so capping that stream would constrain a path whose bound already holds.
    */
-  final class BoundedListingBody implements ExecutionInterceptor {
+  private static final class BoundedListingBody implements ExecutionInterceptor {
 
     @Override
     public Optional<InputStream> modifyHttpResponseContent(
@@ -299,11 +292,28 @@ interface UnityStorageAccessValidator {
     }
     URI normalized;
     try {
-      normalized = URI.create(StorageLocations.normalizeScheme(tableLocation));
+      // A space is percent-encoded before parsing. It is legal in an S3 object key and illegal in
+      // java.net.URI, so a valid table location answered null here and the caller refused it for
+      // having no addressable bucket -- a diagnosis pointing at the bucket when the key was the
+      // problem. getPath decodes it again, so the key the probe asks for is unchanged. Other
+      // characters URI rejects and S3 permits still answer null; a space is the one that occurs.
+      normalized = URI.create(StorageLocations.normalizeScheme(tableLocation).replace(" ", "%20"));
     } catch (IllegalArgumentException notAUri) {
       return null;
     }
     if (!"s3".equalsIgnoreCase(normalized.getScheme()) || bucketOf(normalized) == null) {
+      return null;
+    }
+    // Nothing that carries a secret. bucketOf refuses an authority holding userinfo, but only on
+    // its fallback branch: getHost() answers "bucket" for s3://user:password@bucket/table, so that
+    // check never ran and the location was servable. A provider then stores it as the reconciled
+    // table's storage_location, which puts a password or a signed query into catalog metadata --
+    // readable by every client that can read the table. None of these belong in an S3 object
+    // location anyway, and the endpoint gates already refuse the same four on a catalog URI.
+    if (normalized.getRawUserInfo() != null
+        || normalized.getRawQuery() != null
+        || normalized.getRawFragment() != null
+        || normalized.getPort() != -1) {
       return null;
     }
     return normalized;
@@ -338,8 +348,20 @@ interface UnityStorageAccessValidator {
     return authority;
   }
 
+  /**
+   * Exactly the URI separator, and no more.
+   *
+   * <p>An S3 object key may begin with a slash, so {@code s3://bucket//table} names the key {@code
+   * /table}. The read path removes one leading slash and addresses that key; this removed every one
+   * and probed {@code table/_delta_log/} instead -- a different prefix, so validation either
+   * rejected a readable table or, where something else sat at the stripped prefix, read that Delta
+   * log and reported success for a location no scan would reach.
+   */
   private static String stripLeadingSlash(String value) {
-    return value == null ? "" : value.replaceFirst("^/+", "");
+    if (value == null) {
+      return "";
+    }
+    return value.startsWith("/") ? value.substring(1) : value;
   }
 
   private static String nonBlank(String value) {
--- a/catalog-access/delta-common/src/test/java/ai/floedb/floecat/catalog/delta/DeltaLogStorageProbeTest.java
+++ b/catalog-access/delta-common/src/test/java/ai/floedb/floecat/catalog/delta/DeltaLogStorageProbeTest.java
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package ai.floedb.floecat.catalog.unity;
+package ai.floedb.floecat.catalog.delta;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
@@ -34,7 +34,10 @@ import org.junit.jupiter.api.Test;
  * smoke failure -- which key prefix the Delta log is looked for under, and what an S3 error means
  * about the credential -- are pure, and asserted here.
  */
-class UnityStorageAccessValidatorTest {
+class DeltaLogStorageProbeTest {
+
+  /** Every refusal names its caller, so the tests assert against one name. */
+  private static final String SUBJECT = "Test Catalog";
 
   /**
    * A table at the bucket root has an empty path, and a leading slash would make the prefix
@@ -52,9 +55,7 @@ class UnityStorageAccessValidatorTest {
             new Case("trailing slash", "/tpch/orders/", "tpch/orders/_delta_log/"),
             new Case("no leading slash", "tpch/orders", "tpch/orders/_delta_log/"),
             new Case("null path", null, "_delta_log/"))) {
-      assertThat(UnityStorageAccessValidator.deltaLogPrefix(c.path()))
-          .as(c.name())
-          .isEqualTo(c.expected());
+      assertThat(S3DeltaLogProbe.deltaLogPrefix(c.path())).as(c.name()).isEqualTo(c.expected());
     }
   }
 
@@ -85,7 +86,7 @@ class UnityStorageAccessValidatorTest {
                 "ExpiredToken",
                 400,
                 CatalogAccessException.Code.CREDENTIAL_EXPIRED))) {
-      assertThat(UnityStorageAccessValidator.storageFailureCode(c.errorCode(), c.status()))
+      assertThat(S3DeltaLogProbe.storageFailureCode(c.errorCode(), c.status()))
           .as(c.name())
           .isEqualTo(c.expected());
     }
@@ -109,7 +110,7 @@ class UnityStorageAccessValidatorTest {
             // Neither belongs in an S3 location, so an authority carrying one is not a bucket name.
             new Case("userinfo", "s3://user@my_bucket/x", null),
             new Case("port", "s3://my_bucket:9000/x", null))) {
-      assertThat(UnityStorageAccessValidator.bucketOf(java.net.URI.create(c.location())))
+      assertThat(S3DeltaLogProbe.bucketOf(java.net.URI.create(c.location())))
           .as(c.name())
           .isEqualTo(c.expected());
     }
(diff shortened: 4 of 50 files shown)


kwesterfeld-floe:
Superseded by a six-PR stack along the same commit lines, so it can be reviewed in small batches. Same tree: `kw-delta-sharing-integration-06` is identical to this branch.

- #522 -- refactor(http): share the catalog endpoint gates and response handling (base `main`)
- #523 -- refactor(catalog-access): add delta-common and move the Delta log probe into it
- #524 -- fix(validation): report an upstream refusal by capability as unsupported
- #525 -- feat(delta-sharing): add the recipient client
- #526 -- feat(delta-sharing): add the catalog-access provider
- #527 -- test(smoke): cover the Delta Sharing vend end to end

Each is parented on the previous one. The full rationale and review history stay here for reference.

PR #520, opened 2026-09-08
Simplify shared cache read-through concurrency

mrouvroy-floe:
## Summary

- Define the shared `MemoryCache<K, V>` contract for immutable, identity-keyed values.
- Use Caffeine’s native coordination for concurrent scalar and batch loads.
- Keep `peek` resident-only: it never loads or records a hit.
- Add weighted byte accounting, W-TinyLFU admission, explicit key/partition eviction, and cache telemetry.
- Keep absence out of the cache so a later durable value can still be observed.
- Restore the cache event callbacks required by the existing pointer-cache telemetry.
- Add deterministic tests for concurrent loads, batch reads, failures, budgets, and eviction.
- Document the shared cache behavior.

## Scope

Eviction is memory hygiene. The generic cache has no application-level generation fence, in-flight map, or version protocol.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist


service/src/main/java/ai/floedb/floecat/service/repo/cache/CachingPointerStore.java
```diff
@@ -129,7 +141,12 @@ public Map<String, Pointer> getBatchConsistent(List<String> keys) {
   @Override
   public List<Pointer> listPointersByPrefix(
       String prefix, int limit, String pageToken, StringBuilder nextTokenOut) {
-    return pointers.list(prefix, limit, pageToken, nextTokenOut);
+    lockAllMutations();
```
mcusack-floe, line 144: This appears to lock every pointer across all accounts, effectively serializing list and count operations.

mcusack-floe (changes requested):

service/src/main/java/ai/floedb/floecat/service/repo/cache/CachingPointerStore.java
mcusack-floe: It seems this is fixed in the next PR above this one.

mrouvroy-floe: Yes, the PR above rewrites the whole pointer mechanism and removes the issue

mcusack-floe (approved):

PR #529, opened 2026-09-09
storage/aws: wait for the DynamoDB endpoint before the table bootstrap

mivashko-floe:
tableExists() is the first DynamoDB call the process makes. Under `make start` floecat-runtime is launched ~100ms after storage-local, whose emulator can take 15-20s to bind :4566 on a cold box, while the SDK's built-in connect retries give up after ~7s -- so the service died at startup with "Connection refused ... :4566 (SDK Attempt Count: 9)", the account service could then never create the control account, and the SQL suite timed out on the planner port.

Retry only while the endpoint does not answer at all (SdkClientException) for up to floecat.kv.bootstrap.connect-wait-seconds (default 90s, matching core's KvServiceDependency). A real service answer -- not-found, auth, throttling -- is an AwsServiceException and still surfaces immediately.

## Summary

## Type of Change

- [x] fix (bug fix)

## Testing

- [ ] `make fmt`
- [] `make verify`

- [x] No breaking changes

## Checklist

## Additional Notes

### DIFF
--- a/storage/aws/src/main/java/ai/floedb/floecat/storage/kv/dynamodb/DynamoDbTablesBootstrap.java
+++ b/storage/aws/src/main/java/ai/floedb/floecat/storage/kv/dynamodb/DynamoDbTablesBootstrap.java
@@ -26,6 +26,7 @@ import java.util.function.Function;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.core.exception.SdkClientException;
 import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient;
 import software.amazon.awssdk.services.dynamodb.model.*;
 
@@ -43,6 +44,9 @@ public class DynamoDbTablesBootstrap implements KvAttributes {
 
   @ConfigProperty(name = "floecat.kv.bootstrap.wait-seconds", defaultValue = "20")
   int waitSeconds;
+
+  @ConfigProperty(name = "floecat.kv.bootstrap.connect-wait-seconds", defaultValue = "90")
+  int connectWaitSeconds;
 
   public void ensureTableExists(String tableName, boolean withTtl) {
     try {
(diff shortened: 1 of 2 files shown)


mcusack-floe (approved):
I think this is ok, but perhaps the condition should be "wait for the dynamo table to be available".

storage/aws/src/main/java/ai/floedb/floecat/storage/kv/dynamodb/DynamoDbTablesBootstrap.java
```diff
@@ -223,6 +252,23 @@ private static boolean isResourceInUse(Throwable t) {
     return false;
   }
 
+  private static boolean isEndpointUnreachable(Throwable t) {
```
mrouvroy-floe, line 264: SdkClientException is the SDK's whole client-side error family, not just "nothing is listening on the port". It also covers failures that will never succeed no matter how long you retry:
- credential resolution — Unable to load credentials from any of the providers in the chain 
- bad floecat.storage.aws.dynamodb.endpoint-override : unparseable URI, wrong scheme
- TLS trust failures against a real DynamoDB endpoint (the compose services run with a custom truststore)
- ApiCallTimeoutException / AbortedException, both subclasses of SdkClientException

So "A real service answer is an AwsServiceException and still surfaces immediately" only works for service rejections, but a whole class of permanent client misconfigurations now converts a fast, clearly-attributed startup failure into a 90-second hang that logs DynamoDB endpoint not reachable yet (Unable to load credentials ...) ~90 times before rethrowing. 

We should the predicate to genuine connect-level root causes, and consider renaming it to match:

```java
private static boolean isEndpointUnreachable(Throwable t) {
  boolean sdkClient = false;
  for (Throwable cur = t; cur != null; cur = cur.getCause()) {
    if (cur instanceof SdkClientException) sdkClient = true;
    if (cur instanceof ConnectException
        || cur instanceof NoRouteToHostException
        || cur instanceof UnknownHostException
        || cur instanceof SocketTimeoutException) {
      return sdkClient;
    }
  }
  return false;
}
```

storage/aws/src/main/java/ai/floedb/floecat/storage/kv/dynamodb/DynamoDbTablesBootstrap.java
```diff
@@ -171,14 +175,39 @@ private void ensureTtlEnabled(String tableName, String ttlAttrName) {
     ddbJoin(client -> client.updateTimeToLive(update));
   }
 
+  // First DynamoDB call the process makes. Under `make start` this service comes up ~100ms
+  // after storage-local, which can take 15-20s to bind its port on a cold box, while the
+  // SDK's own connect retries give up in ~7s -- a cold start used to die right here with
+  // "Connection refused". Retry while the endpoint does not answer at all
+  // (SdkClientException) for up to connectWaitSeconds; an actual answer -- not-found, auth,
+  // throttling -- is an AwsServiceException and still surfaces immediately. 90s matches
+  // KvServiceDependency, which is how every other service waits for the same port.
```
mrouvroy-floe: Nit: KvServiceDependency and storage-local are core only things, they don't exist in floecat OSS so I wouldn't mention it in the code if we can avoid it

mrouvroy-floe (approved):

mivashko-floe:
> I think this is ok, but perhaps the condition should be "wait for the dynamo table to be available".

@markcusack The table-available wait already exists a few lines down: waitUntilTableExists + waitUntilActive under bootstrap.wait-seconds, right after createTable. This loop sits upstream of that - it's the existence check, on purpose

mivashko-floe:
@mrouvroy-floe addressed both comments

PR #530, opened 2026-09-09
test: Optimize integration-test cache resets

mcusack-floe:
## Summary

This speeds up integration-test fixture resets while keeping the pointer cache enabled and exercised. Test cleanup now resets the authoritative store and invalidates cached indexes once, avoiding repeated cache lock/index churn; ConnectorIT then warms its commonly used indexes after seeding.

Fixes the ConnectorIT slowdown introduced by the pointer-cache work: each test fixture reset was routing broad destructive deletes through the cache, repeatedly triggering shard-lock acquisition and complete-index invalidation/rebuilds. Resets now wipe the authoritative store directly, clear cached state once, and warm the indexes used by ConnectorIT after seeding—keeping the cache enabled and exercised for normal test operations.

## Type of Change

- [ ] fix (bug fix)
- [X] test (adds/updates tests)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist

### DIFF
--- a/service/src/main/java/ai/floedb/floecat/service/repo/cache/PointerCache.java
+++ b/service/src/main/java/ai/floedb/floecat/service/repo/cache/PointerCache.java
@@ -558,6 +558,20 @@ public final class PointerCache {
     return accountCount(Readiness.COMPLETE);
   }
 
+  /**
+   * Drops every cached pointer and complete index.
+   *
+   * <p>This is intended for test fixture resets that have cleared the underlying store directly.
+   * The next cached read repopulates the required index from that now-authoritative store.
+   */
+  public void clear() {
+    partitions.clear();
+    completeBytes.set(0L);
+    completeEntries.set(0L);
+    resident.evictPartition(key -> true);
+    resident.maximumBytes(maxBytes);
+  }
+
   public long degradedAccountCount() {
     return accountCount(Readiness.DEGRADED);
   }
--- a/service/src/test/java/ai/floedb/floecat/service/util/TestDataResetter.java
+++ b/service/src/test/java/ai/floedb/floecat/service/util/TestDataResetter.java
@@ -18,10 +18,13 @@ package ai.floedb.floecat.service.util;
 
 import ai.floedb.floecat.service.bootstrap.impl.SeedRunner;
 import ai.floedb.floecat.service.reconciler.jobs.durable.store.MemoryReconcileJobIndexBackend;
+import ai.floedb.floecat.service.repo.cache.PointerCache;
 import ai.floedb.floecat.service.repo.model.Keys;
 import ai.floedb.floecat.storage.spi.BlobStore;
+import ai.floedb.floecat.storage.spi.CachedPointerStore;
 import ai.floedb.floecat.storage.spi.PointerStore;
 import ai.floedb.floecat.storage.spi.PointerStoreKeys;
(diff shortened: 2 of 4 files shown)


service/src/main/java/ai/floedb/floecat/service/repo/cache/PointerCache.java
```diff
@@ -558,6 +558,20 @@ public long completeAccountCount() {
     return accountCount(Readiness.COMPLETE);
   }
 
+  /**
+   * Drops every cached pointer and complete index.
+   *
+   * <p>This is intended for test fixture resets that have cleared the underlying store directly.
+   * The next cached read repopulates the required index from that now-authoritative store.
+   */
+  public void clear() {
```
mrouvroy-floe, line 567: `PointerCache.clear()` currently clears `partitions` and resets accounting without fencing in-flight cache loads. A load that started before `wipeAll()` could finish after the clear and publish its old partition/accounting back into the cache. That makes the “invalidate cached indexes once” behavior not quite reliable under background ConnectorIT/reconciler activity.

service/src/test/java/ai/floedb/floecat/service/util/TestDataResetter.java
```diff
@@ -95,13 +100,27 @@ public void wipeAll() {
         blobs.deletePrefix("/accounts/" + tid + "/");
       }
       blobs.deletePrefix("/accounts/");
+      if (pointerCache != null) {
```
mrouvroy-floe, line 103: NIT: I don't think this can happen: CDI injection would fail at runtime if they were NULL

service/src/test/java/ai/floedb/floecat/service/util/TestDataResetter.java
```diff
@@ -95,13 +100,27 @@ public void wipeAll() {
...
 
       if (!ptr.isEmpty()) {
         ptr.dump("AFTER WIPE, NON-EMPTY");
       }
     }
   }
 
+  /** Warms the indexes that every seeded-account test uses before its first assertion. */
+  public void warmPointerCache(String accountId) {
+    if (cachedPointers == null) {
```
mrouvroy-floe, line 115: NIT: I don't think this can happen: CDI injection would fail at runtime if they were NULL

mrouvroy-floe (commented):
I think the idea is correct, but the shape feels wrong.

Design wise, `TestDataResetter` now has to inject raw pointer store, cached pointer store, and `PointerCache`, then coordinate them in the right order. That makes test reset code know too much about the cache internals.
Could we hide this behind one fixture-reset operation instead? Something like “reset pointer fixtures” owned by the pointer/cache layer ?

Correctness wise,  a cache load that started before reset must not be able to finish after reset and put old cache state/accounting back.  
The next PR in the caching stack adds some shared concurrency model that we can probably reuse here so maybe we should wait for that. Otherwise, we should add some other mechanism here

service/src/test/java/ai/floedb/floecat/service/util/TestDataResetter.java
mcusack-floe: I'll revisit after the next PR in your stack is merged.

mrouvroy-floe:
https://github.com/eng-floe/floecat/pull/532 will refactor the pointer layer and address the issue

PR #532, opened 2026-09-11
Serve planner pointers and listings from the authoritative index

mrouvroy-floe:
## Summary

- Keep all planner-visible pointers for an owned account in an in-memory authoritative index.
- Warm account partitions asynchronously when ownership is acquired.
- Use durable KV while a partition is `LOADING`.
- Once `COMPLETE`, serve planner resolution, batch reads, listings, counts, and authoritative absence from memory.
- Commit durable pointer mutations before publishing them into the index.
- Serialize same-key mutations with per-key locks.
- Serialize listings, prefix operations, account lifecycle, and partition replacement with the account lock.
- Invalidate and reload the partition when ownership changes.
- Keep operational pointer families on the durable path.
- Add account-scoped warm-up metrics, tests, and documentation.

## Result

A complete planner partition is the current in-memory representation of the account’s planner pointers. It is not an evictable generic cache: if it is still loading, reads use KV; once complete, a missing planner key is a real `NOT_FOUND`.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist

## Additional Notes


service/src/main/java/ai/floedb/floecat/service/repo/cache/IndexedPointerStore.java
```diff
@@ -0,0 +1,175 @@
...
+    List<String> keys = ops.stream().map(CasOp::key).distinct().toList();
+    return index.mutateKeys(
+        keys,
+        () -> durable.compareAndSetBatch(ops),
+        won -> {
+          if (!won) {
+            for (String key : keys) index.refresh(key, durable.getConsistent(key));
+            return;
+          }
+          for (CasOp op : ops) {
```
mcusack-floe, line 152: We're applying ops in list order, instead of the old publish before remove (compareAndSetBatch). It seems that we can now delete before upsert so following a rename a reader can see both names absent.

service/src/main/java/ai/floedb/floecat/service/repo/model/Keys.java
```diff
@@ -166,6 +161,40 @@ public static String accountRootPrefix() {
...
+  public static PointerNamespace pointerNamespace(String key) {
+    if (key == null || !key.startsWith(accountRootPrefix())) {
+      return PointerNamespace.UNKNOWN;
+    }
+    String remainder = key.substring(accountRootPrefix().length());
+    int slash = remainder.indexOf('/');
+    String account = slash < 0 ? remainder : remainder.substring(0, slash);
+    if (account.isBlank()) return PointerNamespace.UNKNOWN;
+    if (isReservedAccountDirectorySegment(account)) return PointerNamespace.ACCOUNT_DIRECTORY;
+    if (key.contains(SEG_TRANSACTIONS)
```
mcusack-floe: Classifying by substring using contains will misclassify namepsaces by-path if a namespace segment is called markers, transactions, gc etc. It will get dropped by the load filter. Such namespace paths will be missing from ListNamespaces.

mcusack-floe (changes requested):

service/src/main/java/ai/floedb/floecat/service/repo/cache/IndexedPointerStore.java
mrouvroy-floe: Fixed. Insertions all land before any removal now, on the losing path too — refresh is publish-if-present/remove-if-absent, so it had the same hazard.  Both switches are exhaustive over the sealed CasOp again, so a new variant is a compile error rather than a silent drop.

service/src/main/java/ai/floedb/floecat/service/repo/model/Keys.java
mrouvroy-floe: Good catch, fixed by anchoring on position

mcusack-floe (approved):

PR #534, opened 2026-09-13
catalog: configure integration access delegation

mcusack-floe:
## Summary

Adds a property to allow credential vending requests to be removed from headers. This is needed for Iceberg REST catalogs that do not have credential vending enabled.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [X] `make fmt`
- [X] `make verify`

- [X] No breaking changes

## Checklist


mrouvroy-floe (changes requested):
I think an integration with delegation off reports itself invalid ?

`IcebergRestCatalogClient.capabilities()` is a static set which always includes `VEND_STORAGE_CREDENTIALS` and doesn't know about the new mode. So in `CatalogIntegrationDiscovery.validate`, `canVend` is still true, the vending search runs, and with no delegation header every loadTable comes back without a credentials list. The search walks up to `MAX_VALIDATION_VEND_ATTEMPTS = 8` tables inside the validation budget, finds nothing, and validation ends at:

"The catalog did not vend usable storage credentials for an upstream table." → CIVI_CREDENTIAL_VENDING_FAILED, valid=false

Other smaller issues :
- The remove in the none branch is dead. At that point the key cannot be present: rejectResolvedOnlyProperties throws on any header.* in both property maps, and the resolvedCredentials.headers() loop throws on a header named X-Iceberg-Access-Delegation. 

- The mode is read before the auth properties are merged. properties.remove(ACCESS_DELEGATION_MODE_PROPERTY) runs, then properties.putAll(config.authentication().properties()). If the key ever showed up in the auth map it would be re-inserted and shipped to the Iceberg client as an unknown property, with no effect on the header.

mcusack-floe:
> I think an integration with delegation off reports itself invalid ?
> 
> `IcebergRestCatalogClient.capabilities()` is a static set which always includes `VEND_STORAGE_CREDENTIALS` and doesn't know about the new mode. So in `CatalogIntegrationDiscovery.validate`, `canVend` is still true, the vending search runs, and with no delegation header every loadTable comes back without a credentials list. The search walks up to `MAX_VALIDATION_VEND_ATTEMPTS = 8` tables inside the validation budget, finds nothing, and validation ends at:
> 
> "The catalog did not vend usable storage credentials for an upstream table." → CIVI_CREDENTIAL_VENDING_FAILED, valid=false
> 
> Other smaller issues :
> 
> * The remove in the none branch is dead. At that point the key cannot be present: rejectResolvedOnlyProperties throws on any header.* in both property maps, and the resolvedCredentials.headers() loop throws on a header named X-Iceberg-Access-Delegation.
> * The mode is read before the auth properties are merged. properties.remove(ACCESS_DELEGATION_MODE_PROPERTY) runs, then properties.putAll(config.authentication().properties()). If the key ever showed up in the auth map it would be re-inserted and shipped to the Iceberg client as an unknown property, with no effect on the header.

Addressed these. I also added a fallback to storage authority if credential vending is off, which mirrors current connector behaviour.

service/src/main/java/ai/floedb/floecat/service/integration/CatalogIntegrationAccessDelegation.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package ai.floedb.floecat.service.integration;
+
+import ai.floedb.floecat.integration.rpc.CatalogIntegration;
+import ai.floedb.floecat.integration.rpc.CatalogIntegrationType;
+
+/** Reads credential-vending intent from a Catalog Integration. */
+public final class CatalogIntegrationAccessDelegation {
+  public static final String MODE_PROPERTY = "access-delegation-mode";
```
mrouvroy-floe: requestsVendedCredentials treats any value other than none as "request vending", while IcebergRestCatalogClientProvider.catalogProperties enforces a strict allow-list (vended-credentials / none) and throws on anything else.

A misconfigured mode (for instance access-delegation-mode=remote-signing or a typo like nonee) is silently accepted here and only rejected later when the client is opened, surfacing as an opaque IllegalArgumentException / INTERNAL failure rather than a clear configuration error. The two allowed-value sets can also silently drift as new modes are added.

We should unify mode parsing and validation behind a shared enum so invalid values are caught at configuration load time with a descriptive error, and both readers always agree on the valid set.

docs/catalog-integrations.md
mrouvroy-floe: This section states that a storage authority is not an alternative for an Integration-backed table and that Floecat does not fall back to one. That contradicts the new access-delegation-mode=none behavior, where SourceCatalogCredentialVendor returns null and the caller falls through to normal storage-authority resolution.

The same guide later instructs operators to set access-delegation-mode=none when Floecat already has the storage configuration the reader needs — making the two sections mutually exclusive about whether storage-authority resolution is expected to work for Integration-backed tables.

Update this section to describe the default vended-credentials path and explicitly carve out access-delegation-mode=none as the opt-in storage-authority path, so operators understand when each mechanism applies.

mrouvroy-floe (approved):
2 small improvements I think we should do but otherwise LGTM

PR #537, opened 2026-09-16
Remove redundant blob checks from warm metadata reads

mrouvroy-floe:
## Summary

A warm query already has the current pointer for each table. The pointer identifies the blob to use, so checking the blob again only to read its metadata was redundant.

This change:

- Removes the per-table manifest and stats liveness probes.
- Uses the current pointer metadata for planning and GC decisions.
- Derives ETags locally from immutable, content-addressed blob URIs.
- Keeps the existing HEAD fallback for mutable and older blob formats.
- Updates the cost tests and documentation.

For a warm request, the per-table cost is now 0 KV reads, 0 S3 GETs, and 0 S3 HEADs.
The blob is still read normally when its contents are actually required.

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [x] No breaking changes

## Checklist


service/src/main/java/ai/floedb/floecat/service/repo/impl/IndexArtifactRepository.java
```diff
@@ -930,7 +931,14 @@ tableId, snapshotId, generationId, indexArtifactTargetStorageId(target))
                     new BaseResourceRepository.NotFoundException(
                         "index artifact pointer is missing"));
     String pointerKey = pointer.getKey();
-    String etag = blobStore.head(pointer.getBlobUri()).map(header -> header.getEtag()).orElse("");
+    String etag =
+        BlobRefs.etagFromCasUri(pointer.getBlobUri())
```
mcusack-floe: Should this call the gated immutableBlobEtag() instead?

mcusack-floe (changes requested):

mcusack-floe (approved):

service/src/main/java/ai/floedb/floecat/service/repo/impl/IndexArtifactRepository.java
mrouvroy-floe: Good catch, thanks

service/src/main/java/ai/floedb/floecat/service/query/impl/QueryContextStoreImpl.java
```diff
@@ -295,15 +288,44 @@ public void registerResolvingPinBlobs(
     reachabilityGuard.publishing(
         tableId,
         () -> {
-          registerResolvingPinBlobsGuarded(queryId, tableId, clean);
+          validatePinScope(tableId, clean);
```
mcusack-floe: Pin registration must not accept a stats generation after GC has claimed it as DELETING. The publication epoch prevents a claim that overlaps or follows registration, but it cannot reverse a claim completed before registration. Generation cleanup then runs in bounded slices outside the reachability lock, so this query can commit a pin to a partially deleted generation. Atomically reject DELETING/DELETED generations during pin admission, or redesign the handoff so registration cannot occur between the claim and completed cleanup.

mcusack-floe (changes requested):

PR #542, opened 2026-09-21
 Add Generic Catalog RPCs for Engine Integrations

mrouvroy-floe:
## Summary

Introduce a generic, engine-neutral catalog read surface so any query engine can connect to Floecat easily.

### **New generic interface**

`RelationService` exposes three kind-neutral reads:

- `ListRelations`: paged listing scoped to a catalog or namespace, with optional recursion and table/view kind filters.
- `ResolveRelations`: resolves ordered candidate names for logical references such as `orders`, `public.orders`, or `sales.orders`.
- `GetRelation`: fetches a relation directly by resource ID.

Each returned relation contains:

- Resource ID and kind
- Fully qualified `NameRef`
- Display name and origin
- Optional logical schema, properties, and table/view details
- Optional queryability status and current snapshot information

Listing results preserve row order. A failed row returns a structured error with its identity instead of disappearing or failing the entire page. Resolution distinguishes `NOT_FOUND` from permission, validation, and backend failures.

### **Planning integration**

The existing `GetUserObjects` streaming API remains the planning-specific path. It accepts ordered `QueryInput` candidates, streams resolved objects, propagates logical types and schemas, and supports known relation versions to avoid repeating payloads.

### **Implementation**

- Generic listing reuses the existing pointer-backed typed page sources.
- Identity-only listings avoid relation blob reads.
- Hydrated listings reuse the same source and hydration path.
- Existing typed table/view APIs remain for management and gateway workflows.
- CLI, Trino, Iceberg, and reconciler adapters use the shared generic result handling.
- Scanner integrations use explicit topology references directly.
- No new KV keys or metadata writes are introduced.

### Compatibiility

Core PR: https://github.com/eng-floe/core/pull/2594

## Type of Change

- [ ] fix (bug fix)

## Testing

- [x] `make fmt`
- [x] `make verify`

- [ ] No breaking changes

## Checklist


core/engine-utils/src/main/java/ai/floedb/floecat/engine/catalog/RelationResults.java
```diff
@@ -0,0 +1,151 @@
...
+      throw new IllegalStateException(
+          "relation listing is incomplete: " + describeErrors(page.errors()));
+    }
+  }
+
+  /**
+   * Returns the first resolved relation, preserving the distinction between absence and failure. A
+   * missing result is the legacy no-match case; an in-band error is never treated as absence.
+   */
+  public static Relation requireResolved(ResolveRelationsResponse response) {
```
mcusack-floe: Malformed responses are treated as NOT_FOUND: zero results become absence, while multiple results are silently truncated. Require exactly one result.

service/src/main/java/ai/floedb/floecat/service/error/impl/GrpcErrors.java
```diff
@@ -36,6 +36,17 @@
 
 public final class GrpcErrors {
 
+  /** Whether a gRPC failure can be isolated to one relation in a batch operation. */
```
mcusack-floe: Relation-local INTERNAL failures are classified as request-wide failures, causing one corrupt relation to abort the entire listing or resolution batch.

protocol-gateway/iceberg-rest/src/main/java/ai/floedb/floecat/gateway/iceberg/rest/services/resolution/NameResolution.java
```diff
@@ -77,9 +86,34 @@ public static ResourceId resolveView(
...
+  private static Relation resolveRelation(GrpcServiceFacade client, NameRef ref) {
+    var response =
+        client.resolveRelations(
+            ResolveRelationsRequest.newBuilder()
+                .addReferences(RelationReference.newBuilder().addCandidates(ref))
+                .build());
+    if (response == null) {
+      return null;
+    }
+    try {
```
mcusack-floe: In-band resolution errors are rethrown as RelationResolutionException, but the REST gateway has no mapper for that exception. Permission and validation failures become generic HTTP 500 responses.

service/src/main/java/ai/floedb/floecat/service/catalog/impl/surface/CatalogSurfaceRelations.java
```diff
@@ -0,0 +1,407 @@
...
+      boolean includeSchema,
+      boolean includeStatus,
+      String corr) {
+    var result = ResolveRelationResult.newBuilder();
+    for (NameRef candidate : reference.getCandidatesList()) {
+      ResourceId id = resolved.getOrDefault(candidate, Optional.empty()).orElse(null);
+      if (id == null) {
+        continue;
+      }
+      result.setResolvedName(candidate);
```
mcusack-floe: Identity-only resolution uses the input candidate as Relation.name, returning unqualified or incorrectly cased names instead of the canonical fully qualified name.

service/src/main/java/ai/floedb/floecat/service/catalog/impl/surface/CatalogSurfaceRelations.java
```diff
@@ -0,0 +1,407 @@
...
+    RelationPageCursor cursor = RelationPageCursor.decode(pageIn.token, corr);
+    String scopeFingerprint = scopeFingerprint(request, kinds, accountId, context);
+    if (!pageIn.token.isBlank()) {
+      cursor.requireScope(scopeFingerprint, pageIn.token, corr);
+    }
+
+    var results = new ArrayList<RelationListResult>(want);
+    var catalogNames = new HashMap<ResourceId, String>();
+    int total = total(segments, request.getIncludeTotal(), cursor.total(), accountId);
+    String innerToken = cursor.innerToken();
```
mcusack-floe: When the cursor namespace is deleted, its inner page token is incorrectly passed to the next namespace, potentially rejecting the request or skipping relations.

mcusack-floe (changes requested):

core/engine-utils/src/main/java/ai/floedb/floecat/engine/catalog/RelationResults.java
mrouvroy-floe: Fixed in d6992d858. Any count other than exactly one is now a protocol violation (MC_INTERNAL) rather than absence, so zero results no longer read as "not found" and extra results aren't silently dropped. notFoundError() went with it. Three reconciler fakes were returning an empty response to mean not-found; they now model what the server actually sends.

service/src/main/java/ai/floedb/floecat/service/error/impl/GrpcErrors.java
mrouvroy-floe: Fixed in 420598de5. INTERNAL is relation-scoped now, so one relation failing to hydrate is reported in its own error envelope instead of taking the listing or resolve batch with it. Corruption still fails the whole call, deliberately — CorruptionException isn't a gRPC status, so it stays unscoped and propagates, which is what NodeLoader.reload intends.

protocol-gateway/iceberg-rest/src/main/java/ai/floedb/floecat/gateway/iceberg/rest/services/resolution/NameResolution.java
mrouvroy-floe: Fixed in d6992d858. Added RelationResolutionExceptionMapper as a @Provider, reusing the extracted ErrorMapper.httpForCode/typeForCode so it can't drift from the existing mapper. Permission → 403, validation → 400, unmapped → 500. Covered by RelationResolutionExceptionMapperTest.

service/src/main/java/ai/floedb/floecat/service/catalog/impl/surface/CatalogSurfaceRelations.java
mrouvroy-floe: Fixed across 4072b2fff and db370b3c7.

Casing: builtins were matched case-insensitively while user relations were exact. There's one policy now — everything matches its stored spelling. NameRefUtil splits into matchKey (preserves case, keys what a caller-supplied name is matched against) and identityKey (folds, keys signatures and definition merging, where both sides are definitions and type names like VARCHAR are authored uppercase).

Qualification: identity-only resolve echoed the candidate, so a builtin could come back unqualified while the same relation with include_schema came back qualified. It now reports the name the graph holds, via resolveSystemRelationName, which answers from the snapshot without touching the user graph — no extra blob read, so PointerListingStoreCostIT still holds. Pinned by a test asserting both paths report the same name.

service/src/main/java/ai/floedb/floecat/service/catalog/impl/surface/CatalogSurfaceRelations.java
mrouvroy-floe: Fixed in d6992d858. The inner token is only inherited when the resumed segment is genuinely the one that minted it; otherwise the walk starts the next segment from its beginning.

While testing this I found two further paging bugs, fixed in a7729971d: an unbounded page_size sized the page buffer directly, so page_size = 2147483647 hit OutOfMemoryError before reading anything (now capped by floecat.relation.list.max-page-size); and a namespace containing | minted a token the server then rejected, since | is the cursor's field separator — the segment key is base64url'd in the token now.

